PR #188, opened 2015-01-25
grub: use uefi firmware workaround

codeworkx:
Source: https://wiki.archlinux.org/index.php/GRUB#UEFI_firmware_workaround

Please test before pulling!!!


manjaro:
Code looks fine from my end. I can only test it within VirtualBox but not on real hardware. @demmm can you check it?

demmm:
Sorry, not using GRUB, gummiboot is best to install from just /boot and by default installs that file then in that path.

manjaro:
@demmm: so code-wise it is also fine for you?
@codeworkx: as it doesn't hurt anyone I'd suggest to merge your request.
@teo: any thoughts on this one?

teo:
This is non-destructive, and disk space is cheap. I'm merging. Thanks for the patch.

PR #189, opened 2015-02-03
Completed Mexican Spanish support.

jnuserful:


philmmanjaro:
This rather get added to our transifex project than here.

teo:
As @philmmanjaro said, if you add these strings to our Transifex project (https://www.transifex.com/projects/p/calamares/) it will get pulled automatically.

PR #195, opened 2015-02-20
[Proposition] Python Better Docstrings

Gormogon:
Proposing better in-code documentation (eventually covering C++ as well).  This takes care of most/if not all of the Python code (may have missed a few).  These are obviously not filled out yet, but it's a start.

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -123,10 +123,10 @@ def install_gummiboot(efi_directory):
     loader_path = os.path.join(
         install_efi_directory, "loader", "loader.conf")
     subprocess.call(["gummiboot", "--path={!s}".format(install_efi_directory), "install"])
-    kernel_line = get_bootloader_entry_name("default")
+    kernel_line = get_kernel_line("default")
     print("Configure: \"{!s}\"".format(kernel_line))
     create_conf(uuid, conf_path, kernel_line)
-    kernel_line = get_bootloader_entry_name("fallback")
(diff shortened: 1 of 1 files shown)


Gormogon:
You're correct about the single-line short descriptions. I will fix this.  As for the empty doctring, while I agree that empty docstrings are not good, I tried to set up docstrings EVERYWHERE there should be a docstring so that they can eventually be filled out. In other words, the ones that are currently empty are placeholders for where we need doctrings to be filled out.

Since currently the only in-code documentation seems to be a few scattered comments I figured this is a good place to start to at least get started and add consistency...

rshipp:
This is just my opinion, but I believe it's better to leave the docstrings out unless they are being filled in with useful information. Since docstrings are necessary everywhere, leaving a placeholder doesn't seem too useful to me. 

Thanks for taking the time to look over and fix all these Python modules. :sparkles: :cake: :sparkles:

teo:
Thanks for starting this, @Gormogon . I agree with @rshipp though that empty docstrings are less than ideal. Let's do something else: @Gormogon could you please push this to a new branch in the calamares/calamares repo? So we can fill out the docs over the next few days and merge when it's all filled out.
Would that work for everyone?

Gormogon:
That sounds like a greate idea @teo. This is better then cluttering mater with empty docstrings.  Not sure that I have permission to push this directly to calamares/calamares, but I'll set up another branch and give it a shot.

teo:
@Gormogon you do now.

PR #197, opened 2015-02-21
Python docstrings

manjaro:
In-code documentation for Python jobmodules. This way new developers will better understand our modules. Please verify added doc strings and check if your module still works.


Gormogon:
This is looking great! Glad to see these getting filled out. Can you please review and fix PEP 257 violations. I already found a few.

https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

Other than that, this looks good.

philmmanjaro:
When I find time, yes ...

PR #198, opened 2015-02-25
Astyle Reformat

Gormogon:
Re-formats all c++ code in accordance with:
https://github.com/calamares/calamares/blob/master/hacking/astylerc

If for some reason the astylerc file is out-dated or incorrect, please let me know so I can make the appropriate changes.  Don't want to commit code that is way off of the project's style guide.

This does not cover the partitioning module.

### DIFF
--- a/src/calamares/CalamaresApplication.cpp
+++ b/src/calamares/CalamaresApplication.cpp
@@ -37,7 +37,7 @@
 #include <QFileInfo>
 
 
-CalamaresApplication::CalamaresApplication( int& argc, char *argv[] )
+CalamaresApplication::CalamaresApplication( int& argc, char* argv[] )
     : QApplication( argc, argv )
     , m_mainwindow( 0 )
 {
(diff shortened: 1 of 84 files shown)


teo:
astyle is awesome, except when it's not :(
Unfortunately we cannot rely on it completely as we can on autopep8 for Python, for instance, and C++ is a whole different kind of beast compared to plain, clean, simple Python. Specifically, astyle doesn't do what I'd like it to do in long, nested arguments lists. This is especially noticeable in our boost::python interface definitions, for example. I also think that explicit braces {} ought to be left to the programmer to decide on a case by case basis, along with various other things.
I do like that you've fixed all the occurrences of * positioning though.
I hope you don't mind and understand why I'm not inclined to merge this as it is.
I'm not sure it makes sense to spend lots of time on this. I find astyle to be useful as a checker but bad as a formatter.
I propose two courses of action, I'm fine with both, let me know how you'd like to proceed:
1) we drop this, and focus on other stuff (see suggestion below);
2) you use astyle fix up the * positioning only, leaving everything else untouched.

A somewhat related suggestion: in my opinion something *really* useful would be to put together a commit hook that runs astyle read-only to warn the developer of potential astyle violations in modified files for every commit right at the time of committing, without doing hard enforcement.

Gormogon:
@teo I completely understand.  Some things with astyle come out looking great and other look like a mess.  Hopefully later today when I have some time, I'll go back and only change * positioning.  Creating the commit hook does sound like a good idea, so I may work on this at a later date as well.

PR #202, opened 2015-02-26
Align Pointers

Gormogon:
Aligning pointers to type as previously discussed with Teo.


teo:
Thanks for the patch.

PR #203, opened 2015-02-27
Try to umount existing mounts to temp dirs before unpacking.

teo:


philmmanjaro:
**unpackfs** is not affected by this. Only **partition** module. This code should be done inside of that module. That is why I created **umountall** module and call it before **partition** module.

teo:
@philmmanjaro that's right, disregard this.

PR #204, opened 2015-02-27
New ClearTempMountsJob, to be executed before partitioning jobs.

teo:


philmmanjaro:
Works now. Tested and released as 1.0.1.12-1 @manjaro

teo:
Sneaky. Thanks for the fix, @Kirek

PR #205, opened 2015-03-01
Allow local installs with the packages module

rshipp:
Courtesy of @Inkane. Adds the ability to use pacman -U to install
local packages.

Note: the `local_install` option in the YAML will work for pacman only,
until someone edits `install` to support the rest.

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -91,7 +91,7 @@ def run_operations(pkgman, entry):
             pkgman.install(entry[key])
         elif key == "remove":
             pkgman.remove(entry[key])
-        elif key == "local_install":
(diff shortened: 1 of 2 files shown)


teo:
Looks ok, but I'd like you to make the key "localInstall" instead of "local_install" since that's the convention in all our conf files.

philmmanjaro:
Fine with me ...

rshipp:
This is waiting on @AlmAck and I to have time to test it on a real ISO.

AlmAck:
Tested and works fine, only one thing to consider, every extra dependencies must be added to the  localInstall list

s8321414:
So should be all fine here hmmm?

PR #206, opened 2015-03-04
packages: Add Entropy (equo) support

necrose99:
entropy handles fetch rides atop gentoo portage & handles binaries rather well. 

https://wiki.sabayon.org/index.php?title=En:List_Of_Equo_Functions

most commonly used : equo up or equo update (get repo & sync) 
equo install package-foo, equo remove  dev-perl/old-stale-perl-mod-toomb-.01
,equo upgrade/equo u  , 
equo search package name  IE perl-5.x.21  & perl-5.x.24 
shows instalable packages. equo repo mirrorsort sabayon-weekly (gets speed of mirrors and prioritizes faster)

equo repo disable (/enable) sabayon-limbo (disables limbo the not always stable test branch) to equo repo enable..... 

equo supports multi-fetch and now parrallell Install/uninstall. 

hopefully we can get an ebuild up for it to build for spike-pentesting sometime....


src/modules/packages/main.py
```diff
@@ -54,7 +54,9 @@ def install(self, pkgs):
             check_chroot_call(["apt-get", "-q", "-y", "install"] + pkgs)
         elif self.backend == "pacman":
             check_chroot_call(["pacman", "-Sy", "--noconfirm"] + pkgs)
-
+# https://wiki.sabayon.org/index.php?title=En:List_Of_Equo_Functions
+                  elif self.backend == "Entropy":
```
teo: Please name the backend "entropy" all lowercase, and add it to the bad backend check near the end of this file.

src/modules/packages/main.py
```diff
@@ -54,7 +54,9 @@ def install(self, pkgs):
...
+           check_chroot_call(["equo",  "install"] + pkgs)
```
teo: Make sure you are making non-interactive calls to `equo`. Questions like "are you sure you want to install blah blah [Y/n]" from `equo` would block the job.

src/modules/packages/main.py
philmmanjaro: Also I think all elif should be in one line distance:
<pre>elif self.backend == "pacman":
    do something
elif self.backend == "entropy":
    do something</pre>

src/modules/packages/main.py
```diff
@@ -77,7 +79,8 @@ def remove(self, pkgs):
             check_chroot_call(["apt-get", "--purge", "-q", "-y", "autoremove"])
         elif self.backend == "pacman":
             check_chroot_call(["pacman", "-Rs", "--noconfirm"] + pkgs)
-
+                   elif self.backend == "Entropy":
```
teo: It looks like your indentation is off here.

src/modules/packages/main.py
teo: Not just should be, it *must* be, because Python is sensitive to indentation.

philmmanjaro: I just wanted to be nice ;)

kkofler:
It helps if pull requests and commits have clear descriptions. :-) "Update main.py" is very vague considering that every Python module has a main.py file.

necrose99:
via  terminal shell 
equo install xfce4-meta <CR/Enter>  will pull in all of xfce 4  and any depends even if... therer's a ton of them and without any further interaction less there are  MAJOR issues like repo conflits etc.
 
only equo -ask install foobar  gives a Y/N promt ...  

equo install entropy rigo equo rigo-dameon   --relaxed  (lowerers deps checking to releaxed) 
sometimes useful in upgrading entropy 1st... 
equo install foobar --nodeps 
(might need to run equo deptest or equo libtest latter) usefull in forcing a package to upgrade when giving errors. 

I'm a  bit dyslexic so I didnt notice missed indents , also I've had not much sleep. 
https://spike-pentesting.org/pages/repository add in a stock sabayon  live in virtualbox 
http://sourceforge.net/projects/spikepentesting/files/isos/Spike_Linux_1.0a_amd64_mate.iso.pkglist/download  
add in equo install prior to the package names save as a install-me.sh  (if you need a test case)
my box crashed hack the gentoo world file - versions in most cases, except where desierble... 
equo install package name <CR> rinse & repeate 
3500ish(3450)  packages later , KDE , XFCE,LXQT ,Mate , dev tools , Pentest toys back and installed. 
technically I could have parsed equo install  a b c d e f g........ all packages like so just with 10 via hand easy , 3500 not so much (and re-installs due to list and deps pulled in)

Sabayon uses Entropy/portage/packagekit  equo (entropy) depends on portage/packagekit 

Your welcome to tweak the code as you see fit, I'm needing a snozze , 
I'm just a user hating thier port of anaconda as it is so Buggy and half the time dies. or boorks on my older UEFI laptop. 
the older one prior to may 2014 worked & still dose miles better than the newer. (i am forced to Keep older ISO's about ) 

I'm hoping entropy/portage support will encourage them to ditch that POS bugg riddled installer.

kkofler:
Entropy does not depend on PackageKit, it's the other way round (PackageKit has an Entropy backend). This is important to know because anything that uses PackageKit probably won't work in a chroot, because you'd need to also run the PackageKit daemon inside the chroot for it to do the right thing. (That's also why we can't just use pkcon on all distros.)

necrose99:
https://packages.sabayon.org/show/packagekit,117275,sabayon-weekly,amd64,5,standard
anyhow  package kit is core to the system,.... anyhow I'm exuasted and need a bit of sleep. 
getting looppy ..... I just hope the code for entropy helps and the information helps. 
of late the anaconda installer many users for sabayon hate... as its buggy.

src/modules/packages/main.py
demmm: line 58 has wrong indent, fails python check

philmmanjaro:
**Entropy** should be still lower-case as all the other backends are named here too. Same goes for the not yet supported **Portage** ...

src/modules/packages/main.py
```diff
@@ -53,8 +53,15 @@ def install(self, pkgs, from_local=False):
         elif self.backend == "apt":
             check_chroot_call(["apt-get", "-q", "-y", "install"] + pkgs)
         elif self.backend == "pacman":
+
+            check_chroot_call(["pacman", "-Sy", "--noconfirm"] + pkgs)
+                    elif self.backend == "Entropy":
+            check_chroot_call(["equo", "-v", "install"] + pkgs)
+=======
```
teo: It seems like this was left behind after a merge.

necrose99:
already merged into Gentoo /Sabayon builds.

kkofler:
It would still make sense to get this added to the upstream code though.

PR #208, opened 2015-03-16
Show efi mountpoint when creating/editing partitions

Kirek:
Remove fixed mountpoint list in the .ui file and create a list of mount points including the customizable efi mount point, when running in an efi capable system.

### DIFF
--- a/src/modules/partition/gui/CreatePartitionDialog.cpp
+++ b/src/modules/partition/gui/CreatePartitionDialog.cpp
@@ -26,6 +26,8 @@
 #include <ui_CreatePartitionDialog.h>
 
 #include <utils/Logger.h>
+#include "GlobalStorage.h"
+#include "JobQueue.h"
 
 // CalaPM
 #include <core/device.h>
@@ -35,6 +37,7 @@
 
 // Qt
 #include <QComboBox>
+#include <QDir>
 #include <QSet>
 
 static QSet< FileSystem::Type > s_unmountableFS(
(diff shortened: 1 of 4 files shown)


src/modules/partition/gui/CreatePartitionDialog.cpp
```diff
@@ -52,6 +55,14 @@ CreatePartitionDialog::CreatePartitionDialog( Device* device, PartitionNode* par
 {
     m_ui->setupUi( this );
 
+    QStringList mountPoints;
+    mountPoints << "/" << "/boot" << "/home" << "/opt" << "/usr" << "/var";
```
teo: It would be better to use a C++11 initialization list here instead of appending.
Like this: `QStringList mountPoints = { "/", "/boot",... };`

src/modules/partition/gui/EditExistingPartitionDialog.cpp
```diff
@@ -45,6 +48,14 @@ EditExistingPartitionDialog::EditExistingPartitionDialog( Device* device, Partit
 {
     m_ui->setupUi( this );
 
+    QStringList mountPoints;
+    mountPoints << "/" << "/boot" << "/home" << "/opt" << "/usr" << "/var";
```
teo: See CreatePartitionDialog.cpp.

teo:
Thanks for the patch, this is great.
Due to a synchronization issue this was merged, but it's ok. I would appreciate if you could please take care of that minor nitpick, but otherwise it's all good.

PR #211, opened 2015-03-19
Eliminate CalamaresStyle and get style constants from Branding instead.

teo:


demmm:
Tested with a few different custom branding.desc, works as expected, example:
https://github.com/KaOSx/calamares/blob/master/src/branding/kaos_branding/branding.desc#L21
snapshot59

PR #212, opened 2015-03-25
Use which to find plymouth instead of hard coding /usr/bin

shadeslayer:
Use 'which' to find plymouth instead of hard coding /usr/bin
    
Debian ships plymouth in  /bin instead of /usr/bin, so we should use
'which' to figure out if it's installed.

### DIFF
--- a/src/modules/grubcfg/main.py
+++ b/src/modules/grubcfg/main.py
@@ -33,11 +33,11 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     default_dir = os.path.join(root_mount_point, "etc/default")
     default_grub = os.path.join(default_dir, "grub")
     distributor_replace = distributor.replace("'", "'\\''")
-    plymouth_bin = os.path.join(root_mount_point, "usr/bin/plymouth")
+    plymouth_bin = libcalamares.utils.chroot_call("sh -c \"which plymouth\"")
     use_splash = ""
     swap_uuid = ""
 
-    if os.path.exists(plymouth_bin):
+    if plymouth_bin == 0:
         use_splash = "splash"
 
     for partition in partitions:


kkofler:
But the script doesn't run inside the chroot. That's why it prepends root_mount_point to the path, to point to the file inside the chroot. You removed that, so now you're using the plymouth binary from the host system. Your shutil.which call also searches for plymouth on the host system, not inside the chroot.

shadeslayer:
@kkofler fixed :)

src/modules/grubcfg/main.py
```diff
@@ -33,11 +33,11 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     default_dir = os.path.join(root_mount_point, "etc/default")
     default_grub = os.path.join(default_dir, "grub")
     distributor_replace = distributor.replace("'", "'\\''")
-    plymouth_bin = os.path.join(root_mount_point, "usr/bin/plymouth")
+    plymouth_bin = libcalamares.utils.chroot_call("sh -c \"which plymouth\"")
     use_splash = ""
     swap_uuid = ""
 
-    if os.path.exists(plymouth_bin):
+    if plymouth_bin == 0:
```
teo: In Python, empty strings are falsy, so I think it would be more Pythonic to just do `if plymouth_bin:`

src/modules/grubcfg/main.py
```diff
@@ -33,11 +33,11 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     default_dir = os.path.join(root_mount_point, "etc/default")
     default_grub = os.path.join(default_dir, "grub")
     distributor_replace = distributor.replace("'", "'\\''")
-    plymouth_bin = os.path.join(root_mount_point, "usr/bin/plymouth")
+    plymouth_bin = libcalamares.utils.chroot_call("sh -c \"which plymouth\"")
```
teo: `libcalamares.utils.chroot_call` does not return the output. What you're looking for is `check_chroot_output`.

teo: Unless you don't even need the path further on? Were you actually looking to just get an exit status?

teo: Ok, please disregard this.

PR #213, opened 2015-03-31
Preserve kernel parameters that aren't handled by grubcfg.

teo:


kkofler:
Yes. That `if not have_kernel_cmd:` code path is also used when the file was written completely from scratch, which is the case on Fedora. I also guess there may be other distributions where the default grub.cfg exists, but does not contain a GRUB_CMDLINE_LINUX_DEFAULT line.

PR #215, opened 2015-04-23
Patch 1

necrose99:


src/modules/packages/main.py
```diff
@@ -55,7 +55,8 @@ def install(self, pkgs, from_local=False):
         elif self.backend == "pacman":
             pacman_flags = "-U" if from_local else "-Sy"
             check_chroot_call(["pacman", pacman_flags, "--noconfirm"] + pkgs)
-
+                    elif self.backend == "Entropy":
```
teo: I suspect you might be using tabs here. Please always use spaces for indentation, not tabs.

kkofler:
Entropy support is already included in the current Calamares, so this pull request is obsolete.

PR #219, opened 2015-06-08
Improving support for Gentoo based distros 

dgikiller:


CMakeModules/BoostPython3.cmake
```diff
@@ -32,4 +40,15 @@ macro( find_boost_python3 boost_version python_version found_var )
             break()
         endif()
     endforeach()
+    # The following loop chage the searched name for Gentoo based distributions
```
teo: Typo: chage -> changes

PR #220, opened 2015-06-08
Improving support for Gentoo based distros 

dgikiller:


CMakeModules/BoostPython3.cmake
```diff
@@ -32,4 +40,15 @@ macro( find_boost_python3 boost_version python_version found_var )
             break()
         endif()
     endforeach()
+    # The following loop chages the searched name for Gentoo based distributions
```
teo: Still got a typo: chages -> changes.

philmmanjaro:
@dgikiller: you don't have to close your pull request all the time. Simply edit those mentioned typo in your branch. It will be automatically updated here also. Then we can merge it.

PR #227, opened 2015-08-14
Add a webview module

shadeslayer:
Add a webview module so that calamares can load a webpage.
This is useful when configuring webservices on your device such as
ownCloud.

### DIFF
--- a/src/modules/webview/WebViewStep.cpp
+++ b/src/modules/webview/WebViewStep.cpp
@@ -1,6 +1,6 @@
 /* === This file is part of Calamares - <http://github.com/calamares> ===
  *
- *   Copyright 2014-2015, Teo Mrnjavac <teo@kde.org>
+ *   Copyright 2015, Rohan Garg <rohan@garg.io>
  *
  *   Calamares is free software: you can redistribute it and/or modify
  *   it under the terms of the GNU General Public License as published by
@@ -72,7 +72,7 @@ WebViewStep::isNextEnabled() const
 bool
 WebViewStep::isBackEnabled() const
 {
-    return false;
+    return true;
 }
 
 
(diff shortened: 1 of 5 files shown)


shadeslayer:
Screenshot in case anyone is interested

image

demmm:
Can this be connected to the Welcome page links too?  So no need then to launch an external browser to view "Known issues" for example?

shadeslayer:
@demm one could make it so that it loads the release notes when you click next from the welcome page.

src/modules/webview/WebViewStep.cpp
```diff
@@ -0,0 +1,115 @@
+/* === This file is part of Calamares - <http://github.com/calamares> ===
+ *
+ *   Copyright 2014-2015, Teo Mrnjavac <teo@kde.org>
```
teo: Please add your copyright line.

src/modules/webview/WebViewStep.cpp
```diff
@@ -0,0 +1,115 @@
...
+WebViewStep::isNextEnabled() const
+{
+    return true;
+}
+
+
+bool
+WebViewStep::isBackEnabled() const
+{
+    return false;
```
teo: This should probably be `true`.

src/modules/webview/WebViewStep.h
```diff
@@ -0,0 +1,66 @@
+/* === This file is part of Calamares - <http://github.com/calamares> ===
+ *
+ *   Copyright 2014-2015, Teo Mrnjavac <teo@kde.org>
```
teo: Copyright line.

shadeslayer:
@teo fixed all the issues

kkofler:
Shouldn't this module with its QtWebKit dependency be optional (compile-time option)?

teo:
@kkofler it is already possible to exclude certain modules from building with -DSKIP_MODULES

PR #228, opened 2015-08-26
Romanian localization

Jobava:
I could not get an answer on Transifex on contributing with a Romanian localization for Calamares.

I am submitting a translation by means of a pull request, if possible, but it would have to be mirrored on Transifex as well probably.

I also contributed translations for Sabayon Entropy Client, Entropy Rigo and App-witchcraft:

https://www.transifex.com/explore/projects/?search=entropy

In any event, there it is!

Jobava

### DIFF
--- a/lang/calamares_ro.ts
+++ b/lang/calamares_ro.ts
@@ -1,33 +1,35 @@
-<?xml version="1.0" ?><!DOCTYPE TS><TS language="ro" version="2.1">
+<?xml version='1.0'?>
+<!DOCTYPE TS>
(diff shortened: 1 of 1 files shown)


Jobava:
My Transifex profile https://www.transifex.com/user/profile/Jobava/

teo:
That's because the language coordinator for Romanian should have accepted your request, which he didn't. I've added you to the team now, so you can add your translation directly to Transifex. Closing this to avoid redundant workflow.

PR #229, opened 2015-08-31
Allow to disable services

bhush9:
There might be some services that are enabled in live environment but
not useful for the installed environment and needs disabling by default

### DIFF
--- a/src/modules/services/main.py
+++ b/src/modules/services/main.py
@@ -26,6 +26,7 @@ def run():
     """ Setup systemd services """
     services = libcalamares.job.configuration['services']
     targets = libcalamares.job.configuration['targets']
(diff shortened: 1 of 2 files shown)


src/modules/services/main.py
```diff
@@ -51,4 +52,15 @@ def run():
                 libcalamares.utils.debug("Cannot enable systemd target {}".format(tgt['name']))
                 libcalamares.utils.debug("systemctl enable call in chroot returned error code {}".format(ec))
 
+    for dbl in disable:
```
teo: Can you please add some sample config to `services.conf` too?

bhush9: Fixed.

PR #230, opened 2015-11-09
implement https://calamares.io/bugs/browse/CAL-341

demmm:
Let me know if it is an issue this work was done in the choicepreview branch, it is what I'm working in and using.  If it is then will remove this fork and redo in master.
This change is tested to build and work in the GUI.

### DIFF
--- a/src/modules/users/UsersPage.cpp
+++ b/src/modules/users/UsersPage.cpp
@@ -426,3 +426,9 @@ UsersPage::setAutologinDefault( bool checked )
 {
     ui->checkBoxAutoLogin->setChecked( checked );
 }
+
+void
+UsersPage::setBoxReusePassword( bool checked )
+{
+    ui->checkBoxReusePassword->setChecked( checked );
+}
--- a/src/modules/users/UsersPage.cpp
+++ b/src/modules/users/UsersPage.cpp
@@ -428,7 +428,7 @@ UsersPage::setAutologinDefault( bool checked )
 }
 
 void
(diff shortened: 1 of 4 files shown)


philmmanjaro:
Hmm, I quite don't know why the user has to have a checkbox then? What happens if he has it unchecked?

demmm:
@teo 
Even if this pull request is acceptable as is now, please do not merge.
I do see a flaw how it is setup in current git master/choicepreview.
When the default setting is used, use same password for user and root, then when the user unchecks the setting, user can still continue (next is enabled) without entering a matching root password.
I will work on setting next as disabled in that case, and will have that toggle carry over to the option of starting with not re-using password.
So imo, "next" should be disabled anytime the root password box is shown, "next" should be enabled anytime it is not.

philmmanjaro:
Would it not been better to check if the **passwords** match before enabling **Next**? Anyway, good catch in that regard.

demmm:
Nothing to do with password matching, as soon as you start typing the root password check section of userpage.cpp kicks in, and next is disabled.
The issue is as explained in https://calamares.io/bugs/browse/CAL-344

philmmanjaro:
Ok, now I got it. Point taken. Blank **root** passwords would be **Windows** then ;)

demmm:
@teo 
Name change implemented, no luck in fixing bug https://calamares.io/bugs/browse/CAL-344
Probably best to have that in a different pull request/commit anyway.
Can be merged now, if useful for Calamares.

PR #231, opened 2015-11-17
CAL-345: don't set default "users" group on home dirs

codeworkx:
Fixes CAL-345.

Changing group of home dir to the default "users" group might be a security risk
because every user which belongs to the default "users" group
might be able to access private data on home dirs of other users.

### DIFF
--- a/src/modules/users/UsersPage.cpp
+++ b/src/modules/users/UsersPage.cpp
@@ -105,7 +105,7 @@ UsersPage::isReady()
 
 
 QList< Calamares::job_ptr >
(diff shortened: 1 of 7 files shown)


src/modules/users/UsersPage.cpp
```diff
@@ -105,7 +105,7 @@ UsersPage::isReady()
 
 
 QList< Calamares::job_ptr >
-UsersPage::createJobs( const QString& defaultUserGroup, const QStringList& defaultGroupsList )
+UsersPage::createJobs( const QStringList& defaultGroupsList )
```
Pulfer: This change should be reflected in  UsersPage.h as well.

PR #233, opened 2016-01-07
calamares: sidebarTextHighlight for branding

MingcongBai:
This commit adds support for distribution to define `sidebarTextHighlight` to their `branding.desc` file, to control the left panel (sidebar) with highlight behind the current step (`isCurrent`) text label.

And here attaches a screenshot showing what the parameter does.

screenshot_20160106_200615


teo:
Ideally, if an entry is not found in branding.desc, it should use the application palette's background role as before.

adriaandegroot:
I added the fallback code @teo  mentioned and merged this. Thanks!

PR #234, opened 2016-01-13
Creating the option to copy an install log to target

demmm:
instead of having to hope users save an installation log before leaving live mode,
this commit gives distributions the option to have a full calamares -d log available on
the newly installed system
see the readme.md for details
this is fully tested and working properly, using the example launcher script & .desktop line
as noted in the readme

### DIFF
--- a/src/modules/umount/README.md
+++ b/src/modules/umount/README.md
@@ -1,0 +1,17 @@
+### Umount Module
+---------
+This module represents the last part of the installation, the unmounting of partitions used for the install.  It is also the last place where it is possible to copy files to the target system, thus the best place to copy an installation log.
+
+To be able to use the copy an installation log of this module you will need to include a log creation to your launcher script or add it to the used calamares.desktop, example of a launcher script:
+
+```
+#!/bin/sh
+sudo /usr/bin/calamares -d > installation.log 
+```
+Example desktop line:
+```
+Exec=sudo /usr/bin/calamares -d > installation.log
+```
+Set the source and destination path of your install log in umount.conf
+If you do not wish to use the copy of an install log feature, no action needed, the module will not execute the copy of a log code if no log is present.
+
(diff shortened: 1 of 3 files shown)


src/modules/umount/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+To be able to use the copy an installation log of this module you will need to include a log creation to your launcher script or add it to the used calamares.desktop, example of a launcher script:
+
+```
+#!/bin/sh
+sudo /usr/bin/calamares -d > installation.log 
+```
+Example desktop line:
+```
+Exec=sudo /usr/bin/calamares -d > installation.log
```
teo: We already write the log to disk, is this necessary? I've just pushed a change which makes Calamares print out the log path for convenience. This path is built to be in the user's cache directory, which under Plasma ends up as `~/.cache/Calamares/Calamares/Calamares.log`. It's easier and more reliable to just copy the log file from there.

demmm: I know the log is created there, but no where near as complete as sudo calamares -d
Everywhere we always ask for this full log instead of the Calamares.log

demmm: One more note about this, Calamares is not run as user, so the log ends up in the root dir, much harder for users to see.
But no issue to adjust this commit to make 
```/root/.cache/Calamares/Calamares/Calamares.log``` the default and add the option for a sudo calamares -d log as another option

Comment (teo) [inline src/modules/umount/README.md]: Yeah, I guess you're right, the problem with the log file is that it doesn't capture the output of child processes, if any.

File: src/modules/umount/main.py [hunk 2]
```diff
@@ -47,6 +49,13 @@ def run():
     :return:
     """
     root_mount_point = libcalamares.globalstorage.value("rootMountPoint")
+    log_source = libcalamares.job.configuration["srcLog"]
+    log_destination = libcalamares.job.configuration["destLog"]
+    
+    # copy installation log before umount
+    if(os.path.exists('{!s}'.format(log_source))):
```
teo: Can you please just add another if to check whether srcLog is set (is not empty)? That way if the deployer doesn't wish to copy the log they simply don't set srcLog.

PR #237, opened 2016-03-15
Enable support for 32bit UEFI systems with Grub

JoernSchoenyan:
Read the UEFI bitness exposed to the file system and install the correct
variant of Grub, depending on the UEFI bitness.


teo:
Thanks for the patch!

PR #240, opened 2016-03-27
Update calamares_fr.ts

tapwag:
Fixed French typo as reported at https://issues.openmandriva.org/show_bug.cgi?id=1568


philmmanjaro:
Please use the [transifex project](https://www.transifex.com/calamares/calamares/) for this. Thanks.

PR #241, opened 2016-05-05
Disk selections in partitioner are not sticky[CAL-361]

tctara:
FIX bug: [[CAL-361](https://calamares.io/bugs/browse/CAL-361)]

I take this approach because the bootLoadercomboBox's currentTextChanged signal triggered both when it's users manually selecting or repopulation of the core reset. It makes it hard to differentiate between the two and mark the widget as dirty accordingly, so I only save the index right before the core reset happen and select it back right after the core reset.

### DIFF
--- a/src/modules/partition/gui/PartitionPage.cpp
+++ b/src/modules/partition/gui/PartitionPage.cpp
@@ -54,6 +54,7 @@
 PartitionPage::PartitionPage( PartitionCoreModule* core, QWidget* parent )
     : QWidget( parent )
     , m_ui( new Ui_PartitionPage )
+    , m_lastSelectedBootLoaderIndex(-1)
     , m_core( core )
 {
     m_ui->setupUi( this );
@@ -73,6 +74,11 @@ PartitionPage::PartitionPage( PartitionCoreModule* core, QWidget* parent )
              [ this ]( const QString& /* text */ )
     {
         updateFromCurrentDevice();
+    } );
+    connect( m_ui->bootLoaderComboBox, static_cast<void(QComboBox::*)(const QString &)>(&QComboBox::activated),
+                [ this ]( const QString& /* text */ )
+    {
+        m_lastSelectedBootLoaderIndex = m_ui->bootLoaderComboBox->currentIndex();
     } );
 
     connect( m_ui->bootLoaderComboBox, &QComboBox::currentTextChanged,
@@ -94,7 +100,7 @@ PartitionPage::PartitionPage( PartitionCoreModule* core, QWidget* parent )
         m_ui->bootLoaderComboBox->hide();
         m_ui->label_3->hide();
     }
-    
+
     CALAMARES_RETRANSLATE( m_ui->retranslateUi( this ); )
 }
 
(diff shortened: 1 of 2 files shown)


teo:
My concern with this solution is that it looks more complicated and potentially error-prone than it should. Having to set the index in all of those methods is risky for maintenance.

Your observation is correct, it is true that QComboBox's currentTextChanged, currentIndexChanged, etc. are emitted even when the index changes programmatically, but you could use [QObject::blockSignals](https://doc.qt.io/qt-5/qobject.html#blockSignals) before/after your index update. That way you can safely store the selected index as m_lastSelectedBootLoaderIndex in response to user-triggered currentIndexChanged, but avoid triggering it when setting the index back. And in that case, if I'm not mistaken, this reaction only ever needs to happen in PartitionPage::updateFromCurrentDevice.

tctara:
but even if we can safely store the index update, how do you suggest we differentiate between  user and core-reset triggered currentIndexChanged signal ? 
from what I observe when the core reset happens, the model will be cleared and emit a currentIndexChaged, then it will be repopulated and emit another currentIndexChanged.

teo:
You can use [QComboBox::activated](https://doc.qt.io/qt-5/qcombobox.html#activated), which is also triggered when the current item hasn't changed, but only when the interaction comes from the user; as opposed to currentIndexChanged which also happens on programmatic change.

tctara:
After looking more into it, I find that

1.)  After the core reset, when currentIndexChanged signal is emitted, (which is when we want to check if the comboBox is dirty, and set the index accordingly) the model of comboBox is only partially populated (only the first one). That means we can't set the index yet. we have to somehow wait until the model is fully populated then we can set it to the user-selected index. I don't think there's a signal for that :/

2.) activated signal is emitted after currentIndexChanged signal, and activated signal won't be emitted if we programmatically set the index at the currentIndexChanged signal. that leaves us with the same problem of differentiating between the user-triggered and system-triggered signal.

tctara:
Sorry for taking so much of your time. I've forced pushed new changes. please let me know what you think. :)

philmmanjaro:
@tctara: you don't waste anybodies time here. Teo is just giving some free advice on how to improve the code. It is the normal way to only merge the best code if possible. At least it should make sense.

teo:
@tctara what @philmmanjaro said. Good code takes time, learning takes time. I'm helping you out because I want to, so don't apologize.

Your last patch is a huge improvement, nice work!
I can only see one outstanding issue: if you do a full revert (the Revert All Changes button), no bootloader index gets selected. You can probably handle this in `PartitionPage::onRevertClicked`, in the post-run callback which is currently empty on line 230. If a current index is >= 0 then set that, otherwise just set the current to 0.

tctara:
@philmmanjaro, @teo: Thank you. I'll keep them in mind. :)

I thought that by not selecting anything after the revert, we force the user to notice or explicitly select a bootloader again and to avoid  installing the wrong bootloader. but, that could  cause a great deal of confusion and error as well.
Now, I've updated the patch with @teo  suggestion. I hope everything is okay.

PR #243, opened 2016-05-23
Support the already mounted loop fs (example in the blackPanther OS)

blackPantherOS:
**The patch provide this:**

```
--- calamares/src/calamares/modules.orig/unpackfs/main.py       2016-04-28 16:37:21.000000000 +0200
+++ calamares/src/calamares/modules/unpackfs/main.py    2016-05-22 14:50:19.000000000 +0200
@@ -198,7 +198,10 @@
         :param entry:
         :param imgmountdir:
         """
-        subprocess.check_call(["mount", entry.source, imgmountdir, "-t", entry.sourcefs, "-o", "loop"])
+        if os.path.isdir(entry.source):
+            subprocess.check_call(["mount", "--bind", entry.source, imgmountdir])
+        else:
+            subprocess.check_call(["mount", entry.source, imgmountdir, "-t", entry.sourcefs, "-o", "loop"])

     def unpack_image(self, entry, imgmountdir):
         """ Unpacks image.
@@ -278,7 +281,8 @@

         destination = os.path.abspath(root_mount_point + entry["destination"])

-        if not os.path.exists(source) or os.path.isdir(source):
+        if not os.path.exists(source):
+        # Modified for blackPanther OS support (www.blackpantheros.eu)
             return "Bad source", "source=\"{}\"".format(source)

         if not os.path.isdir(destination):

```
**After this patch the calamares working with this format in unpackfs.conf:**

>  source = /path/to/already/mounted/loopfs

And provide the Hungarian blackPanther support too...


philmmanjaro:
@blackPantherOS: I don't see any valid pull request here. So something went wrong on your end. The given commit for **Exclude zram from devices list** is already in master branch. Then current **unpackfs** module is designed for **Arch-based** distros like BBQ-Linux and Manjaro. Especially the second part of your patch:

```
-        if not os.path.exists(source) or os.path.isdir(source):
+        if not os.path.exists(source):
+        # Modified for blackPanther OS support (www.blackpantheros.eu)
             return "Bad source", "source=\"{}\"".format(source)
```
... seems to break compatibility with Manjaro. So either you add code without breaking others or fork the module, call it differntly and maintain it distro-specific like we do with [our modules](https://github.com/calamares/calamares-manjaro/tree/master/src/modules).

blackPantherOS:
Yes something wrong. I looked away...
"seems to break compatibility with Manjaro. "
Why? If the unpackfs.conf contain source=/path/to/squashfs.img the patch still support the original operation here:
```
if os.path.isdir(entry.source):
+            subprocess.check_call(["mount", "--bind", entry.source, imgmountdir])
+        else:
+            subprocess.check_call(["mount", entry.source, imgmountdir, "-t", entry.sourcefs, "-o", "loop"])
# THIS SUBPROCESS the ORIGINAL OPERATION AFTER else:
```

philmmanjaro:
Please read my comment. I'm arguing about the second part of your patch which breaks it ...

philmmanjaro:
Please try again if you have a proper patch.

kkofler:
The comment is improperly indented, which is what breaks things. I'd suggest removing the offending comment line altogether, it doesn't make all that much sense there anyway (as it comments something that's not there anymore!). It is sufficient to have this information in the commit message.

teo:
Thanks @kkofler and @philmmanjaro for reviewing this.
@blackPantherOS I'd be ok with merging this, but as @philmmanjaro pointed out is should not break existing behaviour for other downstreams. Can you please remove that improperly indented comment line and submit a pull request with only your changes, without those additional merges?

blackPantherOS:
Sure,  sorry to late reply, we working on the next blackPanther OS release and now there are very very busy and hard days..

PR #244, opened 2016-06-04
Support the already mounted loop fs (example in the blackPanther OS)

hmikihth:
.

### DIFF
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -26,7 +26,6 @@ import shutil
 import subprocess
 import sys
 import tempfile
-import platform
 
 from libcalamares import *
 
@@ -282,7 +281,7 @@ def run():
 
         destination = os.path.abspath(root_mount_point + entry["destination"])
 
-        if not os.path.exists(source) or (os.path.isdir(source) and platform.dist()[0] != "blackPanther"): 
+        if not os.path.exists(source): 
             return "Bad source", "source=\"{}\"".format(source)
 
         if not os.path.isdir(destination):
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -198,7 +198,10 @@ class UnpackOperation:
         :param entry:
         :param imgmountdir:
         """
-        subprocess.check_call(["mount", entry.source, imgmountdir, "-t", entry.sourcefs, "-o", "loop"])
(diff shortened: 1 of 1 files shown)


philmmanjaro:
The new line 285 might break **Fedora** support. Please see also https://github.com/calamares/calamares/commit/ee2ed0921e0b5aa598c91096b6fe7f75d3914e64

hmikihth:
I hope it will be ok, I check the distro in that line.

philmmanjaro:
@hmikihth: much better. Please also remove the comment `# Modified for blackPanther OS support (www.blackpantheros.eu)` as it is not needed.

kkofler:
@philmmanjaro:
> The new line 285 might break **Fedora** support. Please see also https://github.com/calamares/calamares/commit/ee2ed0921e0b5aa598c91096b6fe7f75d3914e64

Why would it? The code is being made more general in every step:
* original: proceed only if the file exists and is a regular file
* ee2ed0921e0b5aa598c91096b6fe7f75d3914e64: proceed only if the file exists and is not a directory (i.e., it can be a regular file or a special device file)
* 377a1c3a1705c850830abdd979a32dc4a093a782: proceed only if the file exists (i.e., it can be anything)

So the followup 062e9b764c9ee5959896f8df582d8e70863eb453 that hardcodes the distro is **not** needed and a step backwards. Please revert that commit.

For what it's worth, I wonder whether this pull request is needed at all, and whether you can't just point Calamares to some device file in `/dev/mapper` as I'm doing in the Fedora config. (The file name in Fedora is `/dev/mapper/live-base`.) But then again, allowing to use an already mounted directory can't really hurt, and I might even switch to using that in the Fedora config at some point.

philmmanjaro:
Ok, then it is fine for me aswell to have https://github.com/calamares/calamares/commit/062e9b764c9ee5959896f8df582d8e70863eb453 reverted if your **Fedora** is working with it. Hardcoding a distro is always no the best way to do it.

hmikihth:
Why you did not merge it?

philmmanjaro:
@teo: what do you say. The code should be now fine. Also @kkofler reviewed it already.

teo:
Thank you @hmikihth for the patch, and apologies for making you wait this long to merge (I hadn't noticed your last commit).

hmikihth:
Thank you @teo

PR #245, opened 2016-06-25
Netinstall module. See README for complete guide. Allows to configure...groups and packages; selected packages are installed through the 'packages' module.

shainer:
**Contributors**: me (shainer) and AlmAck.

**Important**: I have tested that the packages are added correctly on the global storage (even if you revisit your selection later), and that they are not added twice if you visit the page twice.
I did not test that the installation actually takes place, if needed I can do that here, I imagine it will take time to review the whole module anyway :)
UI not that pretty I suspect, ideas welcome.

Unrelated: the Wiki page about the Global Storage is missing a section for the 'packages' key, I'd write one, but it seems like you can't send pull requests for wiki changes.


src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -0,0 +1,139 @@
+#include "NetInstallPage.h"
+
+#include "./widgets/groupselectionwidget.h"
```
teo: Any reason for the leading dot?

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "ui_page_netinst.h"
+#include "GlobalStorage.h"
+#include "JobQueue.h"
+#include "utils/Logger.h"
+
+#include <QFile>
+#include <QMap>
+#include <QTextStream>
```
teo: I don't consider this an issue, just a minor nitpick for style: Qt includes should go after KF5 includes.

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -0,0 +1,139 @@
...
+    // if the user does not select any group nothing is installed
+    return true;
+}
+
+void NetInstallPage::ReadGroups( const QByteArray& yamlData )
+{
+    YAML::Node groups = YAML::Load( yamlData.constData() );
+    Q_ASSERT( groups.IsSequence() );
+
+    for ( YAML::const_iterator it = groups.begin(); it != groups.end(); ++it )
```
teo: Hint: I'm not sure but you might be able to improve this by using YamlUtils.{h,cpp} in libcalamares.

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -0,0 +1,139 @@
...
+            // Do not present on view.
+            continue;
+        }
+
+        GroupSelectionWidget* groupWidget = new GroupSelectionWidget( group.name, group.description, group.packages, this );
+        m_groupWidgets.insert( groupKey, groupWidget );
+        ui->groupswidget->layout()->addWidget( groupWidget );
+
+        mapper->setMapping( groupWidget, groupKey );
+        connect( groupWidget, SIGNAL( toggled( bool ) ), mapper, SLOT( map() ) );
```
teo: I would prefer C++11/Qt5-style connects.

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return selectedPackages;
+}
+
+void NetInstallPage::loadGroupList()
+{
+    QString confUrl(
+        Calamares::JobQueue::instance()->globalStorage()->value(
+            "groupsUrl" ).toString() );
+
+    KIO::Job* getJob = KIO::storedGet( confUrl, KIO::Reload, KIO::Overwrite | KIO::HideProgressInfo );
```
teo: What's the rationale for using KIO::Job instead of QNetworkAccessManager here?

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -0,0 +1,139 @@
...
+    connect ( getJob, SIGNAL(  result( KJob* ) ), this, SLOT( dataIsHere( KJob* ) ) );
```
teo: C++11/Qt5-style connect would be better, as before.

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -0,0 +1,139 @@
+#include "NetInstallPage.h"
```
teo: Missing copyright header.

src/modules/netinstall/NetInstallPage.h
```diff
@@ -0,0 +1,95 @@
+/*
+ *   Copyright 2016, Luca Giambonini <almack@chakraos.org>
+ *                   Lisa Vitolo     <shainer@chakraos.org>
```
teo: Please add "Copyright 2016" on every line.

src/modules/netinstall/NetInstallPage.h
```diff
@@ -0,0 +1,95 @@
...
+
+    // Internal representation of a package group.
+    struct Group
+    {
+        Group()
+            : Group( "","",false, false ) { }
+        Group( QString name, QString description, bool selected, bool hidden )
+            : name( name ), description( description ), selected( selected ), hidden( hidden ) { }
+        Group( QString name, QString description )
+            : Group( name, description, false, false ) { }
```
teo: I like this. Props for using an awesome but often overlooked C++11 feature.

src/modules/netinstall/NetInstallPage.h
```diff
@@ -0,0 +1,95 @@
...
+        QString description;
+        QStringList packages;
+
+        // See README.md for a description of these two fields.
+        bool selected = false;
+        bool hidden = false;
+    };
+
+public:
+    NetInstallPage( QWidget* parent = 0 );
```
teo: nullptr is better than 0

src/modules/netinstall/NetInstallPage.h
```diff
@@ -0,0 +1,95 @@
...
+    void dataIsHere( KJob* );
+
+signals:
+    void checkReady( bool );
+
+private:
+    // Takes the YAML data representing the groups and reads them into the
+    // m_groups and m_groupOrder internal structures. See the README.md
+    // of this module to know the format expected of the YAML files.
+    void ReadGroups( const QByteArray& yamlData );
```
teo: 1) method identifiers should start with a lowercase character
2) as mentioned earlier, **maybe** this can be streamlined by using YamlUtils - up to you

src/modules/netinstall/NetInstallPage.h
```diff
@@ -0,0 +1,95 @@
...
+
+    Ui::Page_NetInst* ui;
+
+    QHash<QString, Group> m_groups;
```
teo: I'd like an inner space within angle brackets, e.g. `QHash< QString, Group >`

src/modules/netinstall/NetInstallViewStep.cpp
```diff
@@ -0,0 +1,151 @@
+/*
+ *   Copyright 2016, Luca Giambonini <almack@chakraos.org>
+ *                   Lisa Vitolo <shainer@chakraos.org>
```
teo: "Copyright 2016" on every line, as before.

src/modules/netinstall/NetInstallViewStep.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    emit nextStatusChanged( true );
+    connect( m_widget, &NetInstallPage::checkReady,
+             this, &NetInstallViewStep::nextStatusChanged );
+}
+
+
+NetInstallViewStep::~NetInstallViewStep()
+{
+    if ( m_widget && m_widget->parent() == nullptr )
```
teo: Not sure whether this will ever be true, but I guess it doesn't hurt.

src/modules/netinstall/NetInstallViewStep.cpp
```diff
@@ -0,0 +1,151 @@
...
+        m_widget->deleteLater();
+}
+
+
+QString
+NetInstallViewStep::prettyName() const
+{
+    return tr( "NetInstall" );
```
teo: This is the user-visible name, maybe something more descriptive? "Select packages" or "Package selection" or somesuch?

src/modules/netinstall/NetInstallViewStep.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    m_widget->onActivate();
+}
+
+
+void
+NetInstallViewStep::onLeave()
+{
+    cDebug() << "Leaving netinstall, adding packages to be installed"
+             << "to global storage";
```
teo: Have you tested what happens if the user goes next from NetInstall, then back and then next again?

src/modules/netinstall/NetInstallViewStep.h
```diff
@@ -0,0 +1,73 @@
+/*
+ *   Copyright 2016, Luca Giambonini <almack@chakraos.org>
+ *                   Lisa Vitolo     <shainer@chakraos.org>
```
teo: "Copyright 2016" on every line, as before.

src/modules/netinstall/NetInstallViewStep.h
```diff
@@ -0,0 +1,73 @@
...
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef NETINSTALLVIEWSTEP_H
+#define NETINSTALLVIEWSTEP_H
+
+#include <QObject>
```
teo: You don't need this, ViewStep.h already includes QObject.

src/modules/netinstall/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+The netinstall module allows distribution maintainers to ship minimal ISOs with only a basic set of preinstall packages. At installation time, the user is presented with the choice to install groups of packages from a predefined list.
+
+Calamares will then invoke the correct backend to install the packages.
+
+## How are packages configured?
+The *netinstall.conf* file should have this format:
+
+    ----
+    groupsUrl: <URL to YAML file>
```
teo: Any reason why the YAML file must be remote? Could it be a local YAML file, or even a YAML object included in netinstall.conf?

src/modules/netinstall/widgets/groupselectionwidget.cpp
```diff
@@ -0,0 +1,48 @@
+#include "groupselectionwidget.h"
```
teo: Missing copyright header.

src/modules/netinstall/widgets/groupselectionwidget.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <QtDebug>
+
+GroupSelectionWidget::GroupSelectionWidget( QString name, QString description, QStringList packages, QWidget* parent ) :
+    QWidget( parent ),
+    m_isToggled( false )
+{
+    ui.setupUi( this );
+
+    connect( ui.group, SIGNAL( toggled( bool ) ), this, SLOT( toggleGroup( bool ) ) );
```
teo: C++11/Qt5 connect syntax would be preferable.

src/modules/netinstall/widgets/groupselectionwidget.h
```diff
@@ -0,0 +1,49 @@
+/*
+ *   Copyright 2016, Luca Giambonini <almack@chakraos.org>
+ *                   Lisa Vitolo     <shainer@chakraos.org>
```
teo: "Copyright 2016" on every line.

src/modules/netinstall/widgets/groupselectionwidget.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <QSignalMapper>
+#include <QWidget>
+
+// Represents a widget to display and select a group.
+class GroupSelectionWidget : public QWidget
+{
+    Q_OBJECT
+public:
+    explicit GroupSelectionWidget( QString name, QString description, QStringList packages, QWidget* parent = 0 );
```
teo: nullptr instead of 0

teo:
@shainer thank you for the pull request, excellent work. Please see inline comments.
Some testing to ascertain whether a complete install with package selection works would be good.
As for documentation on the packages key in GlobalStorage, feel free to add it directly to the wiki with no pull request.

src/modules/netinstall/NetInstallPage.cpp
shainer: No, I think it was just a leftover. Removed for consistency.

src/modules/netinstall/README.md
shainer: My reasoning was that if the file is local, then changing even the name of one package or adding a new one requires a rebuild for the distro maintainer. I imagine both actions will be performed pretty often.

However I am thinking that reading from a remote location has potential security issues; we would need, at least, to avoid displaying any string above a certain size, and make sure the package names are not malformed (although the threat of remote command execution may be a bit far-fetched). Thoughts? :)

shainer: Or, keep the sanity checks above just to avoid unintentional errors, but require the config file to be signed. This is still a maintenance burden but lower than a rebuild.

Ram-Z: Having a remote file listing the packages will reduce the number of respin that need to be done because a hardcoded pkg name on the ISO doesn't exist anymore and the ISO has become useless.

Maybe the file should be signed and Calamares should verify that signature?

teo: I'm not just thinking security, I'm thinking that some distros might like to ship packages on the ISO to allow a customizable and "fat" offline install, and an offline configuration would be useful for this.

src/modules/netinstall/NetInstallPage.cpp
kkofler: FYI, this is not just a consistency thing. Those leading dots can actually break RPM builds (though in this case, probably not, because the path is actually "./widgets" and not just "."). They end up in `#line` directives to the assembler, from there in the debugging sections of the object file, and then you end up with a path of "." in the path table, which `debugedit` wants to canonicalize to the empty string "", but the hack `debugedit` uses to shrink strings breaks down if the string shrinks by exactly 1 character, so it fails with an error.

src/modules/netinstall/README.md
kkofler: As far as I know, KIO will actually handle a `file://` URL just fine. QNetworkAccessManager apparently does, too, as the description for `networkAccessible` says "If the network is not accessible [...] Requests with URLs with the file:// scheme will still be processed."

demmm: Two  modules already work/mix with url to local file, well tested and working are the webview & license module, examples:
https://github.com/KaOSx/calamares/blob/master/src/modules/license/license.conf#L10
https://github.com/KaOSx/calamares/blob/master/src/modules/webview/webview.conf

src/modules/netinstall/NetInstallPage.cpp
shainer: I did not know about it. I have tried to use it, but for some reason even if setting the max. redirects allowed to 3, it does not follow the HTTP 301 redirect for Chakra's URL. There are several complaints online about this behaviour sadly :)

shainer:
For testing, do I need to boot a live system and call deploycala.py? Can I test the package installation without installing the rest of the live system too?

For the wiki, it seems I don't have permissions: there is no option to edit pages anywhere. From Github docs: "By default, only collaborators on your repository can make changes to wikis, but you can configure this to be enabled for all users on public repositories.".

philmmanjaro:
Regarding the wiki, I think on the current level only **[developers](https://github.com/orgs/calamares/teams/developers)** should edit that, not the public. However, it should be easy to add you there. Simply keep the merge process for reviewing new modules as given.

teo:
@shainer should be able to edit the wiki now.

You may use deploycala.py to test, but I don't know whether it will work on your distro. Currently it makes pacman or yaourt calls, so it works on Manjaro and possibly other Arch based systems. You can probably fix it up to work on any distro quite easily. And if not with deploycala.py, you can also set up a build environment manually on the live system and then test.

Since package installation is supposed to happen in the target system, I can't see a way to test it well without performing an installation.

shainer:
Thanks for the wiki access, I added the section ;)

I managed to build and launch Calamares on a live distro, however it seems the 'packages' module is never run during the installation. I added it to the exec: list in settings.conf, but I think I need to do something else?

shainer:
Back! I am *almost done* configuring & testing the module on a Chakra Linux live ISO.

I run the _packages_ module after _unpackfs_, and it runs the correct pacman command inside the chroot with the root image unpacked. The only issue is that the command then failed because there is no valid resolv.conf in the chroot, so no network. I have modified resolv.conf and launched manually the same command run by calamares and it works; I am pretty confident it will work at this point, but I am going to see if we can retest it with a valid root image.

On a different note, deploycala.py does not update a few dependencies (kpmcore and most of qt5?), I should have a pretty complete list written down, I can update the script (is it on Git somewhere?).

teo:
Splendid!
`deploycala.py` is in https://github.com/calamares/calamares.github.io/ so it all gets deployed to calamares.io immediately upon pushing.

shainer:
Ok I had networkcfg copy resolv.conf (the current code does not do it, and it did not setup a working network) and now the module works. I will update:

- deploycala.py
- netinstall/README.md with more complete instructions.
- the networkcfg module (in a separate request, not here).

There was no change required to the netinstall code here, so this patch can go as it is.

blackcrack:
would also nice if works with urpmi .. or plain rpm via chroot and then install .
the yum and other installer like urpmi, abt get, dpkg and should be as variable  and sources of the packets should be it also for changing the in the Script by script start and to be more possibility for setting up/adjusting the script for different installer and using different listings with dependency solution by single selecting . For more different Distributions :)

best regards
Blacky

kkofler:
As I understand it, the actual installation work is done by the `packages` module, so the code should work with any package manager supported by that module, which already includes `urpmi`.

AlmAck:
Correct kkofler, the `netinstall` module does not manage the installation of the packages directly but instead it generates a list of packages to be installed. This list is then shared with the `packages `module through the GlobalStorage variables of calamares.

@blackcrack you can always use the `packages `module to install your custom packages. In Chakra we have a special squashfs that we mount in the live env. that contains some binary packages that we install only in the target system.

blackcrack:
Ciao, AlmAck :))
mille gracie for the explanation :)

src/modules/netinstall/NetInstallPage.cpp
teo: As of Qt 5.6, there is the attribute enum value `QNetworkRequest::FollowRedirectsAttribute` to make your QNetworkReply follow redirects. Could this work for you?
I'm asking because KIO drags in a bucketload of dependencies that distros based on something other than Plasma might not want. If there is no reasonable way to make it work without KIO, let's use KIO, but I'd like for all of us to be sure about QNetworkAccessManager & friends not being up to the task.

src/modules/netinstall/README.md
```diff
@@ -26,3 +29,33 @@ The file is composed of a list of entry, each describing one group. The keys *na
...
+in the chroot.
+
+## Common issues
+If launching the package manager command returns you negative exit statuses and nothing is actually invoked, this
+is likely an error in the setup of the chroot; check that the parameter **rootMountPoint** is set to the correct
+value in the Calamares configuration.
+
+If the command is run, but exits with error, check that the network is working in the chroot. Make sure /etc/resolv.conf
+exists and that it's not empty.
+
```
teo: We might want to point out that the netinstall module won't work if there's no internet connection. In that case, internet connectivity should be a mandatory requirement in `welcome.conf`.

src/modules/netinstall/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+#include(KDEInstallDirs)
+#include(KDECMakeSettings)
+#include(KDECompilerSettings)
```
teo: Do we need these?

teo:
Ok, I think we're getting really close to merging. Awesome work!
Minor nitpicks aside, I'd appreciate if you could please take another look at QNetworkAccessManager with the suggestion I just left inline.

shainer:
Dependency on KIO removed ;) verified redirects are handled.

src/modules/netinstall/CMakeLists.txt
```diff
@@ -1,19 +1,5 @@
...
-list( APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules" )
-
-find_package(ECM 1.0.0 REQUIRED NO_MODULE)
-list( APPEND CMAKE_MODULE_PATH ${ECM_MODULE_PATH} ${ECM_KDE_MODULE_DIR})
-
-set(KF5_MIN_VERSION "5.2.0")
-find_package(KF5 ${KF5_MIN_VERSION} REQUIRED COMPONENTS
-    KIO             # KIO
-)
-
```
teo: This is so fucking awesome!

teo:
Thank you @shainer and @AlmAck for the patch!

PR #247, opened 2016-07-04
packageOperations should be a map between operation type and list of …

shainer:
…packages. It does not make sense for it to be a list of dictionaries, all with the same format. Also makes conversions with QVariant harder.

Moreover, add conversion from/to pydict and QHash, mirroring the existing implementation for QMap. Nice addition :)

Found these while testing the netinstall module.

### DIFF
--- a/src/libcalamares/PythonHelper.cpp
+++ b/src/libcalamares/PythonHelper.cpp
@@ -29,7 +29,8 @@
 
 namespace bp = boost::python;
 
-namespace CalamaresPython {
+namespace CalamaresPython
+{
 
 
 boost::python::object
@@ -39,6 +40,9 @@ variantToPyObject( const QVariant& variant )
     {
     case QVariant::Map:
         return variantMapToPyDict( variant.toMap() );
+
+    case QVariant::Hash:
+        return variantHashToPyDict( variant.toHash() );
 
     case QVariant::List:
     case QVariant::StringList:
@@ -94,9 +98,7 @@ variantListToPyList( const QVariantList& variantList )
 {
     bp::list pyList;
     foreach ( const QVariant& variant, variantList )
-    {
         pyList.append( variantToPyObject( variant ) );
-    }
     return pyList;
 }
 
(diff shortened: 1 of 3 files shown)


src/libcalamares/PythonHelper.cpp
```diff
@@ -165,20 +197,16 @@ Helper::Helper( QObject* parent )
...
         if ( fi.exists() && fi.isReadable() )
-        {
             m_pythonPaths.append( fi.dir().absolutePath() );
-        }
 
         QDir calaPythonPath( CalamaresUtils::systemLibDir().absolutePath() +
                              QDir::separator() + "calamares" );
         if ( calaPythonPath.exists() &&
-             calaPythonPath.isReadable() )
+                calaPythonPath.isReadable() )
```
teo: Any reason for this whitespace change?

src/libcalamares/PythonHelper.cpp
```diff
@@ -94,9 +98,7 @@ variantListToPyList( const QVariantList& variantList )
 {
     bp::list pyList;
     foreach ( const QVariant& variant, variantList )
-    {
```
teo: I kinda liked these braces even for 1-line blocks. Keeps everyone from making mistakes if adding lines at some point.

src/modules/packages/main.py
```diff
@@ -124,9 +124,6 @@ def run():
         run_operations(pkgman, entry)
 
     if libcalamares.globalstorage.contains("packageOperations"):
-        operations = libcalamares.globalstorage.value("packageOperations")
-
-        for entry in operations:
-            run_operations(pkgman, entry)
+        run_operations(libcalamares.globalstorage.value("packageOperations"))
```
teo: This won't work as long as run_operations requires a package manager as first argument.

src/libcalamares/PythonHelper.cpp
shainer: Both this and the whitespace change were done by astyle (the script in hacking/). We should fix that script :)
I'll correct the Python invocation soon.

src/modules/packages/main.py
shainer: Thanks, done.

PR #248, opened 2016-07-09
networkcfg needs to copy resolv.conf in the chroot

shainer:
During my tests, the network in the chroot was not working without this.


AlmAck:
You added also the netinstall module in this request, is correct?

shainer:
No :D I must have branched from the wrong 'parent'...let me redo it. Thanks.

PR #249, opened 2016-07-10
networkcfg needs to copy resolv.conf in the chroot

shainer:
Same as pull request #248, but from the correct branch so it does not include my unrelated commits.

### DIFF
--- a/src/modules/networkcfg/main.py
+++ b/src/modules/networkcfg/main.py
@@ -53,4 +53,13 @@ def run():
             except FileExistsError:
                 pass
 
+    # We need to overwrite the default resolv.conf in the chroot.
+    source_resolv = "/etc/resolv.conf"
+    target_resolv = os.path.join(root_mount_point, "etc/resolv.conf")
+    if os.path.exists(source_resolv):
+        try:
+            shutil.copy(source_resolv, target_resolv)
+        except FileNotFoundError:
+            libcalamares.utils.debug("Can't copy resolv.conf from {}".format(source_resolv))
+
     return None


src/modules/networkcfg/main.py
```diff
@@ -53,4 +53,13 @@ def run():
             except FileExistsError:
                 pass
 
+    # We need to overwrite the default resolv.conf in the chroot.
+    source_resolv = "/etc/resolv.conf"
+    target_resolv = os.path.join(root_mount_point, "etc/resolv.conf")
+    if os.path.exists(source_resolv):
```
teo: I would like this to happen only if `target_resolv` doesn't exist already, so we don't go overwriting something that a distro might want to ship as a default in the rootfs.

shainer: Most distros I have seen ship an empty file with a few comments, so a check like that would not work.

Shipping a "default" DNS configuration with the root image seems weird. A lot of people set their own DNS anyway, especially if they are on a LAN or similar networks. In short, I think whatever the user has configured on the live system is good as a source of truth, since it is their responsibility to ensure it works at that point, not the distro maintainer's.

If we want to cover the uncommon case of default resolv.conf in the root image we can add a "don't overwrite resolv.conf" parameter in the module settings.

blackcrack: i guess if anybody use an onlineinstall, it is he in able to configurate his own network in any case
and know what it is ip, gateway,dns, boadcast ans so on, there should be a possibility for fully configurating the networkstack by activate the onlineinstall (maybe with a chackbox ?)
All peoples, who use onlineinstall, should know what they do and with the 
configuration of the own Network it is prove if the user in able,
other hands , well, the default configuration can be a dns server and the network
start up via autoconfiguration .. with the possible for configuration the network manager network configuration for the System and copy this configuration directly into the system.. (chackra and the other take also networkmanager ? OpenMandriva does it and you can configuring it with nmtui..)
So, imho should it be possible for configuring the network by install if a checkbox checked (maybe [X] Network Configuration) with different Computer.name,Aliasname,  networkcards(if more as one exist), IP's gateway's, different DNS and so on, what's all need... also for a server, for this a checkbox for [X]Extended Network configuration and a possibility for configuration the whole Netowork.
By the way, thank you very much for doing this Lisa and Teo and the others ! :))))) *hugethankyouverymutch*
and i guess i be not alone with this feelings ! Finally back to a current installation with actual (rpm)packs..*dribblethumpsup*

best regards
Blacky

teo: Fair enough. I think it doesn't matter much either way unless one does static configuration, since NetworkManager happily messes around with `resolv.conf` anyway.

PR #250, opened 2016-08-12
improve portage backend, so that properly removes reverse deps, whitout the need of HUGE package list in packages module configuration

V3n3RiX:
emerge -C means remove package without checking for reverse dependencies. So in order to properly remove calamares and ALL it's reverse dependencies during system installation, one must provide a rather long list in packages module configuration.

faster way to do this is using emerge --depclean -q. It will check for reverse dependencies of packages removed with emerge -C and will clean the system. 

this line was inspired by apt backend, that uses a similar approach


teo:
Thank you @V3n3RiX

PR #251, opened 2016-08-19
displaymanager: write processed DM list to globalstorage

udeved:
displaymanager: write processed DM list to globalstorage so other modules can use it.

Perhaps a detection module would make sense later on, that only writes to globalstorage and collects information, eg DM or init system.
Not sure the required line in module.desc can be used to create depends? Our postcfg module consumes new global storage DM @teo


teo:
@udeved `required` currently does nothing.

PR #252, opened 2016-08-19
grubcfg: add option to set plymouth theme

udeved:
should fit best here? plymouth is already detected in grubcfg
I am just not sure about a default theme name.


src/modules/grubcfg/main.py
```diff
@@ -43,6 +45,9 @@ def modify_grub_default(partitions, root_mount_point, distributor):
 
     if plymouth_bin == 0:
         use_splash = "splash"
+        
+        plymouth_theme = libcalamares.globalstorage.value("plymouth_theme")
```
teo: I think you mean `libcalamares.job.configuration`, not `globalstorage`. Also, since this is an optional configuration entry, you should check `if "plymouth_theme" in libcalamares.job.configuration` before accessing it.

udeved: Yeah, will add a check. 
But it can stay in grubcfg?

philmmanjaro:
I don't know why we should configure **plymouth theme** via **grubcfg** at all ...

udeved:
Because it already has plymouth detection. But it may be better to have it somewhere else altogether?

udeved:
I can do a patch to merge it in one commit if you want. What a disaster.

src/modules/grubcfg/main.py
```diff
@@ -23,6 +23,8 @@
 import os
 import re
 
+def setExpression(pattern, file):
+        libcalamares.utils.target_env_call(["sed", "-e", pattern, "-i", file])
```
teo: I think you might have an alignment issue caused by your IDE? We use 4 spaces and no tabs in Calamares.

src/modules/grubcfg/grubcfg.conf
```diff
@@ -11,3 +11,6 @@ defaults:
     GRUB_DISABLE_SUBMENU: true
     GRUB_TERMINAL_OUTPUT: "console"
     GRUB_DISABLE_RECOVERY: true
+
+# The plymouth theme to be set if plymouth binary is present
+plymouth_theme: manjaro-elegant
```
teo: Is there any other theme that comes with upstream plymouth that we can use as default?

udeved:
@teo 
We could also deliver grubcfg with only commented option. I myself do not use plymouth, I think its only a source of trouble, so I have no idea what default theme comes with it,but I recall default is ugly.

Regarding IDE, what do you use for CAL? I settled with eclipse for the moment until kdevelop 5 is stable enough.

kkofler:
IMHO, the default should just be to not set anything. Fedora already sets the default theme in Plymouth packaging, so Calamares doesn't have to do anything. I'd expect most distros to be set up that way.

udeved:
Do to our tons of different editions, we would need some customization.
The code currently makes it optional.

highvoltage:
+1 on what kkofier said, usually that's something that a distro or edition does in packaging.

udeved:
@philmmanjaro 

This would mean our plymouth guys would have to set theme by package, which is fine with me.

philmmanjaro:
So we NAK this all together?

udeved:
I don't mind either way, I just don't want to re implement what already does the job, besides the fact we should indeed let packages set some stuff, but this requires rebuilding arch packages. We have to crazy customization what should be done at package level, my opinion.

philmmanjaro:
Ok, if we merge it here, however, we have to fix line 27 to four spaces instead of tabs and the default theme is most likely **spinner**.

udeved:
If you want to merge, perhaps squeeze it in one patch instead.

philmmanjaro:
We only care about the final result and if other upstream developers agree with it. As long as it is optional, I personally see it now merge-able ...

udeved:
When the cat ran over the keyboard, and I was just writing commit, the commit msg reads in suitable fashion "sinner" instead of spinner. :)

teo:
So from what I can tell, some prefer to configure the plymouth theme within the packaging, and others would like to set it from Calamares. Either way there's no reason to do it within the grubcfg module.
@udeved if you need to set it from within Calamares, how about you do this in a separate Python module? This way, grubcfg only deals with grub (as it should), and everyone else is free to use or not use a plymouth Calamares module.

udeved:
@teo 

I got even better idea, I agree, plymouth doesn't fit in grubcfg.

So what if we make a plymouth module, which writes hasPlymouth to globalstorage?
The module would be run before grub?

udeved:
So grubcfg code would be something like:

```python
if libcalamares.globalstorage.value("hasPlymouth"):
        use_splash = "splash"
```

udeved:
We could also do it the other way round, and grubcfg writes to globalstorage.

src/modules/grubcfg/main.py
```diff
@@ -35,13 +34,10 @@ def modify_grub_default(partitions, root_mount_point, distributor):
...
     default_grub = os.path.join(default_dir, "grub")
     distributor_replace = distributor.replace("'", "'\\''")
-    plymouth_bin = libcalamares.utils.target_env_call(["sh", "-c", "which plymouth"])
     use_splash = ""
     swap_uuid = ""
 
-    libcalamares.utils.debug("which plymouth exit code: {!s}".format(plymouth_bin))
-
-    if plymouth_bin == 0:
+    if libcalamares.globalstorage.value("hasPlymouth"):
```
teo: Best if you check if `libcalamares.globalstorage.contains("hasPlymouth")` before getting its value, in case the distro isn't using the plymouthcfg module.

udeved: Ok, no problem, but I wondered if we should do it other way round? let grubcfg detect and write in globalstorage?

src/modules/plymouthcfg/main.py
```diff
@@ -0,0 +1,61 @@
...
+    def root(self):
+        return self.__root
+
+    def setTheme(self):
+        plymouth_theme = libcalamares.job.configuration["plymouth_theme"]
+        target_env_call(["sed", "-e", 's|^.*Theme=.*|Theme=' +
+                         plymouth_theme + '|', "-i", "/etc/plymouth/plymouthd.conf"])
+
+    def detect(self):
+        plymouth_bin = target_env_call(["which", "plymouth"])
```
teo: You have not tested this code. `target_env_call` returns an exit code, what you need here is `check_target_env_output`.

teo: Sorry, I just realized you only needed an exit code, intentionally. Got confused by the use of `format` below. My apologies, please ignore this.

teo: `plymouth_bin` to me speaks "path of the plymouth binary", maybe it should be called `plymouth_found` instead?

udeved: I tested, one with and one without plymouth.
However, I removed the "sh -c" parameter from original call, which might have been a mistake, since I guess sh returns exit code.

udeved: @teo 

I had "isPlymouth" in mind instead of "plymouth_bin", either way, for readability you are right, the name should better reflect it.

philmmanjaro:
**grubcfg** might be used, however **plymouthcfg** not always. Other way around might be useful in that case. If you want to configure **plymouth** then you might use the extramodule anyway. If not **grubcfg** writes a value which is not used. That would better work as it won't break anything. If we keep **plymouthcfg** detecting we should make sure if the module is not used **grubcfg** is not broken.

udeved:
Right @philmmanjaro 

We have atm a module depend on plymouthcfg, while it doesn't hurt if it was default in settings.conf, we may clean it up and let plymouthcfg consume hasPlymouth instead.

Just installed kdevelop 5 and purged eclipse.

philmmanjaro:
@udeved: Teo contacted me today via Skype to discuss this pull request. I assumed that we are using it somehow. I'd suggest to add the given advises and Teo will merge it. Depending on our custom modules if they are also usable by other distros we might create a pull request for them in a later state also. It is always the right way to provide enhancements to given upstream modules.

udeved:
Right, will clean it up.
I guess the servicescfg module would be useful to gentoo as well.

src/modules/grubcfg/main.py
philmmanjaro: some like this:
```
if libcalamares.globalstorage.contains("hasPlymouth"):
    plymouth_found = libcalamares.globalstorage.value("hasPlymouth")
...
if plymouth_found:
```

src/modules/plymouthcfg/main.py
philmmanjaro: `sh -c` includes the exit code. I suggest to add it back. Also we should test-drive this code.

src/modules/grubcfg/main.py
udeved: Yes, sure, I'll add Teo's suggestion for the check, just let the confusion and buggy dbus dust settle. ;)

src/modules/plymouthcfg/main.py
udeved: To my knowledge on the CAL python api, target_env_call deliveres already exit code, its why I thought I can remove the sh -c arg.

philmmanjaro: Well, we have to test it then ...

philmmanjaro:
@udeved: what is the current status on this one? See also [MJR#FS30](https://bugs.manjaro.org/index.php?do=details&task_id=30).

src/modules/plymouthcfg/plymouthcfg.conf
```diff
@@ -0,0 +1,4 @@
+---
+# The plymouth theme to be set if plymouth binary is present
+# leave commented if packaged default theme should be used
+# plymouth_theme: spinner
```
philmmanjaro: Default theme is mostly called **spinfinity**.

philmmanjaro:
I think this should patch it up: https://github.com/calamares/calamares-manjaro/commit/4fd18cb9d0a8ee32f49782d9f34edf07c53faee8

src/modules/plymouthcfg/main.py
kkofler: I don't understand the point of using "`sh -c`". If `target_env_call` delivers the exit code, it is redundant. If `target_env_call` does not deliver the exit code, it won't help, because then it won't deliver the exit code of `sh` any more than the one of `which`. So in that case, it is also redundant.

philmmanjaro: @kkofler: it works also without `sh -c`. Final code won't have it.

philmmanjaro:
It would be better to rebase this pull request. Otherwise it has to much noise for code-review.

udeved:
I'd be so much better if calamares-manjaro was a fork. We can also close this PR and make new one.

philmmanjaro:
New PR created via https://github.com/calamares/calamares/pull/256

udeved:
The workflow is not good. We should have a full source tree for manjaro so we can build the project in IDE, and also package much easier. All we would have to do is auto synch fork.

philmmanjaro:
That might help for **downstream** development, however makes **upstream** development a little different ...

PR #254, opened 2016-08-29
Changes to fstab module and addition of initramfscfg module to support luks/FDE on Debian-based distros

mckinney-subgraph:
This pull request contains the following changes:
1- Patches to the fstab module to make luks/FDE work on Debian-based distros. These were needed because the current code was failing to generate correct entries for the `/etc/crypttab` file. I have not tested this at all with Arch-based or other distros so I don't know it is will break existing setups. 

2- I also added an entire module (initramfscfg) to handle the pre-configuration of the initramfs vs update-initramfs (in the `initramfs`) module. When FDE is enabled, this module will copy a hook and the `/etc/crypttab` file into the initramfs that will be generated by update-initramfs. 

The caveat to using this module successfully in Debian live-based distros is that the packages module must be configured to removed '^live-*' packages so that update-initramfs can run properly, otherwise the initramfs hook will never run. The following `packages.conf` should suffice:
```
backend: apt

operations:
    - remove:
        - '^live-*'
```

Lastly, this does not handle the swap case. That is on my TODO list.

### DIFF
--- a/src/modules/fstab/main.py
+++ b/src/modules/fstab/main.py
@@ -152,20 +152,22 @@ class FstabGenerator(object):
         if not mapper_name or not luks_uuid:
             return None
 
-        if mount_point == "/":
+        if mount_point != "/":
             return None
 
         return dict(
             name=mapper_name,
             device="UUID=" + luks_uuid,
             password="/crypto_keyfile.bin",
+            options="luks,keyscript=/bin/cat",
         )
 
     def print_crypttab_line(self, dct, file=None):
         """ Prints line to '/etc/crypttab' file. """
-        line = "{:21} {:<45} {}".format(dct["name"],
+        line = "{:21} {:<45} {} {}".format(dct["name"],
                                         dct["device"],
                                         dct["password"],
+                                        dct["options"],
                                        )
 
         print(line, file=file)
--- a/src/modules/initramfscfg/encrypt_hook
+++ b/src/modules/initramfscfg/encrypt_hook
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+  PREREQ=""
+
+  prereqs()
+  {
+          echo "$PREREQ"
+  }
+
+  case $1 in
+  # get pre-requisites
+  prereqs)
+          prereqs
+          exit 0
+          ;;
+  esac
+
+  . /usr/share/initramfs-tools/hook-functions
+  if [ -f /crypto_keyfile.bin ]
+  then
+  	cp /crypto_keyfile.bin ${DESTDIR}
+  fi
+  if [ -f /etc/crypttab ]
+  then
+  	cp /etc/crypttab ${DESTDIR}/etc/
+  fi
--- a/src/modules/initramfscfg/main.py
+++ b/src/modules/initramfscfg/main.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# === This file is part of Calamares - <http://github.com/calamares> ===
+#
+#   Copyright 2014, Rohan Garg <rohan@kde.org>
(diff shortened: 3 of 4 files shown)


src/modules/fstab/main.py
```diff
@@ -152,20 +152,22 @@ def generate_crypttab_line_info(self, partition):
         if not mapper_name or not luks_uuid:
             return None
 
-        if mount_point == "/":
+        if mount_point != "/":
```
teo: The `/` mount point is specifically excluded from `crypttab` because its unlocking is handled early on. Not sure why you're doing the opposite here?

src/modules/fstab/main.py
```diff
@@ -152,20 +152,22 @@ def generate_crypttab_line_info(self, partition):
...
             return None
 
         return dict(
             name=mapper_name,
             device="UUID=" + luks_uuid,
             password="/crypto_keyfile.bin",
+            options="luks,keyscript=/bin/cat",
```
teo: I'm ok with this, though I don't fully understand what it's for. Is it necessary on Debian?

src/modules/fstab/main.py
mckinney-subgraph: What I found is that the lines are never populated when the `mount_point == "/"` is checked. The commented part crypttab header text is inserted properly but no uncommented lines are ever written because the function returns 'None' before it can return the dictionary. 

In my case, mount_point is always "/". I am not sure what to do about this.

src/modules/fstab/main.py
mckinney-subgraph: This was necessary to automatically unlock the OS filesystem without re-entering a password. All it does is use `/bin/cat` to read the contacts of `/crypto_keyfile.bin` when unlocking the filesystem. It wasn't doing this automatically with the original crypttab entry.

src/modules/fstab/main.py
teo: That's the point though, we don't want the `/` line in `crypttab`. `crypttab` unlocking is only ever supposed to take care of other mount points, like `/home`. If your only mount point is `/`, it is correct for `crypttab` to be empty because `/` is already unlocked early on, before `crypttab` kicks in.

teo: Ouch, I guess unless your Debian actually wants the `/` line in crypttab which then ends up in the initramfs.

teo: In that case, we can completely remove that `if`, I think the `/` line in `crypttab` shouldn't affect behavior on other distros.

src/modules/fstab/main.py
teo: Ok, I understand. Good catch. I must test this on other distros.

src/modules/fstab/main.py
mckinney-subgraph: Sure, removing the `if` clause works. :)

kkofler: +1 to removing the `if` clause, I'm pretty sure we also want the `/` line in `crypttab` for `dracut`. (Please update your patch to do that, the current `!=` clause does not make sense.)

src/modules/fstab/main.py
kkofler: options="luks" should probably be set on all distros. The "keyscript=/bin/cat" thing seems to be Debian-specific (including of course Debian derivatives) though, see:
http://www.pavelkogan.com/2015/01/25/linux-mint-encryption/
which tracks this down to the `cryptroot` script from the Debian `initramfs-tools`.

So I think this will need to be optional. Maybe add a `crypttabOptions: luks` line to fstab.conf, which you can customize to `crypttabOptions: luks,keyscript=/bin/cat` on Debian? (Of course, you also need to add the Python code to actually read the option.)

PR #255, opened 2016-09-01
Port away from most cases of Q_FOREACH to C++11 ranged for loop.

stikonas:
In order to avoid deep copies, Qt containers over which we iterate must be const
The remaining unported cases require qAsConst (Qt 5.7) or std::as_const (C++17)

### DIFF
--- a/src/modules/partition/jobs/ClearMountsJob.cpp
+++ b/src/modules/partition/jobs/ClearMountsJob.cpp
@@ -72,8 +72,8 @@ ClearMountsJob::exec()
     process.start();
     process.waitForFinished();
 
-    QString partitions = process.readAllStandardOutput();
-    QStringList partitionsList = partitions.simplified().split( ' ' );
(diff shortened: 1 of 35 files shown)


src/modules/partition/jobs/ClearMountsJob.cpp
```diff
@@ -246,14 +246,14 @@ ClearMountsJob::tryCryptoClose( const QString& mapperPath )
 }
 
 
-QStringList
-ClearMountsJob::getCryptoDevices()
+const QStringList
```
teo: Can we return non-const here?

stikonas:
According to https://www.kdab.com/goodbye-q_foreach/ rvalues in for loops it have to be const. Alternatively we can do 

const QStringList cryptoDevices = getCryptoDevices();
for (const QString &dev: cryptoDevices)

but this seems more complicated than just adding const to return type.

PR #256, opened 2016-09-10
[plymouthcfg] add module to configure Plymouth

philmmanjaro:
This module is used by **Manjaro Linux** to configure plymouth. It is designed to be used optionally.

* introduces globalstorage variable **hasPlymouth**
* adds new module **plymouthcfg**


udeved:
* module can be used if plymouth is not present in the target root as well
* it will simply do nothing except to tell calamares that no plymouth was detected

PR #257, opened 2016-09-11
Fix spelling mistake

highvoltage:
s/emty/empty.


kkofler:
Why don't you submit the pull request directly from your feature branch (highvoltage/fix-spelling-mistake)? Submitting pull requests from your master is usually a bad idea. In this case, it includes your pull request merge commit, so we would end up with 2 merge commits. And a more critical issue is that if you then push more stuff to master, it will update this pull request with unrelated changes.

kkofler:
Also note that `kdsingleapplicationguard.cpp` is a third-party file.

PR #258, opened 2016-09-12
Use device's immutable copy instead of rescanning.

tctara:


philmmanjaro:
Looks fine for me, @teo  ...

PR #263, opened 2016-09-27
Add support for C++/Qt batch job plugins

kkofler:
These job plugins work similarly to view modules, with the following differences:

* These jobs need to link only `libcalamares`, not `libcalamaresui`. For this reason, `PluginFactory` was moved from `libcalamaresui` to `libcalamares`. (It depends only on `QtCore`.)
* Instead of deriving from `ViewModule`, derive from `CppJob` (which is a subclass of `Job`).
* Like process and Python jobs, a job plugin is a single job, whereas a `ViewModule` can generate a whole list of jobs.

The `CppJob` and `CppJobModule` classes are new. In `Module::fromDescriptor`, the combination _type=job, intf=qtplugin_ is now supported and mapped to `CppJobModule`.

As an example, add a new `dummycpp` C++ job module (ported from `dummypython`).


teo:
This is awesome, thank you @kkofler

PR #267, opened 2016-10-19
dracut : unlock encrypted swap if exists

V3n3RiX:
This will add an extra rd.luks.uuid for encrypted swap if exists and will allow dracut to unlock it.


src/modules/grubcfg/main.py
```diff
@@ -74,6 +78,9 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     if swap_uuid:
         kernel_params.append("resume=UUID={!s}".format(swap_uuid))
 
+    if swap_outer_uuid:
+        kernel_params.append("rd.luks.uuid={!s}".format(swap_outer_uuid))
```
teo: Perhaps this should only happen if using dracut?

PR #270, opened 2016-10-31
Enhanced Users dialog

shainer:
New features:

- multiple users management.
- ability to set different shells based on configuration.
- ability to select avatars from files and then copy them to a target location in the user's home directory. No specific location suggested due to differences between DEs on where to read avatars from. Tested on a KDE system with a PNG avatar (my Github one).

Tested on Chakra Goedel, the new Chakra beta ISO. I added a short README.md listing all the possible config settings.

Features I'd like to add in the future: getting avatars from URLs, (maybe) computing password strength. However, one step at a time :)


philmmanjaro:
@shainer: seems you have to adopt de979f85289b3d0a1cafa2edb32fe5cdc5893960

philmmanjaro:
Currently fails with:

```
make[2]: *** No rule to make target '../src/modules/users/images/arrow-down.png', needed by 'src/modules/users/qrc_users.cpp'.  Stop.
```
Seems you missed to add the images from **netinstall/images** ...

shainer:
Thanks about the commit, I have just included it.

For the images, I tried re-cloning the branch somewhere else and I cannot reproduce your error. Actually the arrow-down and arrow-up images are not needed right now, I copied them over thinking I might need them later. What happens if you remove them from users.qrc?

teo:
Hey @shainer, thank you for your effort. The branch seems to be based on an early point in master, and has some merge conflicts. Could you please rebase it on top of current master?

shainer:
Rebased, there are no conflicts with master right now.

teo:
Looks like there's an extra merge commit now, between two copies of the branch? And a diff full of other stuff that doesn't belong in this branch :(
As advice for the future, the only thing you need in these cases is `git rebase master`, and then resolve conflicts as you go ahead with `git rebase --continue`. This way, you'd update your branch commit by commit to cleanly apply against current master, without those nasty merge commits.

shainer:
I did the rebase but I think I messed up while pushing to the remotes. Anyway if you prefer I can squash all my commits in a single one so you can focus your review on that and ignore the rest.

oberon2007:
I guess I have opened the autologin issue in the wrong place again :wink:
@shainer would you mind have a look [here](https://calamares.io/bugs/browse/CAL-413) ?
Thank you.

shainer:
@oberon2007 sure I'll keep that on my radar. However I will not address/comment on it as part of this request, since this deals only with the GUI part.

kkofler:
@shainer The thing is, that bug is a regression caused by your UI changes.

shainer:
Ops, I had missed the comments at the bottom. I will look into it then, thanks.

kkofler:
The behavior of the `doAutologin` option in `users.conf` still does not match what the original code did. Setting `doAutologin: false` should only set the option to unchecked by default, not disable it as your code does now:
https://github.com/calamares/calamares/blob/7fc4f3b0c389db0c818bb0607febfdaec7a58cc9/src/modules/users/gui/UsersPage.cpp#L183
https://github.com/calamares/calamares/blob/7fc4f3b0c389db0c818bb0607febfdaec7a58cc9/src/modules/users/gui/AddUserDialog.cpp#L83

shainer:
I see, and the default in case of no setting is "checked by default", right?

kkofler:
The default in the code in the current master is actually unchecked by default. There is nothing checking it explicitly if no setting is in the config file, so the setting from the .ui is used, and it is not set there explicitly either, so the default QCheckBox state is used, which is unchecked.

The shipped default `users.conf` sets it to `true` though.

oberon2007:
Thank you @shainer It's working properly now.
I think though, that leaving autologin unchecked by default would be better. In my understanding using a login password should be standard.
Some other things I observe:
- When I navigate from field to field with <Tab>, to get from the 2nd password input field to the next entry, for some reason I have to hit <Tab> three times :wink: 
- The "avatar" choice is grayed out even when **mugshot** with a selection of image files in **/usr/share/pixmaps/faces/** is installed. Is there a way those could be used directly? Or where is calamares expecting them? The other options I don't have available are.
- "login shell" choice isn't available, either. What would be needed in order to make that work?

philmmanjaro:
@oberon2007: please check if our **manjaro-tools** creates the needed [config](https://github.com/calamares/calamares/blob/users/src/modules/users/users.conf) file adjustments. If some expected entries are missing, it is logical that you find some options disabled.

philmmanjaro:
@oberon2007: also keep in mind that @shainer only has KDE as her desktop and it was most likely designed for that DE. If you face issues in other DEs than Plasma5, please be more specific. At least test it also with one of our KDE profiles first. Thx.

oberon2007:
My **users.conf** created looks like this
```
---
defaultGroups:
    - video
    - power
    - disk
    - storage
    - optical
    - network
    - lp
    - scanner
    - wheel
autologinGroup:  autologin
doAutologin:     true
sudoersGroup:    wheel
setRootPassword: false
availableShells: /bin/bash, /bin/zsh
```
So, tow option for login shell should be available. Setting autologin to false should change the default, I assume - will try that.
About the avatars I see now option. I guess that works differently?
Re. the ``Tab`` navigation I can't imagine that the DE environment will make any difference, though. Might be the case with the avatars, of course.

oberon2007:
Just verfied about ``doAutologin:     false``. That has indeed the desired effect. So we can adjust that in the config via manjaro-tools if we want autologin unchecked by default.

oberon2007:
Though, if I'm not mistaken, currently [write_users_conf() in **util-yaml.sh**](https://github.com/manjaro/manjaro-tools/blob/master/lib/util-yaml.sh#L145) acutually uses the autologin setting for the live-environment, which is not necessarily what we want, @artoo?

shainer:
@oberon2007 for the TAB issue in the dialog, sorry, I cannot reproduce here. If you give more details about the system I can try again in a virtualbox.

For the shells, yes you are right; for some reason the config value type is no longer recognized as a QStringList but simply as a QString, which messes with the parsing. I'll look into that tomorrow.

oberon2007:
The Tab issue I experience in both a new manjaro-i3 as an xfce build. What would be more helpful system information for you? I am in a virtualbox, too.
I put the cursor in the username input field and after filling it in I use Tab to move to the first password box, then another Tab for the confirm password box. Then when I want to move on the the 'full name' field I have to press Tab three times...

shainer:
@oberon2007 the DE is fine. I thought you required 3 tabs to go from the first password field to the confirmation one, my bad. I will check to go from confirm password box to full name.

shainer:
@oberon2007 ok the tab order should now be sensible (at least for the flow I expect).

I also fixed the login shell parsing from the configuration.

oberon2007:
Working great, yes! :+1:
Any idea about the avatar thing? Where is calamares currently looking for the faces?

oberon2007:
I understand meanwhile that one needs to define ``avatarFilePath`` in order to make this functional. But to be honest I don't really understand how this is going to work. Could you explain a little more about your ideas on how to best make use of this?
Even when you deploy image files on the install medium, how will the new user know where to find and select them? We have mugshot installed by default for example, and those files are 'hidden' in /usr/share/pixmaps/faces/ ... :wink:
How about the path exactly? Do you define just the destination directory, or the target file name? Will it be possible to install for example to ``$HOME/.face``? Would be great if you could shed some more light on this.

philmmanjaro:
I'd say simply add the line `avatarFilePath: /usr/share/pixmaps/faces/` to **users.conf** and try it out ...

oberon2007:
No it's the other way round! You need to define the destination and the user needs to select the source in the dialog...

oberon2007:
I had already tried ``avatarFilePath: $HOME/.face`` but it didn't arrive in the target install. Maybe it ended up in /root I don't know. I don't have this install anymore ... So I decided to ask, since it takes a while to install everytime :stuck_out_tongue_winking_eye:

philmmanjaro:
Then it might be an issue with the code ... Did you checked the source on the procedure used to save the Avatar? Also did you checked with our KDE edition?

shainer:
@oberon2007 I expanded the description the README: the path needs to include the file name and I only expand ~ for now to the users' home, not environment variables.

I will do another test one of these days just to re-verify this works as expected.

oberon2007:
The image was copied to ~/ Thing is, it is still owned by root... Maybe that should be adjusted. As explained above though, I am just not completely convinced about the usefulness of this feature in general ;)

oberon2007:
@philmmanjaro do we already know about the issue that the "use the password as administrative pwd" is disabled? Did I miss something there? I end up with an installation without a way to use sudo or login as root!

philmmanjaro:
I didn't test the new version of **users** module yet. Normally if it doesn't work as expected it is a regression. Did you do a full debug on the module?

philmmanjaro:
As I see it. By default it is disabled to create a **root** account at all. The main user should have sudo rights, though:

```
setRootPassword: allow to set the root password in the installed system. Defaults to false.
```

oberon2007:
Ah I see. We even have the line in our config. I think we should enable it, because the way it is now you will not have a root filemanager available, nor can you use ``su``. Both I think is to be expected on an average install, don't you think @philmmanjaro ?
@shainer I have another question: Would it be possible to add a button 'Edit' on the user overview screen. Right now we have 'Create' and 'Delete' - I think this would make the functionality complete.

shainer:
Teo asked me to resubmit this pull request so that it's cleaner, as this one includes a lot of unrelated things for when I had to merge.

I will work on that this week, plus I'll look into fixing the avatar permissions, looking at having more possible destinations, and a few other corrections. For the edit, I may leave it for a future improvement. I will mention all the people who commented here on the new one so you know when it's ready. Thanks again for your patience and suggestions :)

philmmanjaro:
@shainer: as @oberon2007 already mentioned, there is some issue with the **root** password setup. I'm currently playing with **v2.4.5** and the old module to check what the matter is. If the root password is not set in any way we will have a security risk. In **v2.4.5** we have followed options:

* use the user password for root with: `setRootPassword: false` & `doReusePassword: true`
* use a different password for root with: `setRootPassword: true` & `doReusePassword: false`

I've to check if `setRootPassword: false` & `doReusePassword: false` won't set any password for root in **v2.4.5**. However there seems to be a problem within your code when `setRootPassword: false` is set.

shainer:
@philmmanjaro I am not sure what should happen when setRootPassword is false, let me know what you find in the current version.

However we could at least change the default to true, it sounds more sensible :)

shainer:
I recreated the pull request from a more recent branch, see #280.

I added the following changes:

- no modal dialog for the first user, but a warning on the bottom.
- change the permissions of the avatar file after copying.
- avoid failing the whole installation if the avatar copy fails.

For the root password, I am still not sure what's the best direction forward (beside my suggestion), please advise or ask the other developers about it :)

shainer:
Ignore my comment about the root password, that was fixed and it's included in my new branch.

oberon2007:
I was now able to inlcude all the modifications in the manjaro branch and we can now test again.
A little thing: I noticed that the order of input boxes for the root password seems to be swapped. So when I enter a machine name and hit Tab, the cursor jumps to the right box (the one for the password repeat) and from there on to the left box.
Apart from that on the KDE ISO I just tried I have problem with the file dialog for avatar selection. Whenever I try to change directory I get an error ``Unable to create io-slave. Cannot talk to klauncher: Not connected to D-Bus server`` - I have no idea, yet if this issue is calamares related at all, or maybe has to do with the ISO itself, though...

adriaandegroot:
What is the relationship of this PR with #280 ? It seems like that one makes this PR obsolete -- if so, @shainer , could you close this one?

philmmanjaro:
@adriaandegroot: yes. This one is obsolete and can be closed.

PR #271, opened 2016-10-31
Make sure we also set a preferred Xsession

shadeslayer:


philmmanjaro:
Seems to be OK from my point of view.

PR #272, opened 2016-11-03
Update Polish localization

m4sk1n:
Hi! I'm just asking is it possible to make Polish translation of Calamares actual? I have completed it on Transifex. That's all.


philmmanjaro:
@m4sk1n: don't use pull requests for downstream issues you may see. This is the wrong way. Calamares has a [bugtracker](https://bugs.calamares.io/). Also the translations done in transiflex get automatically merged to our **master** branch. Your translation will be added with our next **major** release.

PR #275, opened 2016-11-13
Add man page

highvoltage:
Add a man page for calamares.

### DIFF
--- a/man/calamares.1
+++ b/man/calamares.1
@@ -1,0 +1,33 @@
+.TH CALAMARES "1"
+.SH NAME
+calamares \- distribution-independent system installer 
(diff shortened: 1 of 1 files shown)


shainer:
Nice. Why don't you also install it in the correct path? :)

teo:
@highvoltage what @shainer said, this should be installed in the right path by adding the relevant operation to CMake.

teo:
@highvoltage any news on this?

adriaandegroot:
No worries, the CMake stuff is trivial. I am going to move the man page to section 8, though, since it's not a regular user executable command.

PR #276, opened 2016-11-20
Disallow assigning the same mountpoint to two partitions

shainer:
When creating/editing a partition, get a list of mountpoints that were already "assigned" to other partitions (minus the current one, if it exists). In case the user tries to assign one of these mountpoints in the dialog, display an italic error message just below the combo box, and disable the Ok button of the dialog.

This fixes https://calamares.io/bugs/projects/CAL/issues/CAL-58

Tested in a Virtualbox with Chakra.

### DIFF
--- a/src/modules/partition/gui/CreatePartitionDialog.cpp
+++ b/src/modules/partition/gui/CreatePartitionDialog.cpp
@@ -40,8 +40,9 @@
 // Qt
 #include <QComboBox>
 #include <QDir>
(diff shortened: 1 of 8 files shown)


shainer:
Just noticed we had a pull request for this problem already, back in 2014. A more descriptive title would have helped, I clicked on it by accident :(

As that would need careful merging anyway I will not close this one.

PR #279, opened 2017-01-17
added extrapackages module

paddatrapper:
The aptextrapackages module allows the user to install some packages
from a predefined list. These packages are described in the config file
and have the ability to run pre- and post-install scripts for each
package.

This module assumes apt is installed and will fail the install if it is
not. Thus it is only compatible with Debian and its derivatives.


shadeslayer:
Hi
Could provide a use case for this when we already have the packages module that allows installation of packages on a system?

ChrisFernandez:
oh this is awesome I was hopping for this... was going to write my own.. im SO glad someone did this.
now... is there a easy way to make it work with pacman for arch based installs?

paddatrapper:
@ChrisFernandez, you'll need to change the commands that are run by InstallPackageJob and SetSourcesJob.

src/modules/aptextrapackages/ExtraPackagesViewStep.cpp
```diff
@@ -0,0 +1,137 @@
...
+                 "module configuration.";
+    }
+    if ( configurationMap.contains( "sources" ) &&
+            configurationMap.value( "sources" ).toList().size() )
+        m_sources = configurationMap.value( "sources" ).toList();
+    else
+    {
+        cDebug() << "WARNING: No valid sources list found in aptextrapackages "
+                 "module configuration.";
+    }
```
teo: Please either use {} for one-liners, or don't, but keep it consistent.

src/modules/aptextrapackages/InstallPackageJob.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+InstallPackageJob::InstallPackageJob( const QString& name, const QString& preScript, const QString& postScript ) :
+    Calamares::Job(),
+    m_name( name ),
+    m_preScript( preScript ),
+    m_postScript( postScript ),
+    m_jobProgress( InstallJobProgress::PRE_SCRIPT )
+{ }
+
+QString InstallPackageJob::prettyName() const
```
teo: Return type on its own line please.

src/modules/aptextrapackages/SetSourcesJob.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include <QFile>
+#include <QTextStream>
+
+SetSourcesJob::SetSourcesJob( const QString& source ) :
+    Calamares::Job(),
+    m_source( source )
+{ }
+
+QString SetSourcesJob::prettyName() const
```
teo: Return type on its own line :)

src/modules/aptextrapackages/SetSourcesJob.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+    return tr( "Add <strong>%1</strong> to apt sources list" ).arg( m_source );
+}
+
+QString SetSourcesJob::prettyStatusMessage() const
+{
+    return tr( "Adding %1 to sources list" ).arg( m_source );
+}
+
+Calamares::JobResult SetSourcesJob::exec()
```
teo: Just curious, why does Calamares need to add stuff to sources.list? Can't the source list ship already configured on the ISO?

ChrisFernandez:
@paddatrapper cool I will check it out next weekend, but will be nice if you had a config file with the option to do it between for arch or debian base.

paddatrapper:
Feel free to add one. I don't work with Arch well enough to get that
running. Everything I use is Debian... This also may well change a lot
in the next few days as I will be trying to consolidate it with the
packages and netinstall modules (which may make it package manager
independent)

src/modules/aptextrapackages/ExtraPackagesViewStep.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+QWidget* ExtraPackagesViewStep::widget()
+{
+    return m_widget;
+}
+
+void ExtraPackagesViewStep::next()
+{
+    m_widget->checkInternet();
```
paddatrapper: `onActivate` is called when Calamares is started. We think users will see the note about needing internet, then go and plug in/connect to wifi. This means they simply need to go back and forward to enable the module after connecting instead of having to restart the installer

paddatrapper:
@teo We don't yet have an ISO using this module, but it should be fine to run just by adding it to `settings.conf`.

kkofler:
Using the packages module is a good thing, it makes it more portable, but the SetSourcesJob is still very distro-specific.

I also still see this module very much duplicating what the netinstall module does. There is also overlap in purpose with the license module, though that does not take care of the actual installation part and leaves it to custom jobs. I think it would be nice if we would not have conflicting modules for doing the same thing (but if the result then is that people still maintain the conflicting modules and just do it downstream, see e.g. `bootldr`, that is not going to help either).

paddatrapper:
@kkofler `SetSourcesJob` is now optional - leave `sources:` out in the config file and run that job. netinstall's UI is very much designed around groups of packages. While setting one package groups allows you to install individual packages it still looks clumsy. I am open to ideas around how to improve that UI rather in a way that still allows the current groups.

The original intention was to maintain this downstream, as it was very specific to our use-case in the assumptions it made, naming, etc, but Calamares out of source builds, etc are not working currently, hence the removal of the specifics and a move to a more general module

paddatrapper:
I am going to close this PR until I finish with redoing the netinstall UI as that will supersede this.

PR #280, opened 2017-01-17
Enhanced user dialog: equivalent pull request

shainer:
@teo this is the twin request of #270 .

It was created from a much more recent branch and contains a few extra commits (let me know if you want those replicated in the other branch). Tested with Chakra Goedel RC5.


oberon2007:
Just tested avatar installation with Cinnamon desktop and that worked fine! 👍

philmmanjaro:
@shainer: can you add an edit option to that module. If you want to change any detail for a created user you have to delete it and re-enter all details again via creation. Would make thinks easier, if for example the user forgot to check **autologin**.

src/modules/users/gui/UsersPage.h
```diff
@@ -0,0 +1,143 @@
...
+    bool autologin;
+};
+
+/*
+ * Model to contain the users created during the installation
+ * process. This is associated to a QTableView in the main UI.
+ *
+ * Each row is a user and each column a property of the user.
+ */
+class UsersListModel : public QAbstractTableModel {
```
teo: Maybe it would make sense to move this class to its own {h,cpp}?

src/modules/users/gui/UsersPage.h
```diff
@@ -0,0 +1,143 @@
...
+    Q_OBJECT
+
+public:
+    virtual ~UsersListModel();
+
+    int rowCount(const QModelIndex &parent = QModelIndex()) const Q_DECL_OVERRIDE;
```
teo: No need to use `Q_DECL_OVERRIDE`, plain `override` is just fine since we depend on C++11 and Qt 5.

src/modules/users/gui/UsersPage.h
```diff
@@ -0,0 +1,143 @@
...
+
+    QList< Calamares::job_ptr > createJobs( const QStringList& defaultGroupsList );
+
+    void onActivate();
+
+    // Setters for properties of this page, usually from configuration.
+    void setAutologin( bool autologin );
+    void setAvatarFilePath( const QString& path );
+    void setAvailableShells( const QStringList& shells);
+    void setHaveRootPassword(bool);
```
teo: Space inside brackets please :)

src/modules/users/jobs/CreateUserJob.cpp
```diff
@@ -16,7 +16,7 @@
  *   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
  */
```
teo: Please add your copyright line to the files you touch.

oberon2007:
For some reason now when I land on the users page the create user mask does not open automatically as it used to. I have to click on "Create" first to open it. Is this an intentional change? Or maybe related to the fact that I am using a hybrid build right now?

shainer:
@teo and @oberon2007 : yes I removed the modal dialog after it was discussed with teo on IRC. Now you should see a message on the bottom when you first enter the page, and you have to click on Create to open it. That was intentional. I will address teo's comments too later.

teo:
Some more observations on the UI, I've been testing on latest Manjaro, that's calamares-dev 3.0.0.r3012 in their packaging, so I don't know if these issues have already been addressed.
* The check mark icons seem to have different aspect ratios in different places. The QLabels should not stretch the pixmap inside.
* Tab order still seems slightly off, e.g. the confirm root password text box is focused before the first root password text box.
* From what I can tell, there is a QFrame in the modal dialog, which yields the typical KDE4-era "boxes in boxes in boxes" look and feel, which would be good to avoid.
* There seem to be some alignment issues between the "User name" label and entry field, and everything else (e.g. the spacing between the label and the text box is very small, the text box is wider than the full name text box, the vertical alignment is off compared to the password/confirm password lines, etc.). I advise switching to QGridLayout or QFormLayout, which exists precisely for this type of UI.
* I don't think the full name qualifies as "Advanced settings". One doesn't need to be a very advanced user in order to know their own name :D The login shell is probably an advanced setting, but for that too there is a reasonable default, so I don't think the "Advanced settings" label is necessary.
* Avatar selection could be done with a more meaningful UI, for instance like the user management UI in KDE System Settings. Try `kcmshell5 user_manager`.
* Generally speaking, regarding `kcmshell5 user_manager`, I think this UI is the general direction we should move towards, and I believe it's very doable from the current state of things.

philmmanjaro:
I totally agree with @teo here. The new **users** module is still odd to me too. Coming from the old one, it is more logical to keep that way in letting the user entering his full name and based on that the **unix name** gets generated. Also the focus and tap-order needs to be fixed. The **calamares-dev** package has all **users** module changes up to [January 18 2017](https://github.com/shainer/calamares/commit/9cd7dda29a959184dc240f4e88c495ccaa11c0c8).

In regard of Avatars. Users are used to select them from a dialog by clicking on them:

users
All in all it would be more clean and structured.

shainer:
Hi people,

these are good suggestions. I have a lot to learn about UI design :) however some of them require a non-trivial amount of work for me; as you have noticed I am a bit short on time. I will try to address the simple changes in the upcoming weeks, otherwise, would some of you be able to help with the larger stuff?

teo:
@shainer I'd be happy to keep reviewing and helping out with UI stuff. I think @philmmanjaro's screenshot is an excellent design, feel free to copy it. The trickier part is sometimes beating QWidgets into submission, taking care of the little things like layouting, alignment, fonts, focus & tab order, etc. but I'm confident you can master it. I can help you with this.

shainer:
Thanks for offering @teo :) I am simply not sure I have time to dedicate to this at the moment.

Phil's design looks good but it works in the context of a settings manager. It looks very different from the rest of Calamares so I don't how it integrates there.
I think that if you are considering changes to the "basic" UI you should work on that separately or at least have a prototype for now.

adriaandegroot:
This PR reached the "super awesome" stage a while back, but has stagnated over the last bits of polish, and now there's conflicts. @shainer , do you have time to rebase and resolve? Then we can go over it again and merge.

philmmanjaro:
@adriaandegroot: This PR was done a while ago, yes. However the UX design was not perfect. Therefore the last idea was to do some similar than already done as in **User Manager** from **System Settings**. The current code itself is fine and is still used in my development branch of **CAL** for Manjaro, which has already merged this one.

feren:
Bump

PR #281, opened 2017-01-23
Moved netinstall to a tree based UI

paddatrapper:
Netinstall now uses a QTreeView to display a Model of the packages. Old config files still work, but now there is support for subgroups:

````yaml
- name: foo
  description: foo description
  packages:
      - package_name_1
      - package_name_2
  subgroups:
      - name: bar
        description: bar description
        packages:
            - package_name_1
            - package_name_2
````

There is also support for install hooks. The packages module takes entries with 3 keys: `pre-script`, `package` and `post-script`. It will run the `pre-script` entry from within the installed system, then install the package and finally run the `post-script` entry, again from within the installed system:

````yaml
rootMountPoint: /tmp/mount
packageOperations:
  - install:
    - pre-script: touch /foo
      package: vi
      post-script: touch /bar
````

I have built and tested this using Calamares 2.4.5, as there are a couple of Debian specific patches that need to be applied to be able to build Calamares and version 3 is not yet packaged. I have tried to forward port my changes, but I suggest that this is tested against the current master for safety.

### CODE REVIEW

**Comment by oberon2007** [comment]
I've been hoping for that to become acailable, awesome! :+1:

**Comment by kkofler** [comment]
> `subpackages:`

You mean `subgroups:` (it's correct in the code, it's just your example that is wrong).

**Comment by paddatrapper** [comment]
You are right, corrected

File: src/modules/netinstall/PackageModel.h [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+class PackageModel : public QAbstractItemModel
+{
+    Q_OBJECT
+
+public:
+    explicit PackageModel( const YAML::Node& data, const QVariantList& columnHeadings,
+                           QObject* parent = 0 );
+    ~PackageModel();
+
+    QVariant data( const QModelIndex& index, int role ) const Q_DECL_OVERRIDE;
```
teo: It's ok to use plain `override` instead of `Q_DECL_OVERRIDE` because Calamares depends on C++11 and Qt 5.

src/modules/netinstall/PackageTreeItem.cpp
```diff
@@ -0,0 +1,203 @@
...
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "PackageTreeItem.h"
+
+#include "utils/Logger.h" // TODO:Remove
```
teo: TODO: Remove?

src/modules/netinstall/PackageTreeItem.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+PackageTreeItem::PackageTreeItem( const ItemData& data, PackageTreeItem* parent ) :
+    m_data( data ),
+    m_parentItem( parent )
+{ }
+
+PackageTreeItem::PackageTreeItem( const QString packageName, PackageTreeItem* parent ) :
+    m_packageName( packageName ),
+    m_parentItem( parent ),
+    m_selected( parent->isSelected() )
```
teo: Might parent be nullptr?

src/modules/netinstall/PackageModel.cpp
```diff
@@ -0,0 +1,260 @@
...
+        return QVariant();
+    PackageTreeItem* item = static_cast<PackageTreeItem*>( index.internalPointer() );
+    if ( index.column() == 0 && role == Qt::CheckStateRole )
+        return item->isSelected();
+
+    if ( !item->childCount() ) // package
+    {
+        if ( !item->parentItem()->isHidden() && role == Qt::DisplayRole &&
+                index.column() == 0 )
+            return QVariant( item->packageName() );
```
paddatrapper: It is currently - the check is includes `role == Qt::DisplayRole`. The problem is handling the difference between groups, with `prettyName()` and packages with `packageName()`. Thus this section is dealing with the packages.

However the `else if ( role == PackageTreeItem::PackageNameRole )` is never actually used I don't think (I don't see any calls using the role to get the package name), so I do still need to clean this up a bit/use it

philmmanjaro:
We from **Manjaro** adopted this in our [development branch](https://github.com/manjaro/calamares/commit/55b56b4b5df2848f8a43e0f75cf989874087bd94). The initial version of this modification is included in [**calamares-dev-3.0.0.r2999.77b83974-1**](https://lists.manjaro.org/pipermail/manjaro-packages/Week-of-Mon-20170123/009831.html). I assume @oberon2007 will create a test-image ISO soon.

src/modules/netinstall/PackageModel.cpp
```diff
@@ -0,0 +1,260 @@
...
+            return QVariant();
+    }
+
+    if ( item->isHidden() && role == Qt::DisplayRole ) // Hidden group
+        return QVariant();
+
+    switch ( role )
+    {
+    case Qt::DisplayRole:
+        return item->data( index.column() );
```
paddatrapper: Because the item holds the data not the model? But I do see your point with the rest of the function's behaviour

paddatrapper:
We should have an ISO with this in soon. @highvoltage is busy building it

src/modules/netinstall/PackageModel.cpp
```diff
@@ -0,0 +1,266 @@
...
+    for ( auto package : m_hiddenItems )
+        items.append( getItemPackages( package, isCritical ) );
+    QList<QVariant> packages;
+    for ( auto item : items )
+    {
+        QMap<QString, QVariant> itemMap;
+        itemMap.insert( "pre-script", item->parentItem()->preScript() ); // Only groups have hooks
+        itemMap.insert( "package", item->packageName() );
+        itemMap.insert( "post-script", item->parentItem()->postScript() );
+        packages.append( QVariant( itemMap ) );
```
shainer: This conversion to QVariant hides the type you are returning (map from "type" of object to the object name), and it's useless because you always return the same type.

I would prefer returning a list of QMaps with no conversion; if you think the type name is too long you can use typedef, or define a little struct to use in place of the QMaps.

src/modules/netinstall/PackageModel.cpp
```diff
@@ -0,0 +1,266 @@
...
+    }
+    return packages;
+}
+
+QList<PackageTreeItem*>
+PackageModel::getItemPackages( PackageTreeItem* item, bool isCritical ) const
```
shainer: You pass a single boolean around a lot; consider always returning all packages and then differentiating them when you build the install and try-install lists.

src/modules/netinstall/PackageModel.cpp
```diff
@@ -0,0 +1,266 @@
...
+        if ( itemDefinition["critical"] )
+            item->setCritical(
+                CalamaresUtils::yamlToVariant( itemDefinition["critical"] ).toBool() );
+
+        for ( YAML::const_iterator packageIt = itemDefinition["packages"].begin();
+                packageIt != itemDefinition["packages"].end(); ++packageIt )
+            item->appendChild( 
+                    new PackageTreeItem( CalamaresUtils::yamlToVariant( *packageIt ).toString(), item ) );
+
+        if ( itemDefinition["subgroups"] )
```
shainer: Consider enforcing a limit on the level of recursion here, to avoid ugly views.

shainer:
Nice work! Two more comments from me:

- please update the README too to explain the new functionality.
- In the UI, I think a slightly larger text would look better. This looks a bit like an editor :)

src/modules/netinstall/PackageModel.cpp
kkofler: Ugliness is in the eye of the beholder, and that's something the maintainer of the distribution's netinstall group list YAML file will have to sort out. This is not something the end user can mess with. So I don't see a point in hardcoding a limit here.

paddatrapper:
@teo, @shainer corrections made. Should be fine now

paddatrapper:
Looking at the diff I realise I haven't merged the packages module changes properly - will need to do so tomorrow morning

paddatrapper:
All fixed. Packages module now is up to date and everything should be working as expected

oberon2007:
Hi @paddatrapper . On my first testbuild of a hybrid ISO, attempting to use subgroups I get
```
terminate called after throwing an instance of 'YAML::ParserException'
  what():  yaml-cpp: error at line 6, column 8: end of sequence not found
Segmentation fault
```
Is there something wrong with my syntax in the .yaml file?
I'm attempting to use [this one](https://raw.githubusercontent.com/oberon2007/iso-profiles/master/community/cinnamon-hybrid/packages-systemd.yaml)
Mind having a look?

paddatrapper:
@oberon2007 I did not expect groups not to have packages, so it's failing [here](https://github.com/calamares/calamares/pull/281/files#diff-f65d1da97fe644c1ac69d0ee95144b5fR234). You raise an interesting point, that should be fairly trivial to implement (I hope), which is to make packages of a group optional. I'll be able to take a stab at it on Monday 5 Feb, as I am away for next week. It should just require encasing the loop in an `if` clause. I.E:

````c++
if ( itemDefinition["packages"] )
    for ( YAML::const_iterator packageIt = itemDefinition["packages"].begin();
            packageIt != itemDefinition["packages"].end(); ++packageIt )
        item->appendChild(
            new PackageTreeItem( CalamaresUtils::yamlToVariant( *packageIt ).toString(), item ) );
````

**Comment by oberon2007** [comment]
I understand. Thank you! :+1:

**Comment by paddatrapper** [comment]
@oberon2007 packages are now optional

**Comment by oberon2007** [comment]
Cool! Thank you! Will test shortly! :)

**Comment by oberon2007** [comment]
@paddatrapper |'m afraid I still get the same error:
```
QCoreApplication::postEvent: Unexpected null receiver
terminate called after throwing an instance of 'YAML::ParserException'
  what():  yaml-cpp: error at line 5, column 8: end of sequence not found
```

**Comment by philmmanjaro** [comment]
@oberon2007: can you provide a link to the **ISO**, so we can check why that might happen? Also only the **calamares-dev** package uses this new module.

**Comment by oberon2007** [comment]
Manjaro **calamares-dev** package inlcluding https://github.com/manjaro/calamares/commit/dd25323a08b02dae861d5eafed50dd02498ef758 was used. I'll upload the ISO and share a link...

**Comment by oberon2007** [comment]
My test ISO is [here](https://sourceforge.net/projects/manjaro-cinnamon/files/17.0-beta1/cinnamon-hybrid/).

**Comment by paddatrapper** [comment]
@oberon2007 Your yaml is malformed by one space, `description` should start in the same column as `name`. [Here's](https://gist.github.com/paddatrapper/5c73ba3f74c64c3dbea5d6650174ed3f) a corrected version.

**Comment by philmmanjaro** [comment]
@oberon2007: seems the one you linked already [here](https://raw.githubusercontent.com/oberon2007/iso-profiles/master/community/cinnamon-hybrid/packages-systemd.yaml) was also malformated:

**wrong example**:
```
- name: "Internet"
  description: "Browser, Chat and Filetransfer"
  subgroups:
      - name: "Firefox"
       description: "Mozilla Firefox"
       packages:
            - firefox
      - name: "Chromium"
       description: "Chromium"
       packages:
            - chromium
      - name: "Palemoon"
       description: "Palemoon"
       packages:
            - palemoon-bin
```

**corrected example**
```
- name: "Internet"
  description: "Browser, Chat and Filetransfer"
  subgroups:
      - name: "Firefox"
        description: "Mozilla Firefox"
        packages:
            - firefox
      - name: "Chromium"
        description: "Chromium"
        packages:
            - chromium
      - name: "Palemoon"
        description: "Palemoon"
        packages:
            - palemoon-bin
```

**Comment by philmmanjaro** [comment]
To have it a standard, I also recommend to use always four spaces:

**yours:**
```
- name: "Utilities"
  description: "GUI tools"
  packages:
        - baobab
        - galculator
        - gparted
```

**suggestion:**
```
- name: "Utilities"
  description: "GUI tools"
  packages:
      - baobab
      - galculator
      - gparted
```

oberon2007:
Thanks! My bad, I was sure I had copied the example exactly. Apparently the formatting slipped at one point.  :wink: 
Will correct that and check again.

oberon2007:
Ok. here's the stange thing. When I have turned on highlighting "yaml" in **xed** to achieve this:
formatting
I have to use in fact the wrong amount of spaces :stuck_out_tongue_winking_eye:
When I turn the highlighting off things are different and when the formatting is correct it looks off when the highlighting is switched on :laughing: 
Sorry for that. I'll have to report this bug with **linuxmint**

oberon2007:
Great! Working now:

netinst

I especially like very much that you can either select whole subgroups or individual packages from their content. :+1: 
Still an unsolved problem from my POV is the missing feedback during package download. The progress just "freezes" at 26% with the message "Running packages operation" as long as it takes to download and install all requested packages - which can be a very long time.
Most people will just assume that the installation has stopped working - early testing in our community confirms that.
But of course this is not a problem of this PR but a general netinstall issue...
Thank you very much @paddatrapper !

paddatrapper:
No problem. That is actually an issue with the packages module and not netinstall. There is a similar issue for me where the packages to install are not listed in the summary page, but fixing that needs careful consideration around how it is displayed to avoid a wall of text or a very long list

teo:
@paddatrapper this should be fixable in `packages/main.py` by calculating how many package operations have been completed at any given time out of a total, and calling `libcalamares.job.setprogress` (but this is unrelated to the present PR).

teo:
@paddatrapper this is nothing short of awesome! Nice work! I can think of potential improvements in the UI, stuff like icons, tooltips, translatable groups and package descriptions, etc. but all of that is an extra that can be done by anyone in the future, so I'm happy to merge this.

paddatrapper:
Thanks! Create issues and I'll see what I can do

PR #282, opened 2017-01-28
returns if there is a real connection and not only the enabled device

AlmAck:
refactoring of the checkHasInternet() function.
based on https://doc.qt.io/qt-5/qnetworkconfigurationmanager.html#isOnline the isOnline returns the real status if the connection is available, no ping requests are required.

This patch should fix CAL-404 and CAL-432

### DIFF
--- a/src/modules/welcome/checker/RequirementsChecker.cpp
+++ b/src/modules/welcome/checker/RequirementsChecker.cpp
@@ -36,7 +36,9 @@
 #include <QFile>
 #include <QFileInfo>
 #include <QLabel>
-#include <QNetworkAccessManager>
+#include <QNetworkConfigurationManager>
+#include <QNetworkConfiguration>
+
 #include <QProcess>
 #include <QTimer>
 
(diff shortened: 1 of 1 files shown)


demmm:
Was this tested on a system with a loopback device?

philmmanjaro:
@demmm: I only tested it with VirtualBox for now. What kind of speciality do you have with a loopback device?

demmm:
See all issue with using C++ for testing network connections,all just check if a device can connect, period, so there is no distinction whether a connection is made to the outside or internal.
This is an example for loopback device:
```
hwinfo --network
54: None 00.0: 10700 Loopback                                   
  [Created at net.125]
  Unique ID: ZsBS.GQNx7L4uPNA
  SysFS ID: /class/net/lo
  Hardware Class: network interface
  Model: "Loopback network interface"
  Device File: lo
  Link detected: yes
  Config Status: cfg=new, avail=yes, need=no, active=unknown
```

src/modules/welcome/checker/RequirementsChecker.cpp
```diff
@@ -337,8 +339,20 @@ RequirementsChecker::checkHasPower()
...
-    bool hasInternet = QNetworkAccessManager(this).networkAccessible() != QNetworkAccessManager::NotAccessible;
+    // doc from: https://doc.qt.io/qt-5/qnetworkconfigurationmanager.html#isOnline
+    bool hasInternet = false;
+
+    QNetworkConfigurationManager mgr;
+    QList<QNetworkConfiguration> activeConfigs = mgr.allConfigurations(QNetworkConfiguration::Active);
+
+    if (activeConfigs.count() > 0) {
+        Q_ASSERT(mgr.isOnline());
+        hasInternet = mgr.isOnline();
```
teo: This is not what we need. QNetworkConfigurationManager merely checks whether a network interface is online and connected to another host, it yields no information on actual internet connectivity.

AlmAck: I agree, after some discussion in IRC is clear that this code is not robust enough. I see only as a valid solution a "ping" request to a defined (configurable) server. What do you think?

teo: @AlmAck I've just pushed a change to master which should improve the situation. Please test and let me know if it fixes your use case.

efremmc: Is there a way to so a socket connection, open a socket on a well known port and return the results

PR #283, opened 2017-02-05
partition: use always double click to activate a partition tree view item

philmmanjaro:
Using activated signal may use single or double click depending on environment settings.

At Manjaro we tested with **sudo** only. There it is always double-click also without our patch. However by using **sudo -E** or **pkgexec** with [additional theme patch](https://github.com/manjaro/packages-extra/blob/master/polkit/x11vars.patch) it always resulted in the way how the environment settings were configured for mouse behaviour.

This patch fixes it, to have always double-click for selecting partitions.

### DIFF
--- a/src/modules/partition/gui/PartitionPage.cpp
+++ b/src/modules/partition/gui/PartitionPage.cpp
@@ -96,7 +96,7 @@ PartitionPage::PartitionPage( PartitionCoreModule* core, QWidget* parent )
 
     connect( m_core, &PartitionCoreModule::isDirtyChanged, m_ui->revertButton, &QWidget::setEnabled );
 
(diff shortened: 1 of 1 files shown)


philmmanjaro:
Tested within [17.0-beta2-r2](https://sourceforge.net/projects/manjarotest/files/17.0-beta2/kde/).

kkofler:
What I am missing here is a reason why the system setting is not correct to follow here. Do we expect a click to do something different from a double click in this context?

Kirek:
Because if for example you want to delete a partition, selecting it will open a dialog, the you have to close it and delete the partition.

kkofler:
So that makes sense, no objection from me then.

PR #284, opened 2017-02-20
Updated docs to better explain new features

paddatrapper:
The docs do not current reflect the new net-install and packages modules changes well. This updates them to explain the new keys in the configuration files.


teo:
Thanks for the docs update @paddatrapper

PR #696, opened 2017-03-16
WIP: fixes #692  bootloader module fails when /EFI/Boot exists

### LINKED ISSUE #692: bootloader module fails when /EFI/Boot exists
```
Boost.Python error in job "bootloader".
<class 'subprocess.CalledProcessError'>
Command 'mkdir -p /boot/efi/EFI/boot' returned non-zero exit status 1
Traceback:
  File "/usr/lib/calamares/modules/bootloader/main.py", line 309, in run
    prepare_bootloader(fw_type)

  File "/usr/lib/calamares/modules/bootloader/main.py", line 283, in prepare_bootloader
    install_grub(efi_directory, fw_type)

  File "/usr/lib/calamares/modules/bootloader/main.py", line 231, in install_grub
    check_target_env_call(["mkdir", "-p", efi_boot_directory])

  File "<string>", line 2, in <module>
```
In the current case /boot/efi/EFI/Boot was created by MS Win 10 - the bug is reproducible, one of our Users reported exactly the same bug on  new machine with a pre installed Win10.

agaida:
to make it short - it doesn't help much if one tries to find and process any
EFI related things in the live system. The better approach is to search in
the chroot.

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -35,7 +35,8 @@ from libcalamares.utils import check_target_env_call
 
 
 def get_uuid():
-    """ Checks and passes 'uuid' to other routine.
+    """
+    Checks and passes 'uuid' to other routine.
 
     :return:
     """
@@ -53,8 +54,9 @@ def get_uuid():
 
 
 def get_bootloader_entry_name():
-    """ Passes 'bootloader_entry_name' to other routine based
-        on configuration file.
+    """
+    Passes 'bootloader_entry_name' to other routine based
+    on configuration file.
 
     :return:
     """
@@ -66,7 +68,8 @@ def get_bootloader_entry_name():
 
 
 def get_kernel_line(kernel_type):
-    """ Passes 'kernel_line' to other routine based on configuration file.
+    """
+    Passes 'kernel_line' to other routine based on configuration file.
 
     :param kernel_type:
     :return:
@@ -84,7 +87,8 @@ def get_kernel_line(kernel_type):
 
 
 def create_systemd_boot_conf(uuid, conf_path, kernel_line):
-    """ Creates systemd-boot configuration files based on given parameters.
+    """
+    Creates systemd-boot configuration files based on given parameters.
 
     :param uuid:
     :param conf_path:
@@ -105,11 +109,12 @@ def create_systemd_boot_conf(uuid, conf_path, kernel_line):
             swap_uuid = partition["uuid"]
 
         if partition["mountPoint"] == "/" and "luksMapperName" in partition:
-            cryptdevice_params = [
-                "cryptdevice=UUID={!s}:{!s}".format(partition["luksUuid"],
-                                                    partition["luksMapperName"]),
-                "root=/dev/mapper/{!s}".format(partition["luksMapperName"])
-            ]
+            cryptdevice_params = ["cryptdevice=UUID="
+                                  + partition["luksUuid"]
+                                  + ":"
+                                  + partition["luksMapperName"],
+                                  "root=/dev/mapper/"
+                                  + partition["luksMapperName"]]
 
     if cryptdevice_params:
         kernel_params.extend(cryptdevice_params)
@@ -136,7 +141,8 @@ def create_systemd_boot_conf(uuid, conf_path, kernel_line):
 
 
 def create_loader(loader_path):
-    """ Writes configuration for loader.
+    """
+    Writes configuration for loader.
 
     :param loader_path:
     """
@@ -155,7 +161,8 @@ def create_loader(loader_path):
 
 
 def install_systemd_boot(efi_directory):
-    """ Installs systemd-boot as bootloader for EFI setups.
+    """
+    Installs systemd-boot as bootloader for EFI setups.
 
     :param efi_directory:
     """
@@ -169,11 +176,11 @@ def install_systemd_boot(efi_directory):
     conf_path = os.path.join(install_efi_directory,
                              "loader",
                              "entries",
-                             "{!s}.conf".format(distribution_translated))
+                             distribution_translated +".conf")
     fallback_path = os.path.join(install_efi_directory,
                                  "loader",
                                  "entries",
-                                 "{!s}-fallback.conf".format(distribution_translated))
+                                 distribution_translated + "-fallback.conf")
     loader_path = os.path.join(install_efi_directory,
                                "loader",
                                "loader.conf")
@@ -190,7 +197,8 @@ def install_systemd_boot(efi_directory):
 
 
 def install_grub(efi_directory, fw_type):
-    """ Installs grub as bootloader, either in pc or efi mode.
+    """
+    Installs grub as bootloader, either in pc or efi mode.
 
     :param efi_directory:
     :param fw_type:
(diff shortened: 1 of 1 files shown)


src/modules/bootloader/main.py
```diff
@@ -193,9 +193,11 @@ def install_grub(efi_directory, fw_type):
     """
     if fw_type == "efi":
         print("Bootloader: grub (efi)")
+        install_path = libcalamares.globalstorage.value("rootMountPoint")
+        install_efi_directory = install_path + efi_directory
 
-        if not os.path.isdir(efi_directory):
-            check_target_env_call(["mkdir", "-p", "{!s}".format(efi_directory)])
+        if not os.path.isdir(install_efi_directory):
+            subprocess.call(["mkdir", "-p", "{!s}".format(install_efi_directory)])
```
shadeslayer: Could you explain this change? To me it seems like check_target_env_call(efi_directory) and this subprocess.call would do the exact same thing?

src/modules/bootloader/main.py
```diff
@@ -220,24 +222,25 @@ def install_grub(efi_directory, fw_type):
                                "--force"])
 
         # VFAT is weird, see issue CAL-385
-        efi_directory_firmware = os.path.join(efi_directory, "EFI")
-        if os.path.exists(efi_directory_firmware):
-            efi_directory_firmware = vfat_correct_case(efi_directory, "EFI")
+        install_efi_directory_firmware = vfat_correct_case(install_efi_directory, "EFI")
+        if not os.path.exists(install_efi_directory_firmware):
+            subprocess.call(["mkdir", "-p", install_efi_directory_firmware])
```
shadeslayer: Likewise, why not call check_target_env_call?

src/modules/bootloader/main.py
```diff
@@ -220,24 +222,25 @@ def install_grub(efi_directory, fw_type):
...
-        if os.path.exists(efi_boot_directory):
-            efi_boot_directory = vfat_correct_case(efi_directory_firmware, "boot")
-        else:
-            check_target_env_call(["mkdir", "-p", efi_boot_directory])
+        # there might be several values for the boot directory
+        # most usual they are boot, Boot, BOOT
+
+        install_efi_boot_directory = vfat_correct_case(install_efi_directory_firmware, "boot")
+        if not os.path.exists(install_efi_boot_directory):
+            subprocess.call(["mkdir", "-p", install_efi_boot_directory])
```
shadeslayer: Likewise.

src/modules/bootloader/main.py
```diff
@@ -220,24 +222,25 @@ def install_grub(efi_directory, fw_type):
...
-        efi_file_target = {"32": os.path.join(efi_boot_directory, "bootia32.efi"),
-                           "64": os.path.join(efi_boot_directory, "bootx64.efi")}
-        check_target_env_call(["cp",
-                               efi_file_source[efi_bitness],
-                               efi_file_target[efi_bitness]])
+        efi_file_source = {"32": os.path.join(install_efi_directory_firmware, efi_bootloader_id, "grubia32.efi"),
+                           "64": os.path.join(install_efi_directory_firmware, efi_bootloader_id, "grubx64.efi")}
+        efi_file_target = {"32": os.path.join(install_efi_boot_directory, "bootia32.efi"),
+                           "64": os.path.join(install_efi_boot_directory, "bootx64.efi")}
+        subprocess.call(["cp",
```
shadeslayer: Likewise.

agaida: Thought off, but i see code in calamares that doing it the same way - it's was easier and less changes.

agaida:
@shadeslayer - first i noticed that the local system is checked and if no EFI is found check_target_env_call is used to create the missed dir. So my point is: We have the complete set of well set variables available and have only to add the install path to a new one and use this new variable in the check - these checks are strictly local, even the vfat correction is local and give back a corrected path with the chroot part included. So i decided to fire up an subprocess and use this directly.

shadeslayer:
The part of the PR that makes the check use the target install dirs are correct afaik, I'm only requesting changes on the subprocess.call bits.

After grepping for subprocess.call in the code, there's only one instance of it in src/modules/bootloader/main.py which can probably made to use check_target_env_call too. 

My recommendation is to merge the parts of this PR that do the check on the target install while keeping the check_target_env_call parts that run the commands.

tux3l:
@shadeslayer I think @agaida wanted to keep the code clean an readable, it is true that you can use check_target_env_call but then you check with for example install_efi_directory (absolute path seen from live-system) and you are using efi_directory (absolute path seen from chroot) in check_target_env_call to create the directory. This can lead to some confused people as they do not know that os.path. is checking in your livesystem and not the chroot. 
And, if i see correct, the function vfat_correct_case can only work with absolute paths from your live-system, it will not work with chroot paths like efi_directory. So install_efi_directory contains now a valid path for creating the efi-directory. 
(vfat_correct_case uses os.listdir() which will look for the path in your live-system not the chroot-system)

shadeslayer:
If you really want to do this with the full path, why use subprocess.call at all?

Why not use the Python built in's to create folders.

agaida:
@shadeslayer - very good question, see the disclaimer, i'm a python noob - you are right. I will change it

agaida:
@shadeslayer - thanks for the hint, looks cleaner now and 'works for me'™

philmmanjaro:
@agaida: since you changed the code, pls add also your copyright to it. Thx.

src/modules/bootloader/main.py
```diff
@@ -220,24 +223,21 @@ def install_grub(efi_directory, fw_type):
...
 
-        efi_boot_directory = os.path.join(efi_directory_firmware, "boot")
-        if os.path.exists(efi_boot_directory):
-            efi_boot_directory = vfat_correct_case(efi_directory_firmware, "boot")
-        else:
-            check_target_env_call(["mkdir", "-p", efi_boot_directory])
+        # there might be several values for the boot directory
+        # most usual they are boot, Boot, BOOT
+
+        install_efi_boot_directory = vfat_correct_case(install_efi_directory_firmware, "boot")
```
shadeslayer: Needs splitting into two lines to keep pep8 happy

src/modules/bootloader/main.py
```diff
@@ -220,24 +223,21 @@ def install_grub(efi_directory, fw_type):
...
 
         # Workaround for some UEFI firmwares
-        efi_file_source = {"32": os.path.join(efi_directory_firmware, efi_bootloader_id, "grubia32.efi"),
-                           "64": os.path.join(efi_directory_firmware, efi_bootloader_id, "grubx64.efi")}
-        efi_file_target = {"32": os.path.join(efi_boot_directory, "bootia32.efi"),
-                           "64": os.path.join(efi_boot_directory, "bootx64.efi")}
-        check_target_env_call(["cp",
-                               efi_file_source[efi_bitness],
-                               efi_file_target[efi_bitness]])
+        efi_file_source = {"32": os.path.join(install_efi_directory_firmware, efi_bootloader_id, "grubia32.efi"),
```
shadeslayer: Needs splitting to keep pep8 happy

agaida: :) - will do so

shadeslayer:
LGTM

philmmanjaro:
@shadeslayer: so we merge it?

philmmanjaro:
Code looks fine so far. Any objections?

shadeslayer:
Merged

PR #701, opened 2017-03-20
Fixed pep8 whining in module machineid

agaida:
added myself to copyright


shadeslayer:
Perhaps a better idea would be to join strings together instead of using formatting ?

So :

A = B + "foo"

agaida:
to be honest:
```
    target_systemd_machineid_file = root_mount_point + "/etc/machine-id"
    target_dbus_machineid_file = root_mount_point + "/var/lib/dbus/machine-id"
```
looks better to me

shadeslayer:
That's what I'm proposing ;)

agaida:
but it simply works and is not geeky :P

shadeslayer:
LGTM

PR #703, opened 2017-03-21
Fix pep8 whining in module dummypythonqt

agaida:
added myself to copyright

### DIFF
--- a/src/modules/dummypythonqt/main.py
+++ b/src/modules/dummypythonqt/main.py
@@ -4,6 +4,7 @@
 # === This file is part of Calamares - <http://github.com/calamares> ===
 #
 #   Copyright 2016-2017, Teo Mrnjavac <teo@kde.org>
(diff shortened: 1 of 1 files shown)


src/modules/dummypythonqt/main.py
```diff
@@ -168,6 +173,8 @@ def retranslate(self, locale_name):
 # same rules apply as for ViewStep. No decorators are necessary here, because
 # only the ViewStep implementation is the unique entry point, and a module can
 # have any number of jobs.
+
+
```
shadeslayer: Two blank lines not one?

agaida: right - try the pep8 check and you will see that the check whining without two lines before a class definition :)

shadeslayer: Ah ok.

PR #717, opened 2017-04-09
Fixed pep8 whining in module mount

agaida:
Added myself to copyright

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -4,6 +4,7 @@
 # === This file is part of Calamares - <http://github.com/calamares> ===
 #
 #   Copyright 2014, Aurélien Gâteau <agateau@kde.org>
+#   Copyright 2017, Alf Gaida <agaida@siduction.org>
 #
 #   Calamares is free software: you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


src/modules/mount/main.py
```diff
@@ -80,37 +84,44 @@ def mount_partitions(root_mount_point, partitions):
             subprocess.check_call(["umount", "-v", root_mount_point])
 
             if "luksMapperName" in partition:
-                libcalamares.utils.mount("/dev/mapper/{!s}".format(partition["luksMapperName"]),
+                libcalamares.utils.mount("/dev/mapper/"
+                                         + partition["luksMapperName"],
```
adriaandegroot: I wonder (similar to other PR) about changing the explicit conversion to an implicit assumption. A different coding style which pycodestyle accepts (and I would too) is:

```
                    libcalamares.utils.mount(
                        "/dev/mapper/{!s}".format(partition["luksMapperName"]),
                        root_mount_point + "/home",
                        fstype,
                        ",".join(
                            ["subvol=@home", partition.get("options", "")]),
                        )
```

src/modules/mount/main.py
```diff
@@ -80,37 +84,44 @@ def mount_partitions(root_mount_point, partitions):
...
                                          mount_point,
                                          fstype,
                                          ",".join(["subvol=@",
-                                                   partition.get("options", "")]),
+                                                   partition.get("options",
+                                                                 "")]),
```
adriaandegroot: Here, too (and elsewhere in this PR), shifting all the arguments down one line and using one indent, as opposed to visual indent, works and IMO improves readability by keeping  the `get()` together.

agaida: you are right, looks far better

PR #718, opened 2017-04-13
Fixed pep8 whining for module fstab

agaida:


src/modules/fstab/main.py
```diff
@@ -181,8 +181,10 @@ def generate_fstab(self):
             print(FSTAB_HEADER, file=fstab_file)
 
             for partition in self.partitions:
-                # Special treatment for a btrfs root with @ and @home subvolumes
-                if partition["fs"] == "btrfs" and partition["mountPoint"] == "/":
+                # Special treatment for a btrfs root with @ and @home
+                # subvolumes
+                if (partition["fs"] == "btrfs"
+                   and partition["mountPoint"] == "/"):
```
philmmanjaro: Should this not be four after an if?

agaida:
With current intendation the check say; OK
```
pycodestyle --show-pep8 --show-source main.py && echo "OK"
OK
```
With 4 blanks the pep8 check would complain:

```
main.py:187:5: E129 visually indented line with same indent as next logical line
                    and partition["mountPoint"] == "/"):
    ^
    Continuation lines indentation.

    Continuation lines should align wrapped elements either vertically
    using Python's implicit line joining inside parentheses, brackets
    and braces, or using a hanging indent.

    When using a hanging indent these considerations should be applied:
    - there should be no arguments on the first line, and
    - further indentation should be used to clearly distinguish itself as a
      continuation line.
```

src/modules/fstab/main.py
```diff
@@ -252,8 +254,8 @@ def generate_fstab_line_info(self, partition):
             return dict(device="UUID=" + partition["uuid"],
                         mount_point=mount_point,
                         fs=filesystem,
-                        options=",".join(["subvol={}".format(partition["subvol"]),
-                                          options]),
+                        options=",".join(["subvol=" + partition["subvol"]],
```
adriaandegroot: Are you sure `partition["subvol"]` is a string here? The previous code would convert implicitly.

src/modules/fstab/main.py
```diff
@@ -266,11 +268,11 @@ def generate_fstab_line_info(self, partition):
 
     def print_fstab_line(self, dct, file=None):
         """ Prints line to '/etc/fstab' file. """
-        line = "{:41} {:<14} {:<7} {:<10} 0       {}".format(dct["device"],
-                                                             dct["mount_point"],
-                                                             dct["fs"],
-                                                             dct["options"],
-                                                             dct["check"],
+        line = "{:41} {:<14} {:<7} {:<10} 0      {}".format(dct["device"],
```
adriaandegroot: You're moving these by one space just to satisfy the style checker -- why not compact the fstab line all the way to `"{:41} {:<14} {:<7} {:<10} 0 {}"` and leave those extra spaces out?

src/modules/fstab/main.py
agaida: yes - lines 186 ff

src/modules/fstab/main.py
agaida: you are right

PR #736, opened 2017-06-02
Fixed pep8-whining in module luksbootkeyfile

agaida:
Added myself to copyright


src/modules/luksbootkeyfile/main.py
```diff
@@ -53,7 +54,8 @@ def run():
 
     if not luks_root_passphrase:
         return ("Encrypted rootfs setup error",
-                "Rootfs partition {!s} is LUKS but no passphrase found.".format(luks_root_device))
+                "Rootfs partition {!s}".format(luks_root_device),
+                "is LUKS but no passphrase found.")
```
adriaandegroot: , instead of +, but this is also hard on translations. I'm going to reject this PR, because I've fixed it (based on this PR) in master already.

PR #737, opened 2017-06-02
Fixed pep8 whining in module networkcfg

agaida:
added myself to copyright


src/modules/networkcfg/main.py
```diff
@@ -49,7 +52,10 @@ def run():
             try:
                 shutil.copy(source_network, target_network)
             except FileNotFoundError:
-                libcalamares.utils.debug("Can't copy network configuration files in {}".format(source_network))
+                libcalamares.utils.debug(
+                    "Can't copy network configuration files in",
+                    "{}".format(source_network)
+                    )
```
adriaandegroot: `libcalamares.utils.debug()` only takes one parameter; that `,` should be a +, and you've dropped a space in the string.

PR #739, opened 2017-06-02
Fixed pep8 whining in module grubcfg

agaida:
added myself to copyright

### DIFF
--- a/src/modules/grubcfg/main.py
+++ b/src/modules/grubcfg/main.py
@@ -5,6 +5,7 @@
 #
 #   Copyright 2014-2015, Philip Müller <philm@manjaro.org>
 #   Copyright 2015-2017, Teo Mrnjavac <teo@kde.org>
+#   Copyright 2017, Alf Gaida <agaida@siduction.org>
 #
 #   Calamares is free software: you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
@@ -25,7 +26,8 @@ import re
 
 
 def modify_grub_default(partitions, root_mount_point, distributor):
-    """ Configures '/etc/default/grub' for hibernation and plymouth.
+    """
+    Configures '/etc/default/grub' for hibernation and plymouth.
 
     :param partitions:
     :param root_mount_point:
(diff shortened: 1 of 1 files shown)


src/modules/grubcfg/main.py
```diff
@@ -51,20 +55,26 @@ def modify_grub_default(partitions, root_mount_point, distributor):
             if partition["fs"] == "linuxswap":
                 swap_uuid = partition["uuid"]
 
-            if partition["fs"] == "linuxswap" and "luksMapperName" in partition:
+            if partition[
+                    "fs"] == "linuxswap" and "luksMapperName" in partition:
                 swap_outer_uuid = partition["luksUuid"]
```
adriaandegroot: Please don't split it like this; use

```
if (partition["fs"] == "linuxswap"
    and "luksMapperName" in partition):
```

(possibly with "and" at the end of the line -- I'm not sure what PEP8 says on that front). In any case the way you've line-broken here interferes with reading.

src/modules/grubcfg/main.py
```diff
@@ -51,20 +55,26 @@ def modify_grub_default(partitions, root_mount_point, distributor):
...
 
-            if partition["mountPoint"] == "/" and "luksMapperName" in partition:
-                cryptdevice_params = ["rd.luks.uuid={!s}".format(partition["luksUuid"])]
+            if partition[
+                    "mountPoint"] == "/" and "luksMapperName" in partition:
```
adriaandegroot: Difficult-to read linebreak.

src/modules/grubcfg/main.py
```diff
@@ -159,14 +179,16 @@ def modify_grub_default(partitions, root_mount_point, distributor):
...
 
     :return:
     """
 
     fw_type = libcalamares.globalstorage.value("firmwareType")
 
-    if libcalamares.globalstorage.value("bootLoader") is None and fw_type != "efi":
+    if libcalamares.globalstorage.value(
+            "bootLoader") is None and fw_type != "efi":
         return None
```
adriaandegroot: Please break at the "and"

src/modules/grubcfg/main.py
agaida: ok, you are right again - and afaik using a prefix operator is preferred in pycodestyle, it was postfix in pep8. But it might be that i'm not remembering this right.

PR #742, opened 2017-06-02
Fixed pep8 whining in module displaymanager

agaida:
Added myself to copyright

### DIFF
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -6,6 +6,7 @@
 #   Copyright 2014-2016, Philip Müller <philm@manjaro.org>
 #   Copyright 2014-2015, Teo Mrnjavac <teo@kde.org>
 #   Copyright 2014, Kevin Kofler <kevin.kofler@chello.at>
+#   Copyright 2017, Alf Gaida <agaida@siduction.org>
 #
 #   Calamares is free software: you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
@@ -27,7 +28,9 @@ import libcalamares
 import configparser
 
 
-DesktopEnvironment = collections.namedtuple('DesktopEnvironment', ['executable', 'desktop_file'])
+DesktopEnvironment = collections.namedtuple(
+    'DesktopEnvironment', ['executable', 'desktop_file']
+    )
 
 desktop_environments = [
     DesktopEnvironment('/usr/bin/startkde', 'plasma'),  # KDE Plasma 5
@@ -51,35 +54,46 @@ desktop_environments = [
 
 
 def find_desktop_environment(root_mount_point):
-    """ Checks which desktop environment is currently installed.
+    """
+    Checks which desktop environment is currently installed.
 
     :param root_mount_point:
     :return:
     """
     for desktop_environment in desktop_environments:
-        if os.path.exists(
-            "{!s}{!s}".format(root_mount_point, desktop_environment.executable)) \
-            and os.path.exists("{!s}/usr/share/xsessions/{!s}.desktop".format(root_mount_point,
-                                                                              desktop_environment.desktop_file)):
+        if (os.path.exists("{!s}{!s}".format(
+                    root_mount_point, desktop_environment.executable
+                    )
+                ) and os.path.exists(
+                    "{!s}/usr/share/xsessions/{!s}.desktop".format(
+                        root_mount_point, desktop_environment.desktop_file
+                    )
+                )):
             return desktop_environment
-
     return None
 
 
 def have_dm(dm_name, root_mount_point):
-    """ Checks if display manager is properly installed.
+    """
+    Checks if display manager is properly installed.
 
     :param dm_name:
     :param root_mount_point:
     :return:
     """
-    return os.path.exists(
-        "{!s}/usr/bin/{!s}".format(root_mount_point, dm_name)) or os.path.exists(
-        "{!s}/usr/sbin/{!s}".format(root_mount_point, dm_name))
+    bin_path = "{!s}/usr/bin/{!s}".format(root_mount_point, dm_name)
+    sbin_path = "{!s}/usr/sbin/{!s}".format(root_mount_point, dm_name)
+    return (os.path.exists(bin_path)
+            or os.path.exists(sbin_path)
+            )
 
 
-def set_autologin(username, displaymanagers, default_desktop_environment, root_mount_point):
-    """ Enables automatic login for the installed desktop managers.
+def set_autologin(username,
+                  displaymanagers,
+                  default_desktop_environment,
+                  root_mount_point):
+    """
+    Enables automatic login for the installed desktop managers.
 
     :param username:
     :param displaymanagers:
@@ -103,14 +117,23 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
                 for line in text:
                     if '[daemon]' in line:
                         if do_autologin:
-                            line = "[daemon]\nAutomaticLogin={!s}\nAutomaticLoginEnable=True\n".format(username)
+                            line = (
+                                "[daemon]\n"
+                                "AutomaticLogin={!s}\n"
+                                "AutomaticLoginEnable=True\n".format(username)
+                                )
                         else:
-                            line = "[daemon]\nAutomaticLoginEnable=False\n"
+                            line = (
+                                "[daemon]\n"
+                                "AutomaticLoginEnable=False\n"
+                                )
 
                     mdm_conf.write(line)
         else:
             with open(mdm_conf_path, 'w') as mdm_conf:
-                mdm_conf.write('# Calamares - Configure automatic login for user\n')
+                mdm_conf.write(
+                    '# Calamares - Configure automatic login for user\n'
+                    )
                 mdm_conf.write('[daemon]\n')
 
                 if do_autologin:
@@ -131,14 +154,20 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
                 for line in text:
                     if '[daemon]' in line:
                         if do_autologin:
-                            line = "[daemon]\nAutomaticLogin={!s}\nAutomaticLoginEnable=True\n".format(username)
+                            line = (
+                                "[daemon]\n"
+                                "AutomaticLogin={!s}\n"
+                                "AutomaticLoginEnable=True\n".format(username)
+                                )
                         else:
                             line = "[daemon]\nAutomaticLoginEnable=False\n"
 
                     gdm_conf.write(line)
         else:
             with open(gdm_conf_path, 'w') as gdm_conf:
-                gdm_conf.write('# Calamares - Enable automatic login for user\n')
+                gdm_conf.write(
+                    '# Calamares - Enable automatic login for user\n'
+                    )
                 gdm_conf.write('[daemon]\n')
 
                 if do_autologin:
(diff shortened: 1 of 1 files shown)


src/modules/displaymanager/main.py
```diff
@@ -375,21 +493,44 @@ def run():
...
+                libcalamares.utils.target_env_call(
+                    'chmod', '1770', '/var/lib/mdm'
+                    )
 
             if default_desktop_environment is not None:
-                os.system("sed -i \"s|default.desktop|{!s}.desktop|g\" {!s}/etc/mdm/custom.conf".format(
-                          default_desktop_environment.desktop_file, root_mount_point))
+                os.system(
+                    "sed -i \"s|default.desktop|{!s}.desktop|g\"",
+                    "{!s}/etc/mdm/custom.conf".format(
```
agaida: not so clever - the first string will never get a parameter from .format

src/modules/displaymanager/main.py
```diff
@@ -327,27 +403,53 @@ def run():
...
+                libcalamares.utils.target_env_call(
+                    'chmod', '+r' '/etc/lightdm/lightdm.conf'
+                    )
 
             if default_desktop_environment is not None:
-                os.system("sed -i -e \"s/^.*user-session=.*/user-session={!s}/\" {!s}/etc/lightdm/lightdm.conf".format(
-                          default_desktop_environment.desktop_file, root_mount_point))
+                os.system(
+                    "sed -i -e \"s/^.*user-session=.*/user-session={!s}/\"",
+                    "{!s}/etc/lightdm/lightdm.conf".format(
```
agaida: first !s will never be substituted

philmmanjaro: Why? First is `user-session={!s}` with `default_desktop_environment.desktop_file` and second is `{!s}/etc/lightdm/lightdm.conf` with `root_mount_point`. What do you consider to be wrong with the first one?

src/modules/displaymanager/main.py
philmmanjaro: Similar as above ...

src/modules/displaymanager/main.py
adriaandegroot: it's a call to os.system() with two parameters: a string `"sed .."` and a string `"{!s}/etc .."` and the format() method is only called on the second. The comma at the end of line 440 should be a + and there might be parens needed as well.

src/modules/displaymanager/main.py
```diff
@@ -147,19 +177,37 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
...
 
             if default_desktop_environment is not None:
-                os.system("echo \"XSession={!s}\" >> {!s}/var/lib/AccountsService/users/{!s}".format(
-                          default_desktop_environment.desktop_file, root_mount_point, username))
-
-            os.system("echo \"Icon=\" >> {!s}/var/lib/AccountsService/users/{!s}".format(
-                      root_mount_point, username))
+                os.system(
+                    "echo \"XSession={!s}\" >>",
+                    "{!s}/var/lib/AccountsService/users/{!s}".format(
```
adriaandegroot: Here's another misplaced `,` which will lead for formatting failures.

src/modules/displaymanager/main.py
```diff
@@ -147,19 +177,37 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
                 else:
                     gdm_conf.write('AutomaticLoginEnable=False\n')
 
-        if do_autologin and os.path.exists("{!s}/var/lib/AccountsService/users".format(root_mount_point)):
-            os.system("echo \"[User]\" > {!s}/var/lib/AccountsService/users/{!s}".format(root_mount_point, username))
+        if do_autologin and os.path.exists(
+                "{!s}/var/lib/AccountsService/users".format(root_mount_point)):
+            os.system(
+                "echo \"[User]\" >",
```
adriaandegroot: Not particularly related to this review, but it seems kind of dodgy to me to involve os.system() to do what is basically

```
with open("{!s}/var/lib/AccountsService/users/{!s}".format(root_mount_point, username), "w") as f:
    f.write("[User]\n")
```

Unless there's some permissions-trickery going on here, I think this whole module would benefit from a re-write to clean up the os.system() calls.

src/modules/displaymanager/main.py
agaida: right, you will find this bug in some other places, i will replace it. Thats why i marked the PR WIP (work in progress)

src/modules/displaymanager/main.py
agaida: Good point, should be handled in a new PR

src/modules/displaymanager/main.py
```diff
@@ -51,35 +54,46 @@
...
     :param root_mount_point:
     :return:
     """
     return os.path.exists(
-        "{!s}/usr/bin/{!s}".format(root_mount_point, dm_name)) or os.path.exists(
-        "{!s}/usr/sbin/{!s}".format(root_mount_point, dm_name))
+        "{!s}/usr/bin/{!s}".format(root_mount_point, dm_name)
+        ) or os.path.exists(
+            "{!s}/usr/sbin/{!s}".format(root_mount_point, dm_name)
+            )
```
adriaandegroot: If pycodestyle agrees, I'd typically do this as 

```
bin_path = "{!s}/usr/bin/{!s}".format(root_mount_point, dm_name)
sbin_path = "{!s}/usr/sbin/{!s}".format(root_mount_point, dm_name)
return (
    os.path.exists(bin_path)
    or os.path.exists(sbin_path)
    )
```

That makes clearer what is being tested, exactly.

src/modules/displaymanager/main.py
```diff
@@ -103,14 +117,23 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
                 for line in text:
                     if '[daemon]' in line:
                         if do_autologin:
-                            line = "[daemon]\nAutomaticLogin={!s}\nAutomaticLoginEnable=True\n".format(username)
+                            line = (
+                                "[daemon]\n",
+                                "AutomaticLogin={!s}\n",
```
adriaandegroot: This replaces one string with a 3-tuple, and will break in the write() call later. You can use multiple string literals here.

src/modules/displaymanager/main.py
```diff
@@ -103,14 +117,23 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
...
+                                "AutomaticLoginEnable=True\n".format(username)
+                                )
                         else:
-                            line = "[daemon]\nAutomaticLoginEnable=False\n"
+                            line = (
+                                "[daemon]\n",
```
adriaandegroot: Same -- drop the ,

src/modules/displaymanager/main.py
```diff
@@ -131,14 +154,20 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
                 for line in text:
                     if '[daemon]' in line:
                         if do_autologin:
-                            line = "[daemon]\nAutomaticLogin={!s}\nAutomaticLoginEnable=True\n".format(username)
+                            line = (
+                                "[daemon]\n",
```
adriaandegroot: Commas

src/modules/displaymanager/main.py
```diff
@@ -147,19 +176,40 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
...
-        if do_autologin and os.path.exists("{!s}/var/lib/AccountsService/users".format(root_mount_point)):
-            os.system("echo \"[User]\" > {!s}/var/lib/AccountsService/users/{!s}".format(root_mount_point, username))
+        if (do_autologin
+                and os.path.exists("{!s}/var/lib/AccountsService/users".format(
+                    root_mount_point
+                    )
+                )):
+            os.system(
+                "echo \"[User]\" > "
+                + "{!s}/var/lib/AccountsService/users/{!s}".format(
```
adriaandegroot: This is one of those cases where the + is superfluous and we could go for multiple string literals again, but since we've already agreed that this bit needs a re-write on account of using echo(1), let's leave it.

src/modules/displaymanager/main.py
```diff
@@ -147,19 +176,40 @@ def set_autologin(username, displaymanagers, default_desktop_environment, root_m
...
+                "echo \"Icon=\" >> "
+                + "{!s}/var/lib/AccountsService/users/{!s}".format(
+                    root_mount_point, username
+                    )
+                )
 
     if "kdm" in displaymanagers:
         # Systems with KDM as Desktop Manager
-        kdm_conf_path = os.path.join(root_mount_point, "usr/share/config/kdm/kdmrc")
+        kdm_conf_path = os.path.join(
```
adriaandegroot: Given that all the other path manipulations are just doing string concatenation UNIX-style, I'm surprised to see os.path.join here. Not your code, and let's leave it, but there's some consistency missing.

src/modules/displaymanager/main.py
```diff
@@ -293,18 +366,24 @@ def run():
         displaymanagers = libcalamares.globalstorage.value("displayManagers")
 
     if displaymanagers is None:
-        return "No display managers selected for the displaymanager module.", \
-               "The displaymanagers list is empty or undefined in both globalstorage and displaymanager.conf."
+        return (
+            "No display managers selected for the displaymanager module.",
+            "The displaymanagers list is empty or undefined in both",
```
adriaandegroot: Comma bad.

src/modules/displaymanager/main.py
```diff
@@ -327,27 +406,55 @@ def run():
...
-                if libcalamares.utils.target_env_call(['getent', 'passwd', 'lightdm']) != 0:
-                    libcalamares.utils.target_env_call(['useradd', '-c', '"LightDM Display Manager"',
-                                                    '-u', '620', '-g', 'lightdm', '-d', '/var/run/lightdm',
-                                                    '-s', '/usr/bin/nologin', 'lightdm'])
-
-                libcalamares.utils.target_env_call(['passwd', '-l', 'lightdm'])
-                libcalamares.utils.target_env_call(['chown', '-R', 'lightdm:lightdm', '/run/lightdm'])
-                libcalamares.utils.target_env_call(['chmod', '+r' '/etc/lightdm/lightdm.conf'])
+                libcalamares.utils.target_env_call(
+                    'mkdir', '-p', '/run/lightdm'
```
adriaandegroot: You have lost the [] list brackets here and in the following bits .. I don't know if target_env_call wants to be called that way.

src/modules/displaymanager/main.py
```diff
@@ -327,27 +406,55 @@ def run():
...
+                    + "{!s}/etc/lightdm/lightdm.conf".format(
+                        default_desktop_environment.desktop_file,
+                        root_mount_point
+                        )
+                    )
 
             if default_desktop_environment.desktop_file == "deepin":
-                os.system("sed -i -e \"s/^.greeter-session=.*/greeter-session=lightdm-deepin-greeter/\" {!s}/etc/lightdm/lightdm.conf".format(
-                          root_mount_point))
+                os.system(
```
adriaandegroot: Not your code, but we should consider moving away from os.system() to subprocess, that's much easier to do proper argument passing with.

src/modules/displaymanager/main.py
```diff
@@ -357,16 +464,32 @@ def run():
...
-                    libcalamares.utils.target_env_call(['groupadd', '-g', '120', 'gdm'])
-
-                if libcalamares.utils.target_env_call(['getent', 'passwd', 'gdm']) != 0:
-                    libcalamares.utils.target_env_call(['useradd', '-c', '"Gnome Display Manager"',
-                                                    '-u', '120', '-g', 'gdm', '-d', '/var/lib/gdm',
-                                                    '-s', '/usr/bin/nologin', 'gdm'])
-
-                libcalamares.utils.target_env_call(['passwd', '-l', 'gdm'])
-                libcalamares.utils.target_env_call(['chown', '-R', 'gdm:gdm', '/var/lib/gdm'])
+                if libcalamares.utils.target_env_call(
```
adriaandegroot: List brackets have gone missing.

src/modules/displaymanager/main.py
```diff
@@ -375,21 +498,44 @@ def run():
...
-
-                if libcalamares.utils.target_env_call(['getent', 'passwd', 'mdm']) != 0:
-                    libcalamares.utils.target_env_call(['useradd', '-c', '"Linux Mint Display Manager"',
-                                                    '-u', '128', '-g', 'mdm', '-d', '/var/lib/mdm',
-                                                    '-s', '/usr/bin/nologin', 'mdm'])
-
-                libcalamares.utils.target_env_call(['passwd', '-l', 'mdm'])
-                libcalamares.utils.target_env_call(['chown', 'root:mdm', '/var/lib/mdm'])
-                libcalamares.utils.target_env_call(['chmod', '1770', '/var/lib/mdm'])
+                if libcalamares.utils.target_env_call(
```
adriaandegroot: List brackets

src/modules/displaymanager/main.py
```diff
@@ -398,16 +544,31 @@ def run():
     if "lxdm" in displaymanagers:
         if have_dm("lxdm", root_mount_point):
             if enable_basic_setup:
-                if libcalamares.utils.target_env_call(['getent', 'group', 'lxdm']) != 0:
-                    libcalamares.utils.target_env_call(['groupadd', '--system', 'lxdm'])
-
-                libcalamares.utils.target_env_call(['chgrp', '-R', 'lxdm', '/var/lib/lxdm'])
-                libcalamares.utils.target_env_call(['chgrp', 'lxdm', '/etc/lxdm/lxdm.conf'])
-                libcalamares.utils.target_env_call(['chmod', '+r', '/etc/lxdm/lxdm.conf'])
+                if libcalamares.utils.target_env_call(
```
adriaandegroot: List brackets

src/modules/displaymanager/main.py
```diff
@@ -416,23 +577,45 @@ def run():
...
             if enable_basic_setup:
-                if libcalamares.utils.target_env_call(['getent', 'group', 'kdm']) != 0:
-                    libcalamares.utils.target_env_call(['groupadd', '-g', '135', 'kdm'])
-
-                if libcalamares.utils.target_env_call(['getent', 'passwd', 'kdm']) != 0:
-                    libcalamares.utils.target_env_call(['useradd', '-u', '135', '-g', 'kdm', '-d',
-                                                    '/var/lib/kdm', '-s', '/bin/false', '-r', '-M', 'kdm'])
-
-                libcalamares.utils.target_env_call(['chown', '-R', '135:135', 'var/lib/kdm'])
+                if libcalamares.utils.target_env_call(
```
adriaandegroot: List brackets

src/modules/displaymanager/main.py
agaida: full ack

src/modules/displaymanager/main.py
agaida: no, my fault  env_call expect a list

Issue #756, opened 2017-06-27
lightdm greeter not configured


oberon2007:
It turns out that [the condition](https://github.com/calamares/calamares/blob/master/src/modules/displaymanager/main.py#L449) we introduced a while ago to configure **lightdm-deepin-greeter** is apparently not met any longer.
Today I built a Manjaro Deepin ISO and on the targt install **lightdm.conf** is unconfigured while I would expect the displaymanager module to set `greeter-session=lightdm-deepin-greeter` when **/usr/bin/startdde** is present.
Apart from that, since there are more than two different greeters possible for lightdm like for example **slick-greeter** I suggest that in order to detect the correct lightdm greeter to configure we change strategy and check for greeter launchers present in **/usr/share/xgreeters** instead of DE launcher  - similar to what [**manjaro-architect does**](https://github.com/Chrysostomus/manjaro-architect/blob/master/lib/util-desktop.sh#L238)

philmmanjaro:
You may add some debug lines to the python file on your ISO to see if the routine is working.

Example: `libcalamares.utils.debug("kdm selected but not installed")`

Also it would be good to actually link to the **manjaro-architect** code, to showcase what you mean :smile:

oberon2007:
OOps, 😜 here, like that:
https://github.com/Chrysostomus/manjaro-architect/blob/master/lib/util-desktop.sh#L238

oberon2007:
Ok. I found the issue with deepin. It's simply an extra whitespace that makes the `sed` ineffective 😉 
I'll send a PR.
Regarding slick-greeter and possible other lightdm greeters I'll try to come up with something proper... :)

philmmanjaro:
The issue was introduced by this: https://github.com/calamares/calamares/commit/f9695cabac242fbbbaad790b7e1a9bfcf590ab7b

PR #758, opened 2017-06-28
[displaymanager] check for installed lightdm greeter and configure accordingly

oberon2007:
- no matter if for example **slick-greeter** is installed as "lightdm-slick-greeter" or whatever, launcher in **/usr/share/xgreeters** will be used. 
  tested with **lightdm-deepin-greeter** and **slick-greeter**.
- plus some little simplifications
- added Copyright

https://github.com/calamares/calamares/issues/756

### DIFF
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -3,10 +3,11 @@
 #
 # === This file is part of Calamares - <http://github.com/calamares> ===
 #
-#   Copyright 2014-2016, Philip Müller <philm@manjaro.org>
+#   Copyright 2014-2017, Philip Müller <philm@manjaro.org>
 #   Copyright 2014-2015, Teo Mrnjavac <teo@kde.org>
 #   Copyright 2014, Kevin Kofler <kevin.kofler@chello.at>
 #   Copyright 2017, Alf Gaida <agaida@siduction.org>
+#   Copyright 2017, Bernhard Landauer <oberon@manjaro.org>
 #
 #   Calamares is free software: you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
@@ -403,6 +404,10 @@ def run():
     # setup lightdm
     if "lightdm" in displaymanagers:
         if have_dm("lightdm", root_mount_point):
+            lightdm_conf_path = os.path.join(
+                root_mount_point, "etc/lightdm/lightdm.conf"
+                )
+
             if enable_basic_setup:
                 libcalamares.utils.target_env_call(
                     ['mkdir', '-p', '/run/lightdm']
@@ -440,21 +445,42 @@ def run():
             if default_desktop_environment is not None:
                 os.system(
                     "sed -i -e \"s/^.*user-session=.*/user-session={!s}/\" "
-                    "{!s}/etc/lightdm/lightdm.conf".format(
+                    "{!s}".format(
                         default_desktop_environment.desktop_file,
-                        root_mount_point
+                        lightdm_conf_path
                         )
                     )
 
-            if default_desktop_environment.desktop_file == "deepin":
-                os.system(
-                    "sed -i -e \"s/^.greeter-session=.*"
-                    "/greeter-session=lightdm-deepin-greeter/\" "
-                    "{!s}/etc/lightdm/lightdm.conf".format(
-                        root_mount_point
-                        )
-                    )
+            # configure lightdm-greeter
+            greeter_path = os.path.join(
+                root_mount_point, "usr/share/xgreeters"
+                )
 
+            if (os.path.exists(greeter_path)):
+                greeter_configured = False
+
+                # configure first found lightdm-greeter
+                for entry in os.scandir(greeter_path):
+                    if entry.name.endswith('.desktop') \
+                            and not greeter_configured:
+                        greeter = entry.name.split('.')[0]
+                        libcalamares.utils.debug(
+                            "found greeter {!s}".format(greeter)
+                            )
+                        os.system(
+                            "sed -i -e \"s/^.*greeter-session=.*"
+                            "/greeter-session={!s}/\" {!s}".format(
+                                greeter,
+                                lightdm_conf_path
+                                )
+                            )
+                        libcalamares.utils.debug(
+                            "{!s} configured as greeter.".format(greeter)
+                            )
+                        greeter_configured = True
+
+                if not greeter_configured:
+                    return ("No lightdm greeter installed.")
         else:
             libcalamares.utils.debug("lightdm selected but not installed")
             displaymanagers.remove("lightdm")


src/modules/displaymanager/main.py
```diff
@@ -403,6 +401,11 @@ def run():
     # setup lightdm
     if "lightdm" in displaymanagers:
         if have_dm("lightdm", root_mount_point):
+            global lightdm_conf_path
```
adriaandegroot: Why is this a global? The rest of the diff suggests this variable is local to run()

src/modules/displaymanager/main.py
```diff
@@ -439,22 +442,30 @@ def run():
...
                         )
                     )
 
+            # configure lightdm-greeter
+            greeter_path = os.path.join(
+                root_mount_point, "usr/share/xgreeters"
+                )
+            if (os.path.exists(greeter_path)):
+                greeter = os.listdir(greeter_path)[0].split('.')[0]
+                libcalamares.utils.debug("configure {!s}".format(greeter))
```
adriaandegroot: Please be a little more descriptive, e.g. "configure lightdm greeter {!s}".

src/modules/displaymanager/main.py
```diff
@@ -439,22 +442,30 @@ def run():
...
+                        lightdm_conf_path
                         )
                     )
 
+            # configure lightdm-greeter
+            greeter_path = os.path.join(
+                root_mount_point, "usr/share/xgreeters"
+                )
+            if (os.path.exists(greeter_path)):
+                greeter = os.listdir(greeter_path)[0].split('.')[0]
```
adriaandegroot: Are you sure the first file found in the directory is always the right one? Perhaps the list needs some filtering (e.g. for executable, or .desktop files). I don't know lightdm and the xgreeters setup well enough to judge, though.

src/modules/displaymanager/main.py
```diff
@@ -439,22 +442,30 @@ def run():
...
+                libcalamares.utils.debug("configure {!s}".format(greeter))
+                with open(lightdm_conf_path, 'r') as lightdm_conf:
+                    text = lightdm_conf.readlines()
+
+                with open(lightdm_conf_path, 'w') as lightdm_conf:
```
adriaandegroot: Here you're doing sed -i in Python; it feels weird (but not wrong) to have two really different ways of replacing a line in a text file, within one Python module.

src/modules/displaymanager/main.py
philmmanjaro: On Manjaro, the greeters conflict normally. However, this we have to check once more. On my end I get this:

```
ls /usr/share/xgreeters
lightdm-gtk-greeter.desktop
```

src/modules/displaymanager/main.py
philmmanjaro: Since `greeter` would be `lightdm-gtk-greeter` the line would be then `configure lightdm-gtk-greeter` ...

src/modules/displaymanager/main.py
philmmanjaro: I was wrong. You can install multiple greeters. So a filtering is needed ...

```
ls /usr/share/xgreeters
lightdm-gtk-greeter.desktop  lightdm-slick-greeter.desktop
```

adriaandegroot: My worry was more "what if someone adds a README to that dir", but having multiple greeters installed is also an issue. But .. wouldn't that be kind of weird on system install? Seems like an overkill situation to install two things and then pick one of them. That makes my README-worry silly as well: Don't Do That Then.

philmmanjaro: Right. Normally we only install one greeter. So we can add a check to see if it is a actually a **.desktop** file or not.

src/modules/displaymanager/main.py
philmmanjaro: It was needed, since the value was reused at a second point. However I've fixed it now by readding the original code: https://github.com/calamares/calamares/pull/758/commits/e53d4b84dc138f02e6bca2f2591c1b862904d423

adriaandegroot:
Looking good! Phil, please merge into Cala master when you're done.

philmmanjaro:
Still some is wrong, as it errors out with: `Expecting an object of type tuple; got an object of type str instead` ...

src/modules/displaymanager/main.py
```diff
@@ -440,21 +445,34 @@ def run():
...
+                greeter = os.listdir(greeter_path)[0].split('.')[0]
+                greeter_desktop = os.path.join(
+                    greeter_path,
+                    greeter,
+                    ".desktop"
                     )
 
+                if (os.path.exists(greeter_desktop)):
+                    libcalamares.utils.debug("configure {!s}".format(greeter))
+                    "sed -i -e "
```
adriaandegroot: There should be os.system() there?

src/modules/displaymanager/main.py
oberon2007: yes, it's also used in [set_autologin()](https://github.com/calamares/calamares/blob/master/src/modules/displaymanager/main.py#L264) and was trying to avoid repetitive code...

src/modules/displaymanager/main.py
```diff
@@ -440,21 +445,36 @@ def run():
...
+                    libcalamares.utils.debug("configure {!s}".format(greeter))
+                    os.system(
+                        "sed -i -e "
+                        "\"s/^.*greeter-session=.*/greeter-session={!s}/\" "
+                        "{!s}".format(
+                            greeter,
+                            lightdm_conf_path
+                            )
+                        )
+                else:
```
oberon2007: @philmmanjaro reall? Actually this belongs to condition `if (os.path.exists(greeter_path)):`

src/modules/displaymanager/main.py
```diff
@@ -440,21 +445,36 @@ def run():
...
+                    return ("No lightdm greeter installed.")
```
oberon2007: @philmmanjaro really? Actually this error belongs to condition `if (os.path.exists(greeter_path)):`

philmmanjaro:
We still have the issue with `Expecting an object of type tuple; got an object of type str instead` ...

src/modules/displaymanager/main.py
oberon2007: If multiple greeters were installed we'd have no way of knowing which one to configure as the desired one - would need a config for that - in any case it's unlikely that more than one would be present as it doesn't make so much sense... :wink:

adriaandegroot:
On Wednesday 28 June 2017 03:36:28 Philip Müller wrote:
> We still have the issue with `Expecting an object of type tuple; got an
> object of type str instead` ...

Any more backtrace or output than that? Surely there's a line-number, too?

src/modules/displaymanager/main.py
oberon2007: The file was containing both approaches already - so I thought I'd pick the python native one - although ... it is awkward ... 😜 😆

philmmanjaro:
Latest code revision uses a scan and stops at the fist greeter found. Code it tested on my end. Should be fine.

PR #759, opened 2017-07-03
Fix unknown partition-type Fat32 at mount and in fstab

Psy-Kai:
Only the string "fat32" and "fat16" will be replaced with vfat. If an
case sensitive "Fat32" some problems occure:
- mount: partition cannot be mounted (e.g. a fat32 efi partition)
- fstab: system won't even boot because fstab does not know the type "Fat32"

### DIFF
--- a/src/modules/fstab/main.py
+++ b/src/modules/fstab/main.py
@@ -224,7 +224,7 @@ class FstabGenerator(object):
 
     def generate_fstab_line_info(self, partition):
         """ Generates information for each fstab entry. """
-        filesystem = partition["fs"]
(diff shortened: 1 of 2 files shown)


adriaandegroot:
Looks good, but can you explain how "Fat32" would end up in the fs-type string? Does it come from the user, or is it a configuration issue?

Psy-Kai:
Both modified lines return `Fat32` instead of `fat32`. I ran in the described problems with the current manjaro-kde-images (on my notebook with uefi and a qemu vm with uefi). I just wanted calamares to use the whole drive so I did no manual partitioning.

PR #763, opened 2017-07-09
Add "MB" suffix to the "edit partition" size box.

shainer:
Makes it easier to know which unit Calamares is using when
resizing a partition. The "Create partition" dialog has
it already.


kkofler:
Are those really MB or are they actually MiB?

shainer:
Uhm judging from the Calamares and Kpmcore code it's MiB, right. Let me fix it in both dialogs.

PR #771, opened 2017-08-06
Correct update-initramfs command

n3rdopolis:
This allows it to create a new initramfs, and replace it.
This will better ensure that calamares generated initramfs is used

### DIFF
--- a/src/modules/initramfs/main.py
+++ b/src/modules/initramfs/main.py
@@ -27,7 +27,8 @@ def run():
 
     :return:
     """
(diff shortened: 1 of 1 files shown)


adriaandegroot:
You are adding the -t flag to the invocation of update-initramfs, but what is that correcting? Why would there be an image that needs to be taken over? It seems like an innocuous change, but I don't understand why it might be needed.

n3rdopolis:
In case if (...looks like remastersys leaves the original initramfs in /boot...) there is an initramfs already

PR #776, opened 2017-08-09
better looks and logo size

blackPantherOS:
We use these sizes in blackPanther OS, I think this is a better looks

### DIFF
--- a/src/calamares/CalamaresWindow.cpp
+++ b/src/calamares/CalamaresWindow.cpp
@@ -76,7 +76,7 @@ CalamaresWindow::CalamaresWindow( QWidget* parent )
 
     QBoxLayout* sideLayout = new QVBoxLayout;
     sideBox->setLayout( sideLayout );
-    sideBox->setFixedWidth( qBound( 100, CalamaresUtils::defaultFontHeight() * 12, w < windowPreferredWidth ? 100 : 190 ) );
+    sideBox->setFixedWidth( qBound( 100, CalamaresUtils::defaultFontHeight() * 11, w < windowPreferredWidth ? 100 : 190 ) );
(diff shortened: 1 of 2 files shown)


adriaandegroot:
I don't think we should start pushing and pulling on the window sizes from within C++ code; what looks better for you looks worse for others, depending on default themes, fonts, etc. It will also break a bunch of automated downstream tests that use openQA to check for rendering defects, for instance.

I would accept a PR that moves this into a configuration file -- e.g. `branding.conf`, either a new top-level key *window* or under *style*, setting minw, minh, sidebarproportion.

bill-auger:
i will add that this image entity 'Branding::ProductIcon' is currently the same that is used for the WM/DE icon which is always square and also much smaller - the consensus is that the original intention was for this sidebar image to use the 'Branding::ProductLogo' entity - note that it is rendered inside of a 'logoLabel' inside of a 'logoLayout' - see PR #826 

if this PR is accepted as is, then the intended input for 'Branding::ProductIcon' would presumably be rectangular which would make the WM/DE icon to be distorted

it has been suggewsted on IRC that perhaps a new entry be added to branding.desc and a new entity in Branding.cpp such as: 'ProductLogoWide' for an explicitly rectangular input image and allow this feature to be a configuration optoin as suggested to switch between using the 'ProductLogo' (square) and the 'ProductLogoWide' (rectangular) sources

adriaandegroot:
Closing this one because of no response to the request for changes; it may also be superceded by @raurodse  and the CSS-styling changes he is proposing.

bill-auger:
i presume that is PR #961 - i am still interested in this because i have been carrying a patch for this - i will subscribe to #961

Issue #730, opened 2017-05-27
broken swap with dracut  / mkinitcpio + Luks swap 


abucodonosor:
Hello,

when using Luks  entry for bootloader config and fstab/crypttab are wrong.

calamares would generate this 

grub:
... resume=UUID=the_swap_uuid rd.luks.uuid=the_luks_uuid

crypttab:
luks-the_luks_uuid        key          swap

fstab:
UUID=the_swap_uuid

with LUKS the resume= cmdline need point to the device with luks-swap eg: /dev/mapper/luks-xxxx

crypttab need swap,plain,luks as option so systemd or other generators can open it

fstab want swap to be /dev/mapper/luks-xxxx so systemd or whatever init manage swap can mount it

I would suggest to not mix all these but on LUKS use root = && resume=  /dev/mapper  and same for fstab

Regards

demmm:
Possible related to https://github.com/calamares/calamares/issues/606

kkofler:
This is really a `dracut` bug. `resume=UUID=…` is supposed to work and to be handled by the initramfs. `dracut` fails to handle `resume=…`, so the kernel sees it, and the kernel supports only direct /dev nodes.

abucodonosor:
@kkofler 

resume=UUID= does work here for !LUKS 'as' is.. the problem is not dracut but the logic how cala handles
luks installs.

on luks && dracut cala does 
swap_uuid = ""
swap_outer_uuid = "" <-- for dracut

check swap && if found set resume=UUID=real-device-swap-UUID <-- which is not same as the LUKS swap UUID ofc

check LUKS and if found && dracut found set rd.luks.uuid=the-luks-uuid < all fine

missing check on if swap && LUKS in bootloader , fstab , grubcfg modules  but then it go an writes the following filese

fstab
UUID=root-one
UUID=broke-swap-uuid

kernel cmd

.... resume=UUID=broken-one rd.luks.uiid=right.one

now you boot up and dracut does the 'right' thing in the initrd but your 'init' will crap out
with this config since it won't be able to handle your swap.. the swapon call will try to access
your /dev/foo while should swapon the 'open' crypt swap partiton which need be the /dev/mapper/luks-uuid 

and ofc resume won't work there's wrong swap + no swap being enable in userspace

all that is true for all kind supported combos.

is really simple to reproduce and you don't need cala to see where your the bugs lies here.

create a swap luks partition and add the way cala doing now  and reboot your box.
[...]

PR #800, opened 2017-09-13
fstab: fix btrfs with LUKS and systemd we forgot to fix that in #730 …

abucodonosor:
…tested with efi/BIOS+LUKS+btrfs , efi/BIOS+btrfs

### DIFF
--- a/src/modules/fstab/main.py
+++ b/src/modules/fstab/main.py
@@ -225,6 +225,7 @@ class FstabGenerator(object):
     def generate_fstab_line_info(self, partition):
         """ Generates information for each fstab entry. """
         filesystem = partition["fs"].lower()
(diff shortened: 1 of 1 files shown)


src/modules/fstab/main.py
```diff
@@ -252,17 +253,28 @@ def generate_fstab_line_info(self, partition):
...
-                mount_point=mount_point,
-                fs=filesystem,
-                options=",".join(
-                    ["subvol={}".format(partition["subvol"]), options]
-                    ),
-                check=check,
-                )
-
-        if "luksMapperName" in partition:
+            if has_luks:
```
adriaandegroot: There must be a better way. By now we have four giant `return dict()` statements, which differ only in details.

PR #801, opened 2017-09-13
CreateUserJob: do not force any autologin group

abucodonosor:
There is _no_ need to force folks haing a random group
 bc that is a Distro think. SDDM/GDM works just fine without
 having a group for the user to autologin.
 Just setup a group in users.conf .. is why we have a configuration
 option for that.

### DIFF
--- a/src/modules/users/CreateUserJob.cpp
+++ b/src/modules/users/CreateUserJob.cpp
@@ -116,12 +116,11 @@ CreateUserJob::exec()
         QString autologinGroup;
         if ( gs->contains( "autologinGroup" ) &&
              !gs->value( "autologinGroup" ).toString().isEmpty() )
-            autologinGroup = gs->value( "autologinGroup" ).toString();
-        else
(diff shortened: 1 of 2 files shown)


src/modules/users/CreateUserJob.cpp
```diff
@@ -116,12 +116,11 @@ CreateUserJob::exec()
         QString autologinGroup;
         if ( gs->contains( "autologinGroup" ) &&
              !gs->value( "autologinGroup" ).toString().isEmpty() )
-            autologinGroup = gs->value( "autologinGroup" ).toString();
-        else
-            autologinGroup = QStringLiteral( "autologin" );
-
-        CalamaresUtils::System::instance()->targetEnvCall( { "groupadd", autologinGroup } );
-        defaultGroups.append( QString( ",%1" ).arg( autologinGroup ) );
+	{
```
adriaandegroot: This entire block needs to be indented so that it aligns with the `if()`. Please run `ci/calamaresstyle` to stick with the coding conventions.

PR #825, opened 2017-09-30
Remove unnecessary KIO and KIconThemes dependencies.

stikonas:
Still compiles


abucodonosor:
There is src/modules/partition/CMakeLists.txt , README.md  and  Dockerfile to fix too

PR #826, opened 2017-10-03
use orphaned Branding::ProductLogo as sidebar image

bill-auger:
the 'productLogo' field in src/branding/default/branding.desc is loaded into the 'Branding::ProductLogo' entity in src/libcalamaresui/Branding.cpp but never actually used anywhere in the program - the image entity 'Branding::ProductIcon' that is currently used for the sidebar is the same one that is used for the much smaller WM/DE icon  - it appears that the original intention was for the sidebar image to use the 'Branding::ProductLogo' entity and that the current  'Branding::ProductIcon' was probably a typo - note that it is rendered inside of a 'logoLabel' inside of a 'logoLayout'


adriaandegroot:
Documentatiuon has also been added.

PR #828, opened 2017-10-06
Add support for performing operations on LVM logical volumes.

stikonas:
At the moment there is no UI to obtain LVM LV name, so it is hardcoded to rootfs


stikonas:
Unfortunately, right now there is no GUI to obtain LV partition name. So I just added rootfs for testing.

stikonas:
Something similar is required for other jobs as well: ResizePartitionJob, DeletePartitionJob

adriaandegroot:
Can we complete this, clean it up, and have something that isn't just for testing before merging it?

stikonas:
Fixed LVM LV naming issue. But this will need some other changes in kpmcore (or calamares). KPMcore has protected access to some partition functions (in particular setPartitionPath). And since Calamares uses it's own jobs, we don't have access to setPartitionPath.

stikonas:
This will need kpmcore 3.3. Then it will compile

adriaandegroot:
We'll just have to remember that kpmcore 3.2 branch in Calamares is for kpmcore 3.3, and not merge this yet.

PR #834, opened 2017-10-12
Fix typos in partition module files

isaagar:


philmmanjaro:
Why do you see an typo here? Those are the marked hot keys for keyboard shortcuts to achieve tasks without using the mouse. Defined marked character made sense. By redefining those we have to see if we have a benefit gained out of it. I would NACK it for now as I don't see a need for it. Thoughts @adriaandegroot

kkofler:
Have you checked that your changed keyboard shortcuts are non-conflicting? (There is usually a reason to not choose the first character as the shortcut.) Or do you just want a Hacktoberfest t-shirt? ;)

isaagar:
@philmmanjaro oops, didn't know . 
@kkofler nope  :), while translating to my language they seem to be conflicting.
Should we add an extra character at end of word as hot key in bracket like "Encrypt (E)" ?

kkofler:
Every language is responsible for setting its own non-conflicting hotkeys. You should try to use the first characters where possible, otherwise other characters in the word. Adding something like `(&E)` as you propose can be done, but only as a last resort if nothing else works in your language (and then, please add it only in your translation, the original English does not need it).

isaagar:
@kkofler ok, got it. Thanks.

PR #835, opened 2017-10-16
use private env var for travis IRC notify channel

bill-auger:
this will (hopefully) prevent downstreams CI from notifying the upstream IRC channel

NOTE: in order to notify IRC, someone from each fork with access to the travis account will need to declare an environment variable in the travis control panel

for the upstream that would be like:

IRC_NOTIFY_CHANNEL with a value of chat.freenode.net#calamares

unrelated: but i noticed some warnings with the .travis.yml
```
$ travis lint ./.travis.yml
Warnings for ./.travis.yml:
[x] in language section: does not support multiple values, dropping python
[x] specified python, but setting is not relevant for cpp
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ services:
 
 notifications:
   irc:
-    - "chat.freenode.net#calamares"
+    - "$IRC_NOTIFY_CHANNEL"
 
 install:
   - docker build -t calamares .


bill-auger:
i looked into this a bit more - it does appear that travis has no suupport for this use case - github can post to IRC directly but the it is triggered by nearly all events except that it doesnt seem to announce the travis status at all - it looks like the only way to receive the travis announcements without specifying the channel in the source code is to use github webhooks - this is very fine grained but it requires an IRC bot that is capable of receiving and parsing the HTTP POST requests from github - these will have this form:

```
{
  "id": 1664069224,
  "sha": "ee54e017eaeb3852590932f2d2dc2f6b149e2b85",
  "name": "bill-auger/deleteme",
  "target_url": "https://travis-ci.org/bill-auger/deleteme/builds/291806588?utm_source=github_status&utm_medium=notification",
  "context": "continuous-integration/travis-ci/push",
  "description": "The Travis CI build failed",
  "state": "failure",
....
```

abucodonosor:
https://github.com/manjaro/calamares/issues/26

Issue #839, opened 2017-10-23
Support installation in Virtualbox in EFI mode


tpgxyz:
#### Submission type

  - [x] Bug report
  - [] Feature Request

#### Info regarding which version of Calamares is used, which Distribution

> 3.1.6

#### Provide information on how the disks are set up, in detail, with full logs of commands issued 

> …

#### What do you expect to have happen when Calamares installs?

> Calamares creates symlink to /EFI/BOOT/BOOTX64.EFI

#### Describe the issue you encountered

> Install ISO with Calamares on Virtualbox in EFI mode.  Boot installed system in EFI mode. Watch system not boot. Based on https://wiki.archlinux.org/index.php/VirtualBox#Installation_in_EFI_mode

#### Steps to reproduce the problem

> Install ISO with calamares in EFI mode.
> Boot installed system in EFI mode
> Observer Vbox whines from missing /EFI/BOOT/BOOTX64.EFI

#### Include the installation.log:

> …

itchka:
To fill this out a bit. VirtualBox does not store efivars from one session to the next. If you set auto reboot in Calamares at the end the system will reboot but if you kill the install it will not. Currently the install EFI structure is this:-
efi/EFI
├── boot
│   └── grubx64.efi
└── openmandriva
    ├── AllocPages.efi
    ├── debughook.efi
    ├── drv0.efi
    ├── drv0_use.efi
    ├── exit.efi
    ├── FreePages.efi                                                                                                                                                                                                                                        
    ├── grub.cfg                                                                                                                                                                                                                                             
    ├── grub.efi                                                                                                                                                                                                                                             
    ├── grubx64.efi                                                                                                                                                                                                                                          
    ├── modelist.efi                                                                                                                                                                                                                                         
    ├── printenv.efi                                                                                                                                                                                                                                         
    ├── route80h.efi                                                                                                                                                                                                                                         
    ├── setjmp.efi                                                                                                                                                                                                                                           
    ├── t2.efi                                                                                                                                                                                                                                               
    ├── t3.efi                                                                                                                                                                                                                                               
    ├── t4.efi                                                                                                                                                                                                                                               
    ├── t5.efi                                                                                                                                                                                                                                               
    ├── t6.efi                                                                                                                                                                                                                                               
    ├── t7.efi                                                                                                                                                                                                                                               
    ├── t8.efi                                                                                                                                                                                                                                               
    ├── tcc.efi                                                                                                                                                                                                                                              
    └── t.efi    
[...]

PR #840, opened 2017-10-23
fix boot in EFI mode on virtualbox

tpgxyz:
This fixes bug #839

@teo  Please review

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -262,10 +262,10 @@ def install_grub(efi_directory, fw_type):
         # Workaround for some UEFI firmwares
         efi_file_source = {"32": os.path.join(install_efi_directory_firmware,
                                               efi_bootloader_id,
-                                              "grubia32.efi"),
+                                              "bootia32.efi"),
                            "64": os.path.join(install_efi_directory_firmware,
                                               efi_bootloader_id,
-                                              "grubx64.efi")}
+                                              "bootx64.efi")}
         shutil.copy2(efi_file_source[efi_bitness], install_efi_boot_directory)
     else:
         print("Bootloader: grub (bios)")


philmmanjaro:
Can't be merged as it won't work. Proper fix would be https://github.com/manjaro/calamares/commit/3275a33827052faabc62aa3368028c2824dafc38

adriaandegroot:
Hi @tpgxyz , thanks for adding this. There's a couple of things going on at the same time in that code:
 - apparently both `grub32.efi` and `boot32.efi` occur; you're changing the original source name (`grub32.efi`) to boot, and then copying that;
 - Phil is giving an explicit target filename, renaming `grub32.efi` to `boot32.efi` during copying.
Your approach seems to assume the original is mis-named. I can't tell from here what is the right approach.

Both patches, though, leave opportunities on the table for cleaning up the Python code and reducing code repetition. I mean, there's a bunch of constants that depend on bitness, which are being created in a convoluted way -- but that could be a separate PR, really.

kkofler:
@philmmanjaro's approach is clearly the right one. The originals are named `grub*`, or the code that is there now would not work at all. The targets should be named `boot*`.

philmmanjaro:
@adriaandegroot the issue was introduced with https://github.com/calamares/calamares/commit/50cefe2ca80c3d6988e4dc5c0df1951a0244db69. We have to ask @agaida why it was changed. Bitness was introduced with https://github.com/calamares/calamares/commit/7e8129a902f76817d6d7d635b1c4b2985f275a9c and fixed with https://github.com/calamares/calamares/commit/dd7cd42118ef02a1fca6cbfe0435b43dcf76d40a and https://github.com/calamares/calamares/commit/4f11faea521437f4fe71109a9b81ae2561d3927a. Sure we can discuss this, however this PR makes no sense and therefore I close this one.

itchka:
Please reopen as the patch does not seem to work
efi_bootfile_failure

abucodonosor:
@itchka I cannot reproduce that one on openmandriva with that patch ofc.

( also in Live Mode modus on the ISO )

philmmanjaro:
@itchka: if you applied the patch by @tpgxyz you're clearly right it doesn't work. Please revert those changes and apply the proper one: https://github.com/calamares/calamares/commit/10ede796f80a1655dde9de993e25e14de3defc27

itchka:
philmmanharo
Unfortunately your patch doesn't work either because when you apply it calamares returns an error saying it cannot find the file grubx64.efi. See the screenshot on the bug report. It is a copy rename operation that is required. I am trying to implement with os.rename but I have never written python in my life so much of what I am writing is guesswork.

philmmanjaro:
@itchka: this code was written for **Manjaro Linux** in which it works. I don't know how OpenMandriva is installing **grub2** for **GPT**. So you claim to have the issue not finding **bootx64.efi**, which doesn't exist on my end at all nor the **grubx64.efi** which needs to be installed as **bootx64.efi** to get it working, which my code actually does.

philmmanjaro:
Original code was:

```
check_target_env_call(["cp",
                               efi_file_source[efi_bitness],
                               efi_file_target[efi_bitness]])
```
which is now:

```
shutil.copy2(efi_file_source[efi_bitness], efi_file_target[efi_bitness])
```

kkofler:
@itchka: The screenshot (the only one you posted) is **not** from @philmmanjaro's patch (as @adriaandegroot pointed out in #839). You tested the wrong patch.

philmmanjaro:
@itchka: please double check if you have the following code:

```
        # Workaround for some UEFI firmwares
        efi_file_source = {"32": os.path.join(install_efi_directory_firmware,
                                              efi_bootloader_id,
                                              "grubia32.efi"),
                           "64": os.path.join(install_efi_directory_firmware,
                                              efi_bootloader_id,
                                              "grubx64.efi")}

        efi_file_target = {"32": os.path.join(install_efi_boot_directory,
                                              "bootia32.efi"),
                           "64": os.path.join(install_efi_boot_directory,
                                              "bootx64.efi")}

        shutil.copy2(efi_file_source[efi_bitness], efi_file_target[efi_bitness])
```

Revert any changes you had made by applying **PR840**.

azubieta:
https://github.com/calamares/calamares/blob/ff3a40368dffcb7ddc71bc20f4e81a6d71916de7/src/modules/bootloader/main.py#L272

>         efi_file_source = os.path.join(install_efi_directory_firmware,
>                                        efi_bootloader_id,
>                                        efi_grub_file),
>         efi_file_target = os.path.join(install_efi_boot_directory,
> efi_boot_file),

the coma "," at the end of each line is an error and cases "shutil.copy2" fails.

philmmanjaro:
@azubieta: seems that slipped thru. That regression made it not yet to our stable branch. Should be fixed now.

PR #845, opened 2017-10-27
Fix reported partition size.

stikonas:


philmmanjaro:
@stikonas: already fixed in **master** or why does it only be needed in our **stable** branch?

kkofler:
The idea is that it will be upmerged to `master` from the stable branch in the regular upmerges. This (commit to the lowest branch and upmerge the branches) is the common KDE git workflow. (And yes, this differs from how distros usually work.)

adriaandegroot:
What @kkofler said: if it's a bug fix which applies to a stable branch, I would rather have it fixed in stable and merged to master than have it fixed in two separate commits.

philmmanjaro:
@adriaandegroot: ya, makes totally sense.

PR #846, opened 2017-10-27
Use KPMcore Resize Operation.

stikonas:
This commit replaces custom Calamares code that was forked from KPM long time ago with calling the same code directly in kpmcore.


philmmanjaro:
@stikonas Currently we use **KPMcore 3.0.3** as minimum. Then we have a **v3.2** branch and some new features get introduced with **v3.3** series. So does this code also work with the current minimum or is a bump needed?

stikonas:
@philmmanjaro This code even works with kpmcore 2.0. (which is unsupported due to other reasons). No need to bump at all.

philmmanjaro:
@stikonas: excellent. Your code simplifies a lot. I see no issue preventing it to be merged then.

stikonas:
(just renamed a variable in the last commit, nothing major)

PR #849, opened 2017-10-27
Added an explanation of the module and what it does when used, or whe…

### LINKED ISSUE #832: Document plymouthcfg.conf
The plymouthcfg module has a config-file that is all comments. This makes it fail the yaml-config test (which in itself isn't that bad). But it could use some more documentation on what it does, and the example file should contain actual settings that (might) work. So it should explain what the setting means and give a valid non-empty value.

For bonus points, look into the meaning of an "empty" setting (e.g. "") and what that does with plymouth, and if the module can handle that properly.

rajudev:
Added better explanation to the module description. 
Hopefully fixes #832


rajudev:
Let me know if I should do better re-structuring of the explanation, or If I am doing something wrong with the explanation

kkofler:
The documentation that you added is actually not correct: Since plymouth detection is now (since the commit ea1c8a0e5ddfc6b72311cf744e2ac514c6630009 that added the module) done only in `plymouthcfg`, you **must** enable the `plymouthcfg` module if you want `grubcfg` to actually enable plymouth support. Just keep the `plymouth_theme` setting commented out if you don't want it to actually change any settings (which is what the existing comment correctly advises).

rajudev:
Well I get that the documentation is missing the correct explanation because of the recent commit. I would change the documentation description accordingly. Well the existing explanation there was okay but It needed a more detailed explanation as per the issue #832 . 
And thanks for the explanation about the commit. That made my understanding of the code base better.

src/modules/plymouthcfg/plymouthcfg.conf
```diff
@@ -1,4 +1,38 @@
 ---
-# The plymouth theme to be set if plymouth binary is present
-# leave commented if packaged default theme should be used
+# Plymouth Configuration Module
+
```
adriaandegroot: In a long comment / documentation block, please also give lines between paragraphs of the documentation a leading `#`; use a blank line to separate sections of documentation related to different configuration options. Here, everything discusses either the module as a whole, or the single configuration option, so every line should start with `#` (it's one section).

src/modules/plymouthcfg/plymouthcfg.conf
```diff
@@ -1,4 +1,38 @@
...
+# This module can be used to setup the default plymouth theme to 
+# be used with your distribution
+
+# Only set your theme with this module if you are not doing it 
+# already in your operating systems build configurations. 
+
+# You should only use this module if the plymouth package is installed 
+# on the build configurations of your distribution & the plymouth 
+# theme you want to configure is installed as well. 
+
```
adriaandegroot: Per @kkofler's comments, this should probably condense to:
```
# Enable this module to enable Plymouth in grub. You must have plymouth installed
# in the target (installed) system.
#
# Leave the configuration option *plymouth_theme* commented out to
# keep whatever default theme is configured.
```

src/modules/plymouthcfg/plymouthcfg.conf
```diff
@@ -1,4 +1,38 @@
...
+
+# Leave this commented if you want to use the default theme 
+# shipped with your distribution configurations. 
+
 # plymouth_theme: spinfinity
+
```
adriaandegroot: This chunk is redundant wrt. the text below, I'd drop it.

src/modules/plymouthcfg/plymouthcfg.conf
```diff
@@ -1,4 +1,38 @@
...
+# plymouth_theme: softwaves
+
+# For other distributions you can look under the themes folder for plymouth, 
+# which is usually /usr/share/plymouth/themes and supply any of theme names 
+# from there to plymouth_theme 
+
+# If the plymouth_theme value is left empty or with a theme name which does not exists. 
+# The module would still look for the theme name for availability in plymouth themes folder. 
+# and as theme will not be found, it will leave the plymouth settings unchanged.  
+
```
adriaandegroot: This last paragraph is incorrect: Calamares doesn't check anything: it just edits `plymouthd.conf` and substitutes in the configured theme name. The way to avoid configuring a specific theme (keeping the default) is to leave the setting uncommented. If the theme configured here doesn't exist .. then it's up to plymouth to deal with that.

PR #850, opened 2017-10-28
NetInstallPage.cpp: fix displaying for Name and Description

abucodonosor:
- fixes 1d7ad9e0450d0d069b1496de0c412baab7f7fa96 which sets
  both header's labels to 0 so 'Name' was always overridden
  by Description and Description never displayed


adriaandegroot:
Sets column 0's header twice, instead of setting column 0 and 1 headers.

PR #851, opened 2017-10-30
Fixes

abucodonosor:
Fix some cDebug() output.

Added a warning when using fallaback groups since this 
may not be what the Distributor want.

Also do not force /etc/locale.gen

### DIFF
--- a/src/modules/locale/LocaleViewStep.cpp
+++ b/src/modules/locale/LocaleViewStep.cpp
@@ -281,10 +281,6 @@ LocaleViewStep::setConfigurationMap( const QVariantMap& configurationMap )
     {
         m_localeGenPath = configurationMap.value( "localeGenPath" ).toString();
     }
-    else
-    {
-        m_localeGenPath = QStringLiteral( "/etc/locale.gen" );
-    }
 
     // Optional
     if ( configurationMap.contains( "geoipUrl" ) &&
--- a/src/modules/partition/jobs/CreatePartitionTableJob.cpp
+++ b/src/modules/partition/jobs/CreatePartitionTableJob.cpp
@@ -87,22 +87,34 @@ CreatePartitionTableJob::exec()
     }
 
     QScopedPointer< PartitionTable > table( createTable() );
-    cDebug() << "Creating new partition table of type" << table->typeName()
-             << ", uncommitted yet:\n" << table;
+    cDebug() << "Creating new partition table of type:" << table->typeName()
+             << " - Uncommitted yet:" << table;
 
     QProcess lsblk;
     lsblk.setProgram( "lsblk" );
     lsblk.setProcessChannelMode( QProcess::MergedChannels );
     lsblk.start();
     lsblk.waitForFinished();
-    cDebug() << "lsblk:\n" << lsblk.readAllStandardOutput();
+
+    QByteArray byte = lsblk.readAllStandardOutput();
+    QStringList lines = QString(byte).split(("\n"),QString::SkipEmptyParts);
+    cDebug() << "CreatePartitionTableJob asked for lsblk output:";
+
+    foreach (auto line, lines)
+        cDebug() << " " << line;
 
     QProcess mount;
     mount.setProgram( "mount" );
     mount.setProcessChannelMode( QProcess::MergedChannels );
     mount.start();
     mount.waitForFinished();
-    cDebug() << "mount:\n" << mount.readAllStandardOutput();
+
+    QByteArray mbyte = mount.readAllStandardOutput();
+    QStringList mlines = QString(mbyte).split(("\n"),QString::SkipEmptyParts);
+    cDebug() << "CreatePartitionTableJob asked for mount output:";
+
+    foreach (auto mline, mlines)
+        cDebug() << " " << mline;
 
     bool ok = backendDevice->createPartitionTable( report, *table );
     if ( !ok )
--- a/src/modules/users/UsersViewStep.cpp
+++ b/src/modules/users/UsersViewStep.cpp
@@ -131,6 +131,8 @@ UsersViewStep::setConfigurationMap( const QVariantMap& configurationMap )
     }
     else
     {
+        // Warn here these groups may really not match what the distro want
+        cDebug() << "WARNING: Using fallback groups. Please check defaultGroups in users.conf";
         m_defaultGroups = QStringList{ "lp", "video", "network", "storage", "wheel", "audio" };
     }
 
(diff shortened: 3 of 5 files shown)


abucodonosor:
Right and therefore no need to add an fallback when someone 
disables localeGenPath  in the configuaration file.

kkofler:
With your change, the code will try to open a file called the empty string (a default-constructed `QString()`), which is not even a valid file name. This is undefined behavior.

I don't see anything wrong with the code that you are trying to "fix". It already does the right thing.

adriaandegroot:
Please don't mix different fixes into a single branch / PR, unless it's all thematically the same (e.g. doing more debug-logging in many modules). It makes it a lot harder to handle the case of "this independent bit is ok, but this other bit has issues".

abucodonosor:
There is nothing wrong but if you go an force stuff while the user disables 
in the configuration file then there is NO point having such a configuration 
option at all right ?

And yes you are right ,  QFSFileEngine::open doesn't like empty strings.
I'm missing to check for exists() before open()..

src/modules/partition/jobs/CreatePartitionTableJob.cpp
```diff
@@ -87,22 +87,34 @@ CreatePartitionTableJob::exec()
     }
 
     QScopedPointer< PartitionTable > table( createTable() );
-    cDebug() << "Creating new partition table of type" << table->typeName()
-             << ", uncommitted yet:\n" << table;
+    cDebug() << "Creating new partition table of type:" << table->typeName()
+             << " - Uncommitted yet:" << table;
```
adriaandegroot: Since @stikonas is doing all kinds of things with dropping Calamares' partition code and using KPMCore instead, I'd hesitate to be changing debugging in that module **right now**. In general, though, this is a useful little bit of polish.

src/modules/partition/jobs/CreatePartitionTableJob.cpp
```diff
@@ -87,22 +87,34 @@ CreatePartitionTableJob::exec()
...
     lsblk.start();
     lsblk.waitForFinished();
-    cDebug() << "lsblk:\n" << lsblk.readAllStandardOutput();
+
+    QByteArray byte = lsblk.readAllStandardOutput();
+    QStringList lines = QString(byte).split(("\n"),QString::SkipEmptyParts);
+    cDebug() << "CreatePartitionTableJob asked for lsblk output:";
+
+    foreach (auto line, lines)
+        cDebug() << " " << line;
```
adriaandegroot: We're running *lsblk* and *mount* purely for debugging-output purposes here. That seems rather expensive, really. And there's the code duplication between this block, and the processing for *mount*, below. This is only two repetitions .. if there's a third, I'd say it's time to add a nice (template) class into the Logger namespace for neatly dumping stdout from a finished QProcess.

src/modules/users/UsersViewStep.cpp
```diff
@@ -131,6 +131,8 @@ UsersViewStep::setConfigurationMap( const QVariantMap& configurationMap )
     }
     else
     {
+        // Warn here these groups may really not match what the distro want
+        cDebug() << "WARNING: Using fallback groups. Please check defaultGroups in users.conf";
```
adriaandegroot: The comment basically doubles up on the debug line, though.

src/modules/partition/jobs/CreatePartitionTableJob.cpp
stikonas: Well, right now I don't plan to touch CreatePartitionTableJob. Partially because it is not necessary for LVM (it doesn't have partition table, it is enough to just disable creation of partition table).

On the other hand maybe it will still be a good idea to move to kpmcore code in the future for CreatePartitionTable too... Less code dublication.

src/modules/locale/LocaleViewStep.cpp
```diff
@@ -281,10 +281,6 @@ LocaleViewStep::setConfigurationMap( const QVariantMap& configurationMap )
     {
         m_localeGenPath = configurationMap.value( "localeGenPath" ).toString();
     }
-    else
-    {
-        m_localeGenPath = QStringLiteral( "/etc/locale.gen" );
-    }
```
adriaandegroot: This only makes sense in the context of changing the documentation that the setting **must** be set; right now the documentation is weirdly formatted, and since the setting is commented out, that suggests that leaving it unset is a sensible thing to do -- that Calamares falls back to something reasonable.

Really the right thing to do is improve the documentation in locale.conf explaining what the possibilityies are (e.g. that SUPPORTED is preferred, and then *localeGenPath* is used, and if that doesn't open, *locale -a* output.)

abucodonosor: Yes I have an configuartion fix and the fix kkofler poited out for /locale/ modules.

Will push out as soon I'm home

kkofler:
On distros that have `/etc/locale.gen`, the file almost certainly has that name, so it makes sense to default to that.

On distros that don't have `/etc/locale.gen`, it does not matter what the default is because the file does not exist. And IMHO it makes more sense to try to open a valid file name and realize it does not exist than to try to open the empty string, which never makes sense.

If it ain't broke, don't fix it!

src/modules/locale/LocaleViewStep.cpp
kkofler: Even with these added, I still don't see the point of changing this at all.

I am the original author of the `locale -a` reading code, it works, I don't see why you absolutely want to change it.

Please do not change working code that you don't understand.

abucodonosor: Why do you think I don't understand how it works ? 

Even you are  author of that code please can you stop 
assuming everyone is stupid but you ?

Also I'm not saying is broken or is wrong! All I say is:

DO not offer configuartion options when they do NOTHING 
and do NOT respect the users settings..

However I drop all these and keep the patches for me ..

kkofler: Sorry, I was somewhat in a bad mood when I wrote that (also because of the recent pull request #849 **from another submitter** (i.e., **not** your fault!) changing documentation without understanding it). I apologize for this unfriendly sentence.

Let me try to get to the technical point now:

> DO not offer configuartion options when they do NOTHING

The option does something: It tells Calamares what file to try to open. If it does not exist, the fallback code is used.

> and do NOT respect the users settings..

The setting is respected: The code tries to open the file, and will use it if it exists. Leaving the option unset is not a documented setting, and it does the (IMHO) most reasonable thing to do, it tries to open the file name as used on the distros actually shipping such a file.

PR #852, opened 2017-10-30
locale.conf: explain what all the settings are about

abucodonosor:
( this time without touching kofler code .. that doesn't seems to be allowed )

### DIFF
--- a/src/modules/locale/locale.conf
+++ b/src/modules/locale/locale.conf
@@ -13,8 +13,7 @@ zone:                       "New_York"
 # System locales are detected in the following order:
 #
 # /usr/share/i18n/SUPPORTED
-#  localeGenPath ( when set )
-#   /etc/locale.gen
+#  localeGenPath (defaults to /etc/locale.gen if not set)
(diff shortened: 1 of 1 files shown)


src/modules/locale/locale.conf
```diff
@@ -1,7 +1,28 @@
...
 zone:                       "New_York"
 
-# GeoIP settings. Leave commented out to disable GeoIP.
-#localeGenPath:             "/etc/locale.gen"
+
+# System locales are detected in the following order:
+#
+# /usr/share/i18n/SUPPORTED
+#  localeGenPath ( when set )
+#   /etc/locale.gen
```
kkofler: This is not correct. The hardcoded `/etc/locale.gen` is not checked if `localeGenPath` is set.

The correct list is:
```
/usr/share/i18n/SUPPORTED
localeGenPath (defaults to /etc/locale.gen if not set)
'locale -a' output
```

abucodonosor: I what I meant .. Fixed now

PR #854, opened 2017-10-30
Document users.conf

abucodonosor:
Document users.conf 
Also added a warning when we use fallaback groups

### DIFF
--- a/src/modules/users/users.conf
+++ b/src/modules/users/users.conf
@@ -22,13 +22,26 @@ defaultGroups:
     - wheel
     - audio
 
+# Some Distributions require a 'autologin' group for the user.
+# Autologin causes a user to become automatically logged in to
+# the desktop environment on boot.
+# Disable when your Distribution does not require such a group.
 autologinGroup:  autologin
+# You can control the initial state for the 'autologin checkbox' in UsersViewStep here.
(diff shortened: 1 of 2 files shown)


src/modules/users/users.conf
```diff
@@ -22,13 +22,26 @@ defaultGroups:
...
+# Disable when your Distribution does not require such a group.
 autologinGroup:  autologin
+# You can control the initial state for the 'autologin checkbox' in UsersViewStep here.
+# Possible values are: true to enable or false to disable the checkbox by default
 doAutologin:     true
 
-# remove the following line to avoid creating /etc/sudoers.d/10-installer
+# When enabled calamares creates a sudoers file for the user.
+#  /etc/sudoers.d/10-installer
+# Remember to add sudoersGroup to defaultGroups.
```
kkofler: I would add something like:

> Remove this setting (delete or comment out the line below) if your distribution already sets up a group of sudoers in its packaging. Otherwise, the setting will be duplicated in the `/etc/sudoers.d/10-installer` file, potentially confusing users.

adriaandegroot: The word "enabled" suggests a boolean value. The first sentence should probably be "When set to a non-empty string, Calamares..."

PR #855, opened 2017-11-01
add qt crash handler and kate temp files to gitignore

bill-auger:


kkofler:
The crash reporter is no longer used, you should just delete that directory.

bill-auger:
ok im still back a few minor versions - ill remove that

PR #861, opened 2017-11-06
[netinstall]: some more debug

abucodonosor:
Be more verbose about what fails and print url , error and so on


abucodonosor:
duh ignore this .. wrong PR button :)

PR #869, opened 2017-12-02
new module : openrcdmcryptcfg - auto unlock encrypted partitions with OpenRC

V3n3RiX:
Unlike systemd, OpenRC doesn't have a cryptsetup-generator which translates crypttab into native units, so one has to configure /etc/conf.d/dmcrypt in order to auto-unlock LUKS partitions (other than /root and swap which are handled by dracut, initcpio, or whatever initramfs generator) during system boot.

This module will allow to have any number of encrypted partitions on distributions using OpenRC. It will gatter the information needed from **globalstorage** and write it into **/etc/conf.d/dmcrypt**, so when OpenRC fires up dmcrypt service, it will unlock the partitions using the crypto_keyfile.bin .

This is an example of what it will write in /etc/conf.d/dmcrypt :

target=luks-152eced5-3a76-4645-8888-3eaab8193185
source=/dev/disk/by-uuid/152eced5-3a76-4645-8888-3eaab8193185
key=/crypto_keyfile.bin

And an output of OpenRC auto-unlocking the partition on a fresh install using this module patched in :

    Setting up dm-crypt mappings ...
    luks-152eced5-3a76-4645-8888-3eaab8193185 using: open /dev/disk/by-uuid/152eced5-3a76-4645-8888-3eaab8193185 luks-152eced5-3a76-4645-8888-3eaab8193185 ...

The module has been tested using multiple encrypted partitions on Redcore Linux, a Gentoo based distribution which uses OpenRC as init system, and it should work on any distribution that uses OpenRC. Any tests or feedback from Manjaro/Artix guys will be highly appreciated.

This is my second attempt to merge this. While the previous code only supported unlocking $HOME partition, the current improved code will work with any partition user chooses to encrypt. I did run autopep8 on the code, so CI shouldn't complain about it.

### DIFF
--- a/src/modules/openrcdmcryptcfg/main.py
+++ b/src/modules/openrcdmcryptcfg/main.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# === This file is part of Calamares - <http://github.com/calamares> ===
+#
+#   Copyright 2017, Ghiunhan Mamut <venerix@redcorelinux.org>
+#
+#   Calamares is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   Calamares is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
+
+import libcalamares
+import os.path
+
+def write_dmcrypt_conf(partitions, root_mount_point, dmcrypt_conf_path):
+    crypto_target = ""
+    crypto_source = ""
+
+    for partition in partitions:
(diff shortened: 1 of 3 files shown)


abucodonosor:
@adriaandegroot @kkofler 

Or we can plug all these into DracutLuksCfgJob and fstab modules with some 
configuration option?

Right now DracutLuksCfgJob takes care of /boot && / && swap .. while
systemd takes care of the rest. 

Now we have the sitution 'other init systems' with dracut need do something
with the rest of the devices and we end up of using dracut + custom modules ,
where in the custom modules we need skip the parts are handled by dracutcfg's ..

V3n3RiX:
Although Redcore does use dracutlukscfg to handle boot&swap&/, this code should work with any other initramfs generator supported as long as the initramfs generator ifselt handles boot&swap&/. So in Manjaro-Openrc now morphed into Artix it would be initcpio+initcpioswaphook+this...

kkofler:
I agree with @V3n3RiX, I also don't think `dracutlukscfg` is the right place to put OpenRC stuff, even as an option, because distros who haven't even picked up `systemd` are even less likely to have picked up `dracut` (with Redcore being the notable exception, apparently).

The OpenRC stuff needs to be in a separate module.

kkofler:
I think the pull request as submitted (new `openrcdmcryptcfg` job module) is fine.

It is sad that OpenRC doesn't even support `/etc/crypttab`, which is a standard much older than `systemd`, but we have to deal with what's there.

PR #875, opened 2017-12-15
Convert FormatPartitionJob and CreatePartitionTableJob to KPMcore Operations.

stikonas:
Unlike Calamares job, KPMcore job correctly handles the case when partition is actually LVM logical volume.

### DIFF
--- a/src/modules/partition/jobs/CreatePartitionTableJob.cpp
+++ b/src/modules/partition/jobs/CreatePartitionTableJob.cpp
@@ -23,15 +23,12 @@
 #include "utils/Logger.h"
 
 // KPMcore
-#include <kpmcore/backend/corebackend.h>
-#include <kpmcore/backend/corebackendmanager.h>
-#include <kpmcore/backend/corebackenddevice.h>
-#include <kpmcore/backend/corebackendpartitiontable.h>
-#include <kpmcore/core/device.h>
-#include <kpmcore/core/partition.h>
-#include <kpmcore/core/partitiontable.h>
-#include <kpmcore/fs/filesystem.h>
-#include <kpmcore/util/report.h>
+#include <core/device.h>
+#include <core/partition.h>
+#include <core/partitiontable.h>
+#include <fs/filesystem.h>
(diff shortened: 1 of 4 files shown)


stikonas:
Now this branch works fine with kpmcore[sfdisk] branch.

(There is still crash introduced by 55ed2ba4c078f2ab4700d6efdfc9fdc84ba5a13b)

philmmanjaro:
@stikonas does 7ce52ec fix also the issue you have here?

stikonas:
Yeah, the issue here was fixed.

philmmanjaro:
@stikonas: I assume with current **kpmcore v3.3** this code doesn't work yet. so the patches in **sfdisk** branch are needed. What is the plan for them to be included in the regular **kpmcore** package?

stikonas:
@philmmanjaro This code works with any kpmcore version, even 3.0. Only CreatePartition operation requires kpmcore v3.3. So patches in sfdisk branch are not needed. It's in fact vice versa. These patches are needed to make Calamares compile with sfdisk branch.

philmmanjaro:
Ok, thx for clearing that up @stikonas. I already thought that ...

stikonas:
Well, the idea is that I wanted to make API changes as painless as possible. I.e. some Calamares versions should work with both old and new kpmcore. It will take a while before sfdisk branch is released as kpmcore 3.4. By that time at least Calamares 2.2.0 will be released.

philmmanjaro:
@stikonas: you mean **CAL v3.2.0**, right? Anyway, some of my users reported some issue with [not being able to flag](https://forum.manjaro.org/t/manjaro-v17-1-rc2-feedback-thread/36937/7) partitions. Do you see any relation to **kpmcore-3.2** having a regression with it?

philmmanjaro:
@stikonas: somehow this pull creates a regression with **erase disk** option. It seems the new partitions never get mounted and therefore the unsquash will happen to the **tmpfs** mounted as `/tmp`. For more details see [here](https://github.com/calamares/calamares/issues/882).

PR #883, opened 2017-12-26
[unpackfs] skip error check when rm of tmp folder fails

agaida:
but it does without help

Unhandled exeptions are bad in calamares - and backtraces not exactly user-friendly.

### DIFF
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -218,7 +218,7 @@ class UnpackOperation:
 
             return None
         finally:
-            shutil.rmtree(source_mount_path)
+            shutil.rmtree(source_mount_path, ignore_errors=True, onerror=None)
 
     def mount_image(self, entry, imgmountdir):
         """


agaida:
To be true about, right now i don't know what actually fails, will investigate tomorrow - WIP

agaida:
after locking in unpackfs a second time i see no reason to let calamares fail if the rmtree goes wrong - it would be a temp dir what is not cleaned up fully - which reasons it might have. So not nice, but no reason to let calamares fail the whole installation.

philmmanjaro:
@agaida: what is the actual error you may have on your end. Do you have some examples?

agaida:
@philmmanjaro - get a short backtrace, first line was the return, second line the changed one. Thats happend with all of my isos, approx one out of five installation fails. And i would like calamares not to fail because a temp dir is not cleaned up fully. As i wrote, it is not nice that the cleanup fails, but imho no reason to hard break an installation. Will try an old iso and create a trace.

philmmanjaro:
`shutil.rmtree(source_mount_path)` was introduced with: https://github.com/calamares/calamares/commit/e937782a4e88a86541d65ebd34ae1ee10dbfbb73 Original added to use a temp file via https://github.com/calamares/calamares/commit/cb2dcf78609b137c3d4026e220a6727dcc409894.

Adding your hack may "fix" it for you, however is not a proper solution. I can live with it, to skip an error, however never noticed an issue with **Manjaro**.

What do others say about this "hack"?

agaida:
````
[2017-12-27 19:21:26] <agaida> crazy: could you be so kind and have a look at: https://github.com/calamares/calamares/pull/883
[2017-12-27 19:27:49] <crazy> agaida: looks fine to me
[2017-12-27 19:28:21] <agaida> i dunno why, but 1 out of 5 installations fail at this point
[2017-12-27 19:28:34] <agaida> maybe timing or so
[2017-12-27 19:29:09] <agaida> ok, was mostly in kvm, but ...
[2017-12-27 19:30:57] <crazy> agaida: never got that to fail but is for sure possible 
````

adriaandegroot:
The code does this:

1.  Calamares makes a temporary dir with `mkdtemp()`
2.  Calamares makes one or more image-mount directories in there
3.  Unpacks from the now-mounted image into the target system
4.  Lazy unmounts (`umount -l`) the mounted image
5.  Removes the temporary dir

I expect somewhere between 4 and 5 something is busy in the mounted image, which then blocks the complete removal of the mounted image, and then the rmtree ends failing. Like Phil said, it'd be nice to have an actual and complete error message from rmtree, but I agree with Alf that failing here is nonsensical.

kkofler:
Uh, `rmtree` of the mount is outright dangerous though! You do **not** want to `rmtree` the mounted images, if unmounting failed for whatever reason.

PR #910, opened 2018-03-02
Enable the correct language in the locale.gen list 

AlmAck:
The original code does not distinguish the document comments inside the locale.gen file from the real locale list. The language was then enabled from the header comments of the file instead of the correct value in the list.

The new code verify tha the complete locale string is just after the first character of the string, enablig only the correct value of the locale list.
An example:
```
#  en_US.UTF-8 UTF-8 --> document header, should not be enabled
#en_US.UTF-8 UTF-8 --> correct section to enable
```

Related to this request:
https://code.chakralinux.org/tools/calamares-chakra/issues/2


AlmAck:
Don't know why but the merge request imported some comments from demm, they can be removed? or I have to create a new one?

abucodonosor:
Not only that but merge comits with conflicts etc..

Try an clean checkout , or add a branch or something

AlmAck:
Created a new one, more clean :-)

https://github.com/calamares/calamares/pull/911

PR #911, opened 2018-03-03
Enable the correct language in the locale.gen list

AlmAck:
The original code does not distinguish the document comments inside the locale.gen file from the real locale list. The language was then enabled from the header comments of the file instead of the correct value in the list.

The new code verify that the complete locale string is just after the first character of the string, enabling only the correct value of the locale list.
An example:
```
#  en_US.UTF-8 UTF-8 --> document header, should not be enabled
#en_US.UTF-8 UTF-8 --> correct section to enable
```

Related to this request:
https://code.chakralinux.org/tools/calamares-chakra/issues/2

### DIFF
--- a/src/modules/localecfg/main.py
+++ b/src/modules/localecfg/main.py
@@ -65,12 +65,15 @@ def run():
         with open("{!s}/etc/locale.gen".format(install_path), "w") as gen:
             for line in text:
                 # always enable en_US
-                if en_us_locale in line and line[0] == "#":
+                if line.startswith(en_us_locale, 1):
                     # uncomment line
                     line = line[1:].lstrip()
 
                 for locale_value in locale_values:
-                    if locale_value in line and line[0] == "#":
+                    # check the locale value starting from
+                    # the second index because we expect that
+                    # the first one is a '#'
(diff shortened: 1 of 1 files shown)


kkofler:
I think this is pretty much a hack. It distinguishes 2 kinds of comments, which are not defined by the file format, but by how your distro uses the file.

I also think that you should have kept the ` and line[0] == "#"` checks, or used `line.startswith("#" + en_us_locale)` as the condition. Right now, your code would match things like `zen_US.UTF-8 UTF-8`, which would be a 3-letter language code that should not be touched. It is only a comment if the character you skip is actually a `#` sign.

AlmAck:
I agree on everything you wrote above. The header text is provided by our distribution (same for Arch). I like a lot the suggested solution to include the "#" as match condition, looks less as a hack and should be useful for other distributions too.

adriaandegroot:
There's plenty else to polish in this python code, unrelated to the original problem: what are those lstrip()s for? What about locales with an extra suffix? Those are enabled now too. Is the convention of this file -- that it lists all the **available** locales, commented out, and you can uncomment those you want -- documented anywhere?

Issue #887, opened 2017-12-29
Calamares should only allow installing bootloader on (KPMcore) DiskDevices


stikonas:
#### Submission type

  - [x] Bug report
  - [x] Feature Request

#### Info regarding which version of Calamares is used, which Distribution

> Manjaro 17.1 rc4 (Calamares 3.1.x with kpmcore-3.2 patches)

#### Provide information on how the disks are set up, in detail, with full logs of commands issued 

> /dev/sda contains a single LVM PV partition. I created a single LVM volume group containing /dev/sda1. (LVM Volume Group and LVM PV were created in advance with KDE Partition Manager)

#### What do you expect to have happen when Calamares installs?

> Only /dev/sda is listed in bootloader location combobox

#### Describe the issue you encountered

> All Devices are listed there, including LVM devices.

adriaandegroot:
Fixed with merge of Caio's branch.

PR #945, opened 2018-04-30
[partition] Ignoring LVM devices in PartUtils::getDevices to prevent installing bootloader in LVM VG.

cjlcarvalho:
Related to: https://github.com/calamares/calamares/issues/887

### DIFF
--- a/src/modules/partition/core/DeviceList.cpp
+++ b/src/modules/partition/core/DeviceList.cpp
@@ -129,7 +129,12 @@ QList< Device* > getDevices( DeviceType which, qint64 minimumSize )
 
     // Remove the device which contains / from the list
     for ( DeviceList::iterator it = devices.begin(); it != devices.end(); )
-        if ( ! ( *it ) ||
+        if ( (*it)->type() != Device::Type::Disk_Device )
+        {
+            cDebug() << " .. Removing device that is not a Disk_Device from list " << it;
+            it = erase(devices, it );
+        }
+        else if ( ! ( *it ) ||
                 ( *it )->deviceNode().startsWith( "/dev/zram" )
         )
         {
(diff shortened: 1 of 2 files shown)


src/modules/partition/core/DeviceList.cpp
```diff
@@ -152,6 +152,11 @@ QList< Device* > getDevices( DeviceType which, qint64 minimumSize )
             cDebug() << "  .. Removing too-small" << it;
             it = erase(devices, it );
         }
+        else if ( (*it)->type() == Device::LVM_Device )
```
stikonas: I think the opposite logic would be better. Allow installing only if it is Device::Type::Disk_Device.
Otherwise you'll have to add similar check for other Device types if we introduce them later (maybe RAID, not sure yet if we'll need it).

(Also I would suggest including enum scope i.e. Type, so that at some point we can switch to C++11 scoped enums)

stikonas:
I think you can just force push a new commit and pull request will update.

On 2 May 2018 18:13:58 BST, "Caio Jordão Carvalho" <notifications@github.com> wrote:
>cjlcarvalho commented on this pull request.
>
>
>
>> @@ -152,6 +152,11 @@ QList< Device* > getDevices( DeviceType which,
>qint64 minimumSize )
>             cDebug() << "  .. Removing too-small" << it;
>             it = erase(devices, it );
>         }
>+        else if ( (*it)->type() == Device::LVM_Device )
>
>Ok!
>
>Should I send another PR and ignore this one?
>
>-- 
>You are receiving this because you commented.
>https://github.com/calamares/calamares/pull/945#discussion_r185571428

-- 

cjlcarvalho:
There were some problems with libpmlibpartedbackendplugin dependency during Travis build. :(

PR #950, opened 2018-05-10
[partition] Only bootloader model should ignore devices that are not of Disk_Device type

cjlcarvalho:
Fixing bug related to https://github.com/calamares/calamares/pull/945, because only bootloader model should ignore devices that aren't of Disk_Device type in PartitionCoreModule::doInit().


philmmanjaro:
looks fine for me.

PR #955, opened 2018-05-17
Update CreateUserJob.cpp

fhdk:
Removal of the hardcoded shell, letting the `/etc/default/useradd` be responsible for the shell.

```
# useradd defaults file for Manjaro Linux
# useradd defaults file for ArchLinux
# original changes by TomK
GROUP=users
HOME=/home
INACTIVE=-1
EXPIRE=
SHELL=/bin/zsh
SKEL=/etc/skel
CREATE_MAIL_SPOOL=no

```

### DIFF
--- a/src/modules/users/CreateUserJob.cpp
+++ b/src/modules/users/CreateUserJob.cpp
@@ -149,8 +149,6 @@ CreateUserJob::exec()
     int ec = CalamaresUtils::System::instance()->
              targetEnvCall( { "useradd",
                               "-m",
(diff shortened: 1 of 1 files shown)


adriaandegroot:
That seems obvious, in retrospect. I would have called the pull request "Remove hardcoded shell from user module", myself -- that's nit-picking though. This is certainly a change that needs to be mentioned in release notes, because distro's might be relying on the existing -s behavior.

PR #958, opened 2018-05-19
avoid pinging hard-coded server for netinstall.yaml

bill-auger:
groupsUrl is optional and should be commented out by default - also this URL is 404 - also only chakra would ever want to ping the chakra server anyways

### DIFF
--- a/src/modules/netinstall/netinstall.conf
+++ b/src/modules/netinstall/netinstall.conf
@@ -1,7 +1,10 @@
 ---
 # This is the URL that is retrieved to get the netinstall groups-and-packages
-# data (which should be in the format described in netinstall.yaml).
-groupsUrl: http://chakraos.org/netinstall.php
+# data (which should be in the format described in netinstall.yaml), e.g.:
+#   groupsUrl: http://example.org/netinstall.php
+# or it can be a locally installed file:
+#   groupsUrl: file:///usr/share/calamares/netinstall.yaml
+# groupsUrl: file:///usr/share/calamares/netinstall.yaml
 
 # If the installation can proceed without netinstall (e.g. the Live CD
 # can create a working installed system, but netinstall is preferred


src/modules/netinstall/netinstall.conf
```diff
@@ -1,7 +1,9 @@
 ---
 # This is the URL that is retrieved to get the netinstall groups-and-packages
 # data (which should be in the format described in netinstall.yaml).
-groupsUrl: http://chakraos.org/netinstall.php
+# groupsUrl: http://example.org/netinstall.php
```
adriaandegroot: Please indent if this is an example in the documentation, e.g.
```
#    groupsUrl: http://example.org/netinstall.php
# The URL can also point to a local file:
#    groupsUrl: file:///usr/local/calamares/netinstall.yaml
# There is no default value. If not set, then the netinstall page
# will show empty groups, which is strange.
```

src/modules/netinstall/netinstall.conf
```diff
@@ -1,7 +1,9 @@
...
+# or it can be a locally installed file (add netinstall.yaml to ./CMakeLists.txt)
```
adriaandegroot: Don't mention changing CMakeLists.txt, that's really not the way downstream should be putting configuration files into the live system.

src/modules/netinstall/netinstall.conf
```diff
@@ -1,7 +1,9 @@
...
+# groupsUrl: /usr/share/calamares/netinstall.yaml
```
adriaandegroot: Leave one, commented-out, example URL, with only octothorpe-space in front of it. This one's OK.

bill-auger:
just FYI the reason i came across this is because i found an alternate use-case for this module - one of our ISOs is the "complete" edition with a full set of packages on the ISO for offline install - i originally thought to write a new module for the user to choose some optional packages such as a "graphics" and "developement" set - then i found that the netinstall module is fully appropriate for that task except only for the requirement that the package list be fetched from the network - this module is already general enough that it could be accurately renamed to the 'PackageSelection' module - the only thing that characterizes it as the 'NetInstall' module is the single fact that the package list is assumed to be fetched from the network - that is perhaps how the module was intended to be used but it seems quite sensible in the context of an offline install to use a local file in this module for selecting optional packages

i agree that the package list should be able to be changed without rebuilding the program - i removed the note about CMakeLists.txt - that is not the way that i am using it either - what i actually ended up doing was to list the package groups directly in the netinstall.conf file - in that way the 'packageGroups' data is automatically available to the view - you can review my implementation in this PR #962 if you think it would be a valuable addition

adriaandegroot:
With your last comment, it might even be a good idea to improve the description of the module (if there even is one) in the deployer's guide, to remind people what modules there are and what they do (because, as you point out, the netinstall name doesn't really scream "pick packages to install").

PR #961, opened 2018-05-21
Branding

raurodse:
This patch enables the thematization of all calamares interface elements with qss technology without having to modify system's qt theme.
Style file is located in /etc/calamares/branding/THEMENAME/stylesheet.qss

To achieve this, I have set names to every calamares qt element, shown in the following attached file.
calamares

### DIFF
--- a/src/calamares/CalamaresWindow.cpp
+++ b/src/calamares/CalamaresWindow.cpp
@@ -35,6 +35,7 @@
 #include <QLabel>
 #include <QTreeView>
 #include <QFile>
+#include <QFileInfo>
 
 CalamaresWindow::CalamaresWindow( QWidget* parent )
     : QWidget( parent )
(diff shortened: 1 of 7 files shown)


adriaandegroot:
That's really nice work, and works fairly well. What it probably still needs is something for the sidebar -- you mentioned that elsewhere -- and an example file showing how to do the actual settings.

I have done some work on your code:
 - moved the stylesheet loading to the branding component; that's simpler
 - applied Calamares coding style (which is mostly more-spaces-around-parentheses)
 - added copyright headers for you

There is a branch calamares/raurodse-branding in the official calamares repo, if you would like to continue work from there.

adriaandegroot:
(for clarity, that means this is halfway-merged)

bill-auger:
i hope that this styling spec accommodates the features described in PR #776 and PR #826, namely for the input image to a square and to be presented un-distorted as a square and for the width of the sidebar itself to be adjustable rather than the hard-coded width (190px IIRC) - this is currently a crazy mixture of configuration keys, hard-coded values, and runtime formulas

as the original sidebar styling issue was closed, i will repeat here the suggestion that perhaps a new entry is added to branding.desc and a new entity in Branding.cpp such as: 'ProductLogoWide' for an explicitly rectangular input image and allow this feature to be a configuration option as suggested to switch between using the exsting 'ProductLogo' (to be rendered as a square) and the 'ProductLogoWide' (to be rendered as a wide rectangle) sources

tsimonq2:
Bump, I'd really like this for my distribution. :)

adriaandegroot:
This kind of languished: @raurodse I was hoping you would finish the corner cases. In any case, the existing code has been merged, some documentation has been added, so in 3.2.6 you'll be able to play with the stylesheets and find all the corner cases that have been missed.

PR #962, opened 2018-05-22
implement option to load package groups from netinstall.conf

bill-auger:
this patch allows the netinstall package groups to be defined in this netinstall.conf file for the purpose of offering a selecton of optional packages during offline install - to keep the delta small,  rather than overriding the YAML parsing method to handle the QT structs, i wrote a new method: 'parseGroupList()' that is analagous to and mutually exclusive with the existing 'loadGroupList()/dataIsHere()' pair; using the YAML library to convert the netinstall.conf QVariantMap to YAML and write that out to the local file specified as 'groupsUrl' - but then i decided it was better to avoid writing to the filesystem (and the additional include) and refactored the populating of the 'groupswidget' and enabling of the "next" button out of 'dataIsHere()' into a new method 'populateGroupsWidget()' to allow the two loader methods to share that functionality - the delta is larger this way but IMHO it makes the 'dataIsHere()' method more concise by consolidating the multiple 'reply->deleteLater()' and 'checkReady()' calls

### DIFF
--- a/src/modules/netinstall/NetInstallPage.cpp
+++ b/src/modules/netinstall/NetInstallPage.cpp
@@ -48,6 +48,10 @@
 
 using CalamaresUtils::yamlToVariant;
 
+
+const char* NetInstallPage::HEADER_TEXT = "<h1>Package Selection</h1>";
+
+
 NetInstallPage::NetInstallPage( QWidget* parent )
     : QWidget( parent )
     , ui( new Ui::Page_NetInst )
@@ -55,6 +59,7 @@ NetInstallPage::NetInstallPage( QWidget* parent )
     , m_groups( nullptr )
 {
     ui->setupUi( this );
+    ui->header->setText( tr( HEADER_TEXT ) );
 }
 
 bool
@@ -84,6 +89,8 @@ NetInstallPage::readGroups( const QByteArray& yamlData )
 void
 NetInstallPage::dataIsHere( QNetworkReply* reply )
 {
+    bool is_valid_package_data = false;
+
     // If m_required is *false* then we still say we're ready
     // even if the reply is corrupt or missing.
     if ( reply->error() != QNetworkReply::NoError )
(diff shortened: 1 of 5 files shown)


kkofler:
Hmmm, the thing is, `netinstall.conf` is actually in YAML, so why read it into a `QVariantMap` and convert it back to `libyaml` structures?

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -48,13 +48,18 @@
 
 using CalamaresUtils::yamlToVariant;
 
+
+const char* NetInstallPage::HEADER_TEXT = "<h1>Package Selection</h1>";
```
adriaandegroot: - Translations don't work that way. You need to keep the literal (English) string inside the tr() call.
 - It would be `static const char HEADER_TEXT[]=` anyway, because there's no need to expose it in the class

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -48,13 +48,18 @@
...
+
+
 NetInstallPage::NetInstallPage( QWidget* parent )
     : QWidget( parent )
     , ui( new Ui::Page_NetInst )
     , m_networkManager( this )
     , m_groups( nullptr )
 {
     ui->setupUi( this );
+    ui->header->setText( tr( HEADER_TEXT ) );
```
adriaandegroot: Why not change that in the UI file already?

abucodonosor:
Hmm .. but you can do offine installs with the yaml file already..

groupsUrl: file:///some_path_to_the.yaml

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -92,27 +99,82 @@ NetInstallPage::dataIsHere( QNetworkReply* reply )
...
+        QVariantMap package_group = package_groups.value(groups_n).toMap();
+        std::string group_name = package_group["name"].toString().toStdString();
+        std::string group_desc = package_group["description"].toString().toStdString();
+        std::string group_critical = package_group["critical"].toString().toStdString();
+        QStringList group_packages = package_group.value( "packages" ).toStringList();
+        std::vector< std::string > packages;
+        foreach ( const QString& package , group_packages )
+            packages.push_back( package.toStdString() );
+
+// TODO: deleteme DEBUG
```
adriaandegroot: Please remember to delete them before doing the pull request

adriaandegroot:
As @kkofler  and @abucodonosor have already said, this seems like a lot of coding effort to avoid using two files (netinstall.conf and a yaml file with the actual groups). That said, you could:
 - add a check in `readGroups()` (which takes a byte array, and handles that as YAML data and hands that off to PackageModel) that if the YAML data **isn't** a list, but **does** have a *groups* key, and *groups* is a list, then use the YAML node of the value of *groups* instead.
By doing that, you could add a *groups* key as you've done here, to the netinstall.conf file, and then set groupsUrl to point to the same netinstall.conf file! Only a few lines of code, and you're done.

Another approach would be to make the PackageModel() work with QVariantMap or QVariantList and do conversions earlier (in `readGroups()`) and optionally call that from `setConfigurationMap()`.

src/modules/netinstall/NetInstallPage.cpp
bill-auger: this is also used as prettyName() in the view

bill-auger:
yes - using the .yaml file would be much simpler than this - indeed that could be the netinstall.conf file itself; and that was the first thing i tried - but the thing is that i have data for other options besides just the packages; so i needed to modify the parsing behavior in some way to tease that out before it goes to the PackageModel - i remember when i tried doing that, having trouble convincing the YAML library to do that for me - and i seem to remember it's documentation being more confusing than helpful - so i scrapped that approach and tried from the opposite direction, writing with the YAML library instead of reading; and that got me a working result quicker - also the semantics of the method name 'loadGroupList()' would be confusing if the YAML file contains anything other than package groups data

as i said, until today i had it teasing out the non-package data in 'setConfigurationMap()' and writing out the YAML data for the package list to the filesystem at runtime to be handled in the normal way - this current implementation retains the correct semantics and is a better design but carries a significantly more intrusive diff - my guiding principle as a downstream is favor minimizing the potential for merge conflicts over elegance; so i will probably revert to the implementation that writes out a .yaml file for the simplicity of maintenance - that would reduce this patch to the body of the 'parseGroupList()' method moved inside 'setConfigurationMap()'

the reason for converting the QVariantMap to YAML is because the .conf data was already present as a QVariantMap in 'setConfigurationMap()'; but the 'readGroups()' method expects a bytearray of YAML; so again, i was just using the plumbing that was already there and isolating my patches to avoid modifying or interleaving with upstream code as much as possible

again, i did not expect this would be a particularly interesting contribution - i only intended this PR to demonstrate my use case - if you do think it may be generally useful, i will update this after i re-work it to be simpler

src/modules/netinstall/NetInstallPage.cpp
bill-auger: it is static and private, defined in the .h - where would you prefer such constants to be? especially if they pretain to only this module

kkofler:
So, can we discard (close without merging) this pull request?

adriaandegroot:
Let's close this one, then, having reached a couple of conclusions:
 - it **is** an interesting idea
 - multiple implementations are possible, depending on what part of the code you want to refactor:
   - make the read-data more flexible and allow a special key
   - convert some methods to using QVariantList to share between reading-from-URL and receiving-from-config

I'm going to call this PR a **success** from the point of view of "the art of the Pull Request" (Paul Adams; see youtube, FOSS-North, elsewhere) as it illustrates you can use a PR to discuss implementation before going all the way to the end.

PR #965, opened 2018-05-26
Servicescfg

udeved:
Serviescfg is the non systemd equivalent to services module.
It supports currently openrc and runit.


abucodonosor:
I don't like that .. ( ignoring the code right now ) ..

Make it per-init don't start another  all-in-one mess.

udeved:
@adriaandegroot 

From code perspective, I would love to see a globalstorage value for the init type set in settings.conf or being detected by an init module or so, instead of configuring each module individually, eg dbus.
The initsys is basically ugly expansion to support runit.
Regarding the setExpression, well, it properly sets keymaps on openrc. runit is a pretty recent addition. Do you dislike the name of the method or do you suggest using fileinput class?

@abucodonosor 

I don't see a reason to double the module, its using same module conf, does basically the same for both inits, settings symlinks.

abucodonosor:
@udeved 

if you think that does not matter then that pull request is wrong anyway to start with.

Then drop all that in services module , no need for yet another  one , right ?

udeved:
@abucodonosor 

I really do not get your issue here.
Why two modules that do essentially same thing, using same conf yaml?
Would be different if runit should be more configurable in terms where services are located.

abucodonosor:
@udeved 

If we say we drop all kind init handling into *one* file then we already have one.
Go and make this one handle all kind inits and Distros .. Which you'll find
out very soon it start being a mess.

Do one thing per-module and do it right.

Also your code is broken anyway .. You should call it openrc-from-gentoo-to-archlinux-clones-sort modules.

https://github.com/calamares/calamares/pull/965/commits/1ece4fe73c0aafd8fe295e37af4b71ddd723ab90#diff-d632b7e48c3825df022a2e7f59e03f1dR49
https://github.com/calamares/calamares/commit/1ece4fe73c0aafd8fe295e37af4b71ddd723ab90#diff-d632b7e48c3825df022a2e7f59e03f1dR62
and so on ..

This kind stuff is broken and Distro specific..

V3n3RiX:
I have no clue about runit, so I'll just say my thoughts about openrc part of this code.

First, I really wonder why did you felt the need to even touch /etc/rc.conf . That file is used to configure OpenRC itself, and has nothing to do with enable/disable services as this module is intended to be. 

You should really limit the module to "rc-update add service default" to enable a service. This is what systemd counterpart does anyway.

I won't even go to the point where you just assume the PATH for some things are exactly the same in all distros. For instance in Gentoo : 

shakuras ~ # which sulogin
/sbin/sulogin

abucodonosor:
@V3n3RiX 

That PATH seems to only exists in ArchlInux .. For sure having sulogin in /bin/ is wrong.

Distro with split user dirs have it in /usr/sbin and linked back to /sbin , while
all others in /sbin

Well everyone seems (ab)using calamares these days to workaround
Distro specific issue like packaging etc..

In fact such sed calls are not needed when your Distribution provides sane packages.

udeved:
@V3n3RiX 

The openrc part works on gentoo as well. On artixlinux, we use the same sysconfdir locations for openrc as on gentoo.
I could certainly remove the sed calls and have the rc.conf stuff set in the package.
The code is basically from a time when still on manjaro, where we I had to work around the issue that packages came from arch linux. On artix, we build our packages. So I can change that.

Regarding runit, this is gonna be hard to make it work on any runit distro, except I make the sv dir locations configurable, There is no standard where distros place the services dir and it defaults to /services.

V3n3RiX:
This already looks a lot better to me. Plain simple enable/disable services using OpenRC. I'll do some tests on Redcore Linux when I have time...

abucodonosor:
Yes that looks much better.

@udeved 

how about you rename it now to services-openrc or openrc-services ?

Also about runit yes , things could have any PATH should be configurable
and then it may work on almost any Dsitribution with runit as init.

@V3n3RiX  @udeved 

should we check return codes and exit when 'some' service
fails to enable / disable ? ( same like systemd services modules is doing )

V3n3RiX:
@abucodonosor 

Yes, return codes should be checked if this module is intended to have same features as systemd one. Aka mark a service as critical and halt if it fails to enable. So altough it looks so much better, is not as complete as the systemd counterpart.

udeved:
https://github.com/calamares/calamares/pull/966

I created new PR, renamed module.
We can close this PR.

udeved:
@V3n3RiX 

regarding runit, I'll make the sv dir locations configurable, but need to test. PR with runitcfg later.

adriaandegroot:
OK, closing this one in favor of the newer PR. Thanks for taking the time to improve the code and work on the design.

udeved:
@adriaandegroot 

If I may, I'd like to ask a question, since documentation is sparse.

I got a module that basically does a pacman net boot strap to install the base system. It replaces the default unpackfs module and uses the netinstall module for package selection.
The problem with this module is that the overall progress of calamares cannot be scaled properly, ie no matter what I tried to scale the progress, it didn't work out to eg reserve 50% of overall progress for the module to have more fine grained progress within the module with package download and install.
Have there been changes on that front?

PR #966, opened 2018-05-27
add openrccfg

udeved:
renamed openrccfg and cleaned up, due to a new runitcfg module to have a naming schema.

### DIFF
--- a/src/modules/openrccfg/main.py
+++ b/src/modules/openrccfg/main.py
@@ -27,22 +27,25 @@ from os.path import exists, join
 
 
 class OpenrcController:
-    """This is the service controller
+    """This is the openrc service controller
     """
 
     def __init__(self):
         self.root = libcalamares.globalstorage.value('rootMountPoint')
         self.services = libcalamares.job.configuration.get('services', [])
+        self.initdDir = libcalamares.job.configuration['initdDir']
+        self.runlevelsDir = libcalamares.job.configuration['runlevelsDir']
 
     def update(self, state):
-        """Update init scripts
+        """call rc-update for each service listed
         """
 
         for svc in self.services[state]:
-            if exists(self.root + "/etc/init.d/" + svc["name"]):
-                target_env_call(
-                    ["rc-update", state, svc["name"], svc["runlevel"]]
-                )
+            if exists(self.root + self.initdDir + "/" + svc["name"]):
+                if exists(self.root + self.runlevelsDir + "/" + svc["runlevel"]):
+                    target_env_call(
+                        ["rc-update", state, svc["name"], svc["runlevel"]]
+                    )
 
     def run(self):
         """Run the controller
--- a/src/modules/openrccfg/main.py
+++ b/src/modules/openrccfg/main.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# === This file is part of Calamares - <http://github.com/calamares> ===
+#
+#   Copyright 2016, Artoo <artoo@manjaro.org>
+#   Copyright 2017, Philip Müller <philm@manjaro.org>
+#   Copyright 2018, Artoo <artoo@artixlinux.org>
+#
+#   Calamares is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   Calamares is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
+
+import libcalamares
+
+from libcalamares.utils import target_env_call
+from os.path import exists, join
+
+
+class OpenrcController:
+    """This is the service controller
+    """
+
+    def __init__(self):
+        self.root = libcalamares.globalstorage.value('rootMountPoint')
+        self.services = libcalamares.job.configuration.get('services', [])
+
+    def update(self, state):
+        """Update init scripts
+        """
+
+        for svc in self.services[state]:
+            if exists(self.root + "/etc/init.d/" + svc["name"]):
+                target_env_call(
+                    ["rc-update", state, svc["name"], svc["runlevel"]]
+                )
+
+    def run(self):
+        """Run the controller
+        """
+
+        for state in self.services.keys():
+            self.update(state)
+
+def run():
+    """Setup services
+    """
+
+    return OpenrcController().run()
--- a/src/modules/openrccfg/module.desc
+++ b/src/modules/openrccfg/module.desc
@@ -1,0 +1,5 @@
+---
+type:       "job"
+name:       "openrccfg"
+interface:  "python"
+script:     "main.py"
--- a/src/modules/openrccfg/openrccfg.conf
+++ b/src/modules/openrccfg/openrccfg.conf
@@ -1,8 +1,39 @@
+# operccfg
+# openrc services module to set service runlevels via rc-update in the chroot
+#
+# format of the conf
+###
+# services:
+#     add:
+#         - name: foo1
+#           runlevel: default
+#         - name: foo2
+#           runlevel: nonetwork
+#     del:
+#         - name: foo3
+#           runlevel: default
+#
+# initdDir: /etc/init.d
+#
+# runlevelsDir: /etc/runlevels
+####
+# add: list of services and their runlevels to add
+# del: list of services and their runlevels to delete
+# name: the service name
+# runlevel: can hold any runlevel present on the target system
+# initdDir: holds the openrc service directory location
+# runlevelsDir: holds the runlevels directory location
+#
+# handle del with care and only use it if absolutely necessary
+# if a service is listed in the conf but is not present/detected on the target system,
+# or a runlevel does not exist, it will be ignored and skipped
+#
 ---
 services:
     add:
         - name: "NetworkManager"
           runlevel: "default"
-#    del:
-#        - name: "hwclock"
-#          runlevel: "boot"
+
+initdDir: /etc/init.d
+
+runlevelsDir: /etc/runlevels
--- a/src/modules/openrccfg/openrccfg.conf
+++ b/src/modules/openrccfg/openrccfg.conf
@@ -1,0 +1,8 @@
+---
+services:
+    add:
+        - name: "NetworkManager"
+          runlevel: "default"
+#    del:
+#        - name: "hwclock"
+#          runlevel: "boot"


src/modules/openrccfg/module.desc
```diff
@@ -0,0 +1,5 @@
+---
```
adriaandegroot: Most modules have some documentation explaining what the module does, above this --- comment line.

src/modules/openrccfg/openrccfg.conf
```diff
@@ -0,0 +1,8 @@
+---
```
adriaandegroot: Most modules have a short bit of documentation about the configuration in general, above this --- comment line.

src/modules/openrccfg/openrccfg.conf
```diff
@@ -0,0 +1,8 @@
...
+services:
```
adriaandegroot: Most modules have documentation on the keys and possible values. Here, something like 

> *services* is a map with one of more object keys (e.g. *add*) that correspond to
> the first argument of rc-update. The following keys are in use: *add*, *del*.
> The value of each of those keys is a list of service maps; each of those service
> maps has a key *name* and a key *runlevel*. The named service is added, or
> deleted, from the given runlevel. To add or remove a service from multiple
> runlevels, list it multiple times with different runlevels.

src/modules/openrccfg/main.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self):
+        self.root = libcalamares.globalstorage.value('rootMountPoint')
+        self.services = libcalamares.job.configuration.get('services', [])
+
+    def update(self, state):
+        """Update init scripts
+        """
+
+        for svc in self.services[state]:
+            if exists(self.root + "/etc/init.d/" + svc["name"]):
```
adriaandegroot: Should something be logged if it doesn't exist? That means there's a mismatch between the configuration, and the installed system.

src/modules/openrccfg/main.py
```diff
@@ -0,0 +1,58 @@
...
+                target_env_call(
+                    ["rc-update", state, svc["name"], svc["runlevel"]]
```
adriaandegroot: Maybe `check_target_env_call`, and catch the possible exception in order to log it (or let it go, then the user gets a real cancel message).

adriaandegroot:
@udeved Looks a lot better and cleaner, too. It does have a lot of review comments, but those are "icing on the cake" things.

src/modules/openrccfg/main.py
V3n3RiX: To be honest, I don't think this even needs to be in here. I assume distro maintainers will know exactly what services are present on the ISO image, so this check is kinda redundant.

src/modules/openrccfg/main.py
V3n3RiX: rc-update add/del service only creates/removes a symlink from /etc/init.d/service to /etc/runlevels/runlevel/service ... so unless the service file is missing it will always succeed ... 

at this point, my previous comment may look redundant (if exists is needed after all), but I stand by it : distro maintainers should know exactly what services are present on the ISO...

src/modules/openrccfg/openrccfg.conf
V3n3RiX: altough OpenRC allows adding same service into multiple runlevels, this is a bad behaviour...OpenRC is a dependency based init system, everything must be started at a specific point...I don't want to encourage people to add say NetworkManager to sysinit runlevel, which is only intended to start critical services (udev, devfs, sysfs)...

that's why I really think this module should only limit itself to the default runlevel...other runlevels (sysinit, boot) are critical to bring the box up, so the installer must not touch them...

src/modules/openrccfg/main.py
udeved: Don't be pedantic here please @V3n3RiX 
This is an installer module, not a thing to play around with for users.
Reason for a variable holding a runlevel is elogind, which needs to be in boot runlevel. Some distros may not have set it, and I want to keep it open for such cases.

udeved: The check for the service to exist is error tolerance. I find it better than catching exit code.

abucodonosor: @udeved  
And that is a $Distribution problem not the installer one.
If you write configuration files on ISO creation then be sure they do
the right thing.

It is insane to allow touching boot runleves.

Well again this is a problem of that Distro to NOT taking care of sane
defaults.

I agree with @V3n3RiX  here .. allow default and nothing else.

udeved: Well, I can also withdraw the PR. It is really quite simple, either live with a variable holding a runlevel, or don't and I withdraw.

abucodonosor: @udeved 

don't get me wrong but everyone seems to (ab)use
calamares to push any kind weird workarounds
for broken ISOs , packaging issues etc , upstream.

Yes you offer something and we all discuss a good way to
get it in.

Eat that or die is not going to work , is really simple.

udeved: I am not the nanny to enforce whatever on distros.
Variable runlevel will stay as it is.

V3n3RiX: I don't disrespect you, and if I said something that makes you feel that way, I apologize ... I happen to maintain an OpenRC based distro, and I know for a fact that some things can be done outside of the installer, that's why my strong disagreement with the design of this module ... Everything I said was from a technical view, nothing personal, and I'm sorry you feel that way ...

udeved: I was basically complaining about @abucodonosor tone @V3n3RiX 

I know you maintain an openrc distro, and I am generally willing to contribute a module for non systemd. Hence the PR, but the variable holding the runlevel won't be changed, no way, its a featrze distros may or may not use. No harm is done with a variable, and if distros responsibly use the module, its fine. If they use only default runlevel, its fine, if some distro wishes to add a different runlevel, the code supports it. Its not only elgogind eventually, it could be lvmetad whatever some distro may require on iso.

src/modules/openrccfg/main.py
kkofler: @V3n3RiX:
> I assume distro maintainers will know exactly what services are present on the ISO image

This is not really true. There can be custom remixes of the ISO based on the packages (including the Calamares configuration) from the upstream distribution. The upstream distribution (which writes the configuration) has no way to know what ends up on such a custom ISO.

src/modules/openrccfg/main.py
kkofler: I am in no way an OpenRC expert (I use systemd exclusively), but I also don't see why the runlevel to use needs to be hardcoded.

udeved: @kkofler 

Precisely, tomorrow, a distro dev of an perhaps unknown openrc distro comes and asks for a feature to set a different runlevel, It may even be a custom runlevel, eg nonetwork whatever. We would add back stuff we just ripped out. I find the insistence to hardcode the runlevel not wise.

src/modules/openrccfg/main.py
V3n3RiX: I don't think is upstream distributions job to even care about what ends up in such remixes. You have your own Fedora remix, do you take everything as is, or do you make adjustments to suit your needs?

kkofler: Well, the situation is a bit particular because I maintain both the remix (Kannolo) and the Fedora package for Calamares. I tried hard to make the configuration shipped by the Fedora package work for remixes too. (I even pick up the branding from where a Fedora Remix normally puts it to autogenerate a Calamares branding.) I still do a couple of tweaks to the Kannolo Calamares configs, but the idea is that Calamares ought to work even without those tweaks.

udeved: Interesting @kkofler 

We take a slight different approach to the conf files.
upstream installs the standard conf files in /sur/share/calamares/modules, we override them in /etc/calamares/modules.
But, the build script for iso writes some of the distro overrides conf, eg services based on the build script's conf. Ie, we hand over the the conf values of the buildscript to calamares module conf files where required to get around the issue of packaging hell with basically variable conf.

kkofler: The exact way the conf files are handled in the Fedora Calamares package is that I install my configuration to `/usr/share/calamares/modules` (I patch the upstream conf files to the settings I need), and remixes such as Kannolo can then write further customizations to `/etc/calamares/modules` (which ships empty in the package).

src/modules/openrccfg/main.py
bionel: I think this discussion deviated a lot :) .. but may be just me.

IMHO, an installer should do only his intended job: do some preparation on the HDD, deploy the system, do some configs. But those configs should be kept to a minimum, to the level of 'what an inexperienced user can choose and handle'.

Major system tweaks, at the level where the system can easily render un-bootable/unusable should be handled by the distro devel/maintainer, otherwise the installer is no longer an installer but tend to became a patching/hacking tool.

I am watching calamares since it was a small octopus and I loved exactly the idea of modularity: you can have as much as many modules you want for any particular case.
And if one does not exist for an edge case, the distro dev can easily add his own, or tweak an existing one, the actual modules are written with self explanatory code for my eyes.

But having mega-super modules, tweaked for all possible scenarios, I think is a terrible idea. The code should be kept to a level of 'do what is generic and really useful'.
I really don't understand why other runlevels are needed ... even the author says is not using them in fact, the rest of the cases where 'may be needed later' are just speculations.
I agree, hard-coding it is not a good idea and i hate such too ... but in the end, the discussion is:
There is **a real** need for **other** runlevels to be handled by the installer?

bionel:
not particularly related to the actual discussion:
- the name of the module may me a bit confusing. A name containing 'cfg' leads me to the idea it is configuring that service/package .. which may not be 100% true, in fact the this module is just enabling/disabling others.
May be more suggestive if the module is named 'openrc-services' ?
(and the actual existing 'services' module renamed to 'systemd-services?')

I know, some distro maintainers will scream if suddenly an existing module is renamed, but in my opinion will be more suggestive ....

src/modules/openrccfg/main.py
bionel: > Precisely, tomorrow, a distro dev of an perhaps unknown openrc distro comes and asks for a feature to set a different runlevel, It may even be a custom runlevel, eg nonetwork whatever.

Good point!
So, what is the limit to extend calamares? Put into it 'whatever somebody may dream tomorrow' .. or keep it at a decent level of flexibility and stability (which still has to be defined, but that is a different story .. )?

src/modules/openrccfg/main.py
bionel: as for the 'hard-coding' issue, the solution may very simple:
```
SYNOPSIS
     rc-update [-s, --stack] add service [runlevel ...]
 # rc-update add cupsd
 * rc-update: cupsd already installed in runlevel `default'; skipping
# rc-update del cupsd
 * service cupsd removed from runlevel default
 # rc-update add cupsd
 * service cupsd added to runlevel default
```
There is no need to use 'runlevel' at all. Without it, rc-update will default .. to *default*.
And I think there is a serious reason to be designed like that ... but we have to ask the openrc dev guys for it ...

src/modules/openrccfg/main.py
bionel: as for the 'hard-coding' issue, the solution may be very simple:
```
SYNOPSIS
     rc-update [-s, --stack] add service [runlevel ...]
# rc-update add cupsd
 * rc-update: cupsd already installed in runlevel `default'; skipping
# rc-update del cupsd
 * service cupsd removed from runlevel default
# rc-update add cupsd
 * service cupsd added to runlevel default
```
There is no need to use 'runlevel' at all. Without it, rc-update will default .. to *default*.
And I think there is a serious reason to be designed like that ... but we have to ask the openrc dev guys for it ...

udeved: @bionel 
There is a real need for other than default runlevels to be handled by the installer?


Are you guys seriously asking to hardcode "default" runlevel? Seriously?
The distro provides the conf, and if the conf has only default in it, all fine.
I will not hardcode a specific runlevel, if it can be configured via conf already.

bionel: @udeved 
See my latest comment. In fact, we suggest to remove the runlevel completely, as default is a sane default also by openrc design

udeved: @bionel 

I reject such thing.
I regard your insistence to take away a feature already present has absolutely no valid reason.

udeved: That said, I take for example arch philosophy into account, who would discourage from setting runlevel or so with a package. I may not like it, but it exists, and it is taken into account.

bionel: @udeved 
well .. i think you misunderstand something.
i am not suggesting anything, in the end it is not my business what will be merged or not, as i am not a very active developer (my time for hobby project is limited)
My comments are from a technical perspective and from my experience with other stupid mistakes I made in the past
If you take this personally, than I am sorry, this was not my intention.

I am trying to be neutral, even if I am involved at a certain level in redcore development.
I always try to keep my mind away for what I am currently involved in and trying to see both positive and negative aspects.

Indeed, having access to runlevels may be regarded as a 'feature'.
But the question remains: It is sane to have it at the installer level? And is it really needed?
With great features come greater responsibilities ...

Such feature may open a door for further potential issues caused by wrong usage.
Will you deal further with all potential issues raised by miss-usage of this?
Or once the code is merged into upstream ... no longer my problem, calamares team should handle it?

udeved: @bionel 

I return the question.

 It is sane to have it at the installer level? And is it really needed?

Yes, ansolutely sane, since the distro devs chose the runlevel for the service, and distro devs may have a reason to quickly test or configure non default.

If the need arises to have a different than default runlevel configured, what are distro devs gonna do? Patching, readding a lost feature because people insisted to have a hardcoded runlevel? What is you suggestion there?

bionel: > Patching, readding a lost feature because people insisted to have a hardcoded runlevel?
As i said, there is no need to hardcode anything. Just stick with the default.
Different runlevels has to be handled at a different development stage (e.g before/after squash building.
And that will be quicker too, because sanity testing should be done before actually packing/shipping the ISO image.
On the ISO himself, only smoke/regression testing is done in a sane development cycle.

udeved: @bionel 

Do you understand the issue we discuss?
The code has no reference to a specific runlevel, it loads it from conf. 
Some insist to remove this in favor of setting a default runlevel in code, which is crazy in my view.

bionel: @udeved 
Did you read the rc-update manpage?
```
SYNOPSIS
     rc-update [-s, --stack] add service [runlevel ...]
```
So, adapt the code, remove the level hardcoding from the module and from the conf. Problem solved.

udeved: As long as gentoo documentation on their wiki, in the ebuilds, on their forums use explicitly the openrc concept of named runlevels, I will not change it. Your point is moot.

udeved: @bionel 

"change the code" you demand, to what?

Gentoo have calamares in portage
A module is supposed to cover as many distros as possible
I neither cater to redcore specifically, nor to artix, the module is meant to be as flexible as possible.
Gentoo, the mother of openrc, specifically use 

rc-update <add/del> <svc> <runlevel>

all over the place. 
The code of the module does the same, just as rc-update is intended to be used, and that will not change.

bionel: I give up, i am out of ideea's
If the arguments are 'gentoo have that' without actually knowing how/when it is used, or 'i wrote that and i don't want to remove it' ... there is no logical reason for me to continue the thread.
I let the decision to calamares staff, in the end it is their team which will have to deal further with the code, once merged.

udeved: Technically, I am part of calamares staff.
I can reduce the issue to its core, that is, some insist to hardcore the runlevel and force consuming distros to be dictated by upstream module what runlevel they can use. If they have different needs, they are supposed to patch the module.
No way.

abucodonosor: @udeved 

While hardcoding is not nice , allowing touching critical runleles that way
is broken too.

OpenRC by itself is not as smart to dissallow you breaking your system
my enabling/disabling services in critical runlevels.

We do not allow that to systemd , while 'possible' and very easy 
to add that code and it should not be allowed to any other implementation.

From 'Calamres' point of view , whatever you agree or not , there are 
*two* cases ( from which one is not yet fully implementet ) to touch boot
runlevel in OpenRC.

* 1 LUKS
* 2 LVM

For case 1 that could be done in the openrcdmcryptcfg if needed.
For case 2 we need wait and see how the end implementation looks like.

However I see only 2 ways of allowing touching boot and sysinit

* 1 Allow ony 'add' to everything ( even when bad services can be added
     that is really Distribution problem ) and filter boot && sysvinit for 'del'
* 2 Filter boot and sysinit for both cases and allow whatever else

Also this is not about 'dicating' anything but about having something
sane ( or almost sane ) upstream , since we the ustream have to
maintain and deal with user bug reports , not you.

'No way' .. 'damn eat my code' is not going to work when you want
to get your code upstream.

I'm not against you or your Distribution , it doesn't matter to me who uses that.

I would tell @V3n3RiX or @bionel or even any gentoo peoples the same thing.

ISO creation things , how we do remixes and these things doesn't matter at all here.

It also does not matter what could be done but what should be done , 
do you agree ?

V3n3RiX: "Gentoo, the mother of openrc, specifically use
rc-update <add/del> <svc> <runlevel>"

You do understand that is only a syntax right? Runlevel is only needed when for some reason you want/need to add a service to a specific runlevel. Otherwise the service it will be added to 'default' runlevel.

shakuras /etc/conf.d # rc-update add fancontrol
 * service fancontrol added to runlevel default
shakuras /etc/conf.d # rc-update del fancontrol
 * service fancontrol removed from runlevel default

So, I adjusted my PR to go with OpenRC's defaults, removing the hardcode for the runlevel. I will also adjust **openrcdmcryptcfg** for when LUKS is used and when LVM support comes into calamares I'll also take care of that one. 

Calamares is a high priority in Redcore, and I intend to fully support the code I wrote for it, and expand it when good arguments are being made, when required. I'll not just add code for the sake of "one days pigs may fly, and some other unknown openrc distro **may** want to do this".

Just as a side note : calamares is indeed in Gentoo's tree, and I'm happy to report, I'm the one who wrote the first ebuild for it, back in calamares 1.1 days, about 2 years ago.

udeved: So,what is all the fuzz about?
You want to enforce a default only, I am against that.
If you modules goes upstream, I won't use it and will stick with a flexible module that allows any runlevel, even though I may not use it right now.
I still fail to see the in my view ridiculpous resistance to a runlevel variable, and why we have two competing PRs.

V3n3RiX: don't use it then...I won't use yours if it goes upstream...I'll even ban it from my calamares package, so it never gets installed in the first place...since what this module can do is beyond insane...

bionel: maybe because you refuse, without any logical reason, to improve the code you wrote?
if you don't want to remove the runlevel's usage, than treat the dangerous cases (prevent it to remove services from from boot and sysvinit level at a minimum)
Is that hard to add few lines of code to make it proper?

Also, the 'error handling' can be improved, but seems like you know the only one way of doing things ... and i am sorry to say, that doesn't looks like a sane/safe way ...

udeved: I won't do any further work with two competing PR. This needs to be resolved first.

bionel: as i thought from the beginning .. it is not about the right way to do it .. but it is about who is doing it :)

well .. mixing ego into the code is not the best way to properly contribute to an open project ....

adriaandegroot:
Please leave a comment in #992 ; that asks for (a) civil discussion and (b) long detailed answers with examples.

adriaandegroot:
See closing comments in #992

Issue #967, opened 2018-05-28
[timezonewidget] wrong currentZoneImage for Reykjavik


oberon2007:
#### Submission type

  - [x ] Bug report
  - [ ] Feature Request

#### Info regarding which version of Calamares is used, which Distribution

> v3.2.0 on Manjaro

#### description?

> Reykjavik, Iceland belongs to timezone UTC+00:00, which is actually depicted correctly on the timezone map. So When locale Atlantic/Reykjavic is selected, timezone GMT should be highlighted.
Instead, the highlight is placed on GST:

![london](https://user-images.githubusercontent.com/11800714/40607184-0d7335ae-6268-11e8-9a28-3d34215d6504.png)

![reykjavik](https://user-images.githubusercontent.com/11800714/40607194-1053ed54-6268-11e8-85d7-db7a8ee3bc65.png)

kkofler:
The thing is that Reykjavik does not observe DST (it is always on UTC year round) whereas London observes DST in summer, i.e., it is on UTC+01:00 right now. So they are not really in the same time zone in summer. That said, Iceland should of course be highlighted if the Reykjavik zone is selected.

oberon2007:
Something seems to be off with locations in that area in general. The pin for Reykjavik is displayed too far South. Same happens for example for America/Scoresbysund or America/Danmarkshavn.

oberon2007:
In fact the pin for Reykjavik is displayed too far South, so actually in the wrong zone.

oberon2007:
Hm. I guess what actually needs to be adjusted is the map. But if you would just move Iceland to where the pin goes, this will place it more south than the tip of Greenland. Looks like our map projection isn't really compatible with the coordinates provided be **tzdata** there :wink:

oberon2007:
If you look closely you will also find that Copenhagen is practically displayed in Germany and Helsinki somewhere in Estonia or Latvia ... 
Since the pin is rather big in relation to the map, maybe we can just use some (non-transparent) blue where about the Reykjavik lovation is displayed. Or maybe a compromise of moving Iceland a bit, plus apply some opacity even more south of that...

oberon2007:
Moving Iceland just a little bit seems to do the job:

![screenshot_2018-05-28_15-35-58](https://user-images.githubusercontent.com/11800714/40614801-8a403836-6284-11e8-84e6-691f1f28cd77.png)

[...]

PR #968, opened 2018-05-28
move Iceland a little South

oberon2007:
to get Reykjavik displayed in the correct timezone
fixes https://github.com/calamares/calamares/issues/967


adriaandegroot:
As mentioned in #967, we shouldn't be editing the picture except to fix TZ detection; moving islands / Icelands around in the ocean is not going to happen.

PR #972, opened 2018-06-01
[modules]: services.conf stop enabling thing by default

abucodonosor:
- That is a Distribution decision and every Distribution
  *need* edit these to match what they want to do.
  I saw a lot of disabling pacman-init in random logs from issue
  tracker which indicates some Distributors forget to set these right.
  Switch default to 'nothing' like it should be.

### DIFF
--- a/src/modules/services/services.conf
+++ b/src/modules/services/services.conf
@@ -1,20 +1,24 @@
 ---
-#systemd services and targets are enabled in this precise order
+# systemd services and targets example:
 
-services:
-  - name: "NetworkManager"  #name of the service file
-    mandatory: false        #true=> if enabling fails the installer errors out and quits
-                            #false=>if enabling fails print warning to console and continue
-  - name: "cups"
-    mandatory: false
-
-targets:
-  - name: "graphical"
-    mandatory: true
-
-disable:
-  - name: "pacman-init"
-    mandatory: false
+# services:
+#   - name: "NetworkManager"  # name of the service file
+#     mandatory: false        # true=> if enabling fails the installer errors out and quits
+#                             # false=>if enabling fails print warning to console and continue
+#   - name: "cups"
+#     mandatory: false
+#
+# targets:
+#   - name: "graphical"
+#     mandatory: true
+#
+# disable:
+#   - name: "pacman-init"
+#     mandatory: false
 
 # Example to express an empty list:
 # disable: []
+
+services: []
+targets: []
+disable: []


kkofler:
How so? I think having portable default configurations would be a win. I also do not see any heated discussion on this particular pull request, so can you please unlock the conversation for all users?

adriaandegroot:
If you are touching this file anyway, please document it properly while you're at it. This change leaves just one big cargo-cult example in, followed by three empty-list settings. By writing documentation explaining what the different lists do, what kind of entries are needed there, you make it possible for downstreams (distro's) to do something relevant with the file.

Then, independently of that, there's the "distro's ship these example files without changing them, which is stupid and potentially dangerous" issue. Changing the default file to not do anything **may** send a message to those distro's (or it might just confuse them).

src/modules/services/services.conf
```diff
@@ -1,20 +1,24 @@
 ---
```
adriaandegroot: Should get a top-of-file comment "systemd services configuration. this allows post-install changes to systemd configuration by *something-something-something*".

src/modules/services/services.conf
```diff
@@ -1,20 +1,24 @@
...
-#systemd services and targets are enabled in this precise order
+# systemd services and targets example:
```
adriaandegroot: Needs documentation on the meaning of services, targets, and disable. "There are three keys in this configuration, ... . Each holds a list of systemd units (?) to *something-something*. Use [] to explicitly denote an empty list (which is the default value for each key). .."

src/modules/services/services.conf
```diff
@@ -1,20 +1,24 @@
...
-  - name: "graphical"
-    mandatory: true
-
-disable:
-  - name: "pacman-init"
-    mandatory: false
+# services:
+#   - name: "NetworkManager"  # name of the service file
+#     mandatory: false        # true=> if enabling fails the installer errors out and quits
+#                             # false=>if enabling fails print warning to console and continue
```
adriaandegroot: Those three end-of-line comments should be above, describing what entries / keys are needed in each of the lists.

src/modules/services/services.conf
```diff
@@ -1,20 +1,24 @@
...
+#     mandatory: true
+#
+# disable:
+#   - name: "pacman-init"
+#     mandatory: false
 
 # Example to express an empty list:
 # disable: []
+
+services: []
```
adriaandegroot:
```
# By default, no services or targets are changed (so you may as well disable the module)
```

adriaandegroot:
Closing since the requested documentation improvements have been written (by me) in a different commit.

PR #974, opened 2018-06-01
openrc-services module

V3n3RiX:
* openrc-services module
* why reivent stuff? take existing code from systemd one and reuse it
* runlevel is static and 'default' only, any other runlevel setup is insane

This is my implementation of openrc-services module, thanks to @abucodonosor for forcing me to write this.

As my comments in the other PR with similar functionality, any other runlevel setup is redundant, so this is limited to 'default' only. I reused as much code as possible from systemd one...

### DIFF
--- a/src/modules/services-openrc/main.py
+++ b/src/modules/services-openrc/main.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# === This file is part of Calamares - <https://github.com/calamares> ===
+#
+#   Copyright 2014, Philip Müller <philm@manjaro.org>
+#   Copyright 2014, Teo Mrnjavac <teo@kde.org>
+#   Copyright 2017, Alf Gaida <agaida@siduction.org>
+#   Copyright 2018, Gabriel Craciunescu <crazy@frugalware.org>
+#   Copyright 2018, Ghiunhan Mamut <venerix@redcorelinux.org>
+#
+#   Calamares is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   Calamares is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
+
+import libcalamares
+
+
+def run():
+    """
+    Setup openrc services
+    """
+    services = libcalamares.job.configuration['services']
+    disable = libcalamares.job.configuration['disable']
+
+    # enable services
+    for svc in services:
+        ec = libcalamares.utils.target_env_call(
+            ['rc-update', 'add', '{}'.format(svc['name']), 'default']
+            )
+
+        if ec != 0:
+            if svc['mandatory']:
+                return ("Cannot enable openrc service {}".format(svc['name']),
+                        "rc-update add  call in chroot returned error code "
+                        "{}".format(ec)
+                        )
+            else:
+                libcalamares.utils.debug(
+                    "Cannot enable openrc service {}".format(svc['name'])
+                    )
+                libcalamares.utils.debug(
+                    "rc-update call in chroot returned error code "
+                    "{}".format(ec)
+                    )
+
+    # disable services
+    for dbl in disable:
+        ec = libcalamares.utils.target_env_call(
+            ['rc-update', 'del', '{}'.format(dbl['name']), 'default']
+            )
+
+        if ec != 0:
+            if dbl['mandatory']:
+                return ("Cannot disable openrc service"
+                        "{}".format(dbl['name']),
+                        "rc-update del call in chroot returned error code"
+                        "{}".format(ec))
+            else:
+                libcalamares.utils.debug(
+                    "Cannot disable openrc service {}".format(dbl['name'])
+                    )
+                libcalamares.utils.debug(
+                    "rc-update del call in chroot returned error code"
+                    "{}".format(ec)
+                    )
+
+    return None
--- a/src/modules/services-openrc/module.desc
+++ b/src/modules/services-openrc/module.desc
@@ -1,0 +1,6 @@
+---
+type:       "job"
+name:       "services-openrc"
+interface:  "python"
+requires:   []
+script:     "main.py"
--- a/src/modules/services-openrc/services-openrc.conf
+++ b/src/modules/services-openrc/services-openrc.conf
@@ -1,0 +1,24 @@
+---
+##
+# NOTE: allowing any other level but default 
+# would be insane , therefore 'level' is static
+###
+###
+# openrc services example:
+#
+# services:
+#   - name: "NetworkManager"  #name of the service file
+#     mandatory: false        #true=> if enabling fails the installer errors out and quits
+#                             #false=>if enabling fails print warning to console and continue
+#   - name: "cups"
+#     mandatory: false
+# disable:
+#   - name: ""
+#     mandatory: false
+#
+# Example to express an empty list:
+# disable: []
+###
+
+services: []
+disable: []


udeved:
Ok, go with your module, its now pointless to have two modules, I'll close my PR. 
I keep patching.

V3n3RiX:
It's pointless to have 2 modules, true, but it's not pointless to have 2 PR's and let @adriaandegroot decide ...  I made some comments on your PR, this is just to show how the module would look according to those comments and to show I'm not just a blabbermouth ...

udeved:
@V3n3RiX 

This isn't a **** waving contest, really. I know you are an able distro guy, I was just absolutely discontent with the other guy's tone. 
But really, it is more insane to statically do runlevels, only topped by people who set crazy stuff. Then again, this is a installer module, and I trust distro devs they know what they are doing. I just do not get along with enforcing stuff on others, and I like to keep code more flexible.
Calamares is rather a minor focus on artix, time for installer is sparse, if you maintain and develop other stuff.

V3n3RiX:
We're probably the only 2 OpenRC distros using Calamares as installer and we have a different point of view of what the installer should do regarding services.  The code in this PR is what **I** think is a sane approach, since 99% of services belong to 'default' runlevel (multi-user.target.wants in systemd world) ... 

Your own ISO scripts only touch 'default' runlevel (I took a quick look) same as ours ... It's a bit weird to me you want to expand that to do more stuff in calamares instead ... I guess we will have to agree we disagree ...

kkofler:
I think having 2 competing modules doing the same thing floating around is not going to help anybody in the long run. This is going to be `bootloader` vs. `bootldr` all over again. Can you please agree on one implementation covering all use cases? If some people think that adding services to non-default runlevels is useful, then surely that should be an option. All these are packager options and not settable by the average end user anyway.

bill-auger:
i plan on using calamares to install openrc also - if youre looking for a tie breaker i dont think it is necessary to refer to the specifics - probably logic should be enough here

if the argument is that 99% of cases are "foo" and so therefore "foo" should be hard-coded in; that is the same as saying that anyone who wants to do something in the remaining 1% will need to patch - so the question would be: is this too much work to support that 1% use-case? or would it be so difficult to implement that it should be left to the special case that requires it (and maybe they will send in a patch here)? or would accommodating that 1% use-case? be problematic for the 99% general cases?

what exactly would be the value of hard-coding anything unless there is a critical reason to specifically prevent any alternative? what could possibly be the downside of leaving the option open? after all, an option that is not changed from its default is equivalent to the hard-coded mandate - yes?

V3n3RiX:
my argument is distro maintainers should do their part, and try not to force distro specific workarounds into the installer...calamares is an installer and shouldn't be used by distros as a fix-up tool for some internal quirks...

the only argument for supporting boot runlevel is elogind, which is not a requirement, and I already explained myself in the other PR...

but why stop there? ... let's support sysinit runlevel as well which is much more insane ...

V3n3RiX:
let's give some examples : 

Runlevel: sysinit
 sysfs                [  started  ]
 devfs                [  started  ]
 udev                 [  started  ]
 dmesg                [  started  ]
 kmod-static-nodes    [  started  ]
 opentmpfiles-dev     [  started  ]
 udev-trigger         [  started  ]
 udev-settle          [  started  ] 

Runlevel: boot
 hwclock           [  started  ]
 modules           [  started  ]
 lvmetad           [  started  ]
 lvm               [  started  ]
 device-mapper     [  started  ]
 dmcrypt           [  started  ]
 fsck              [  started  ]
 root              [  started  ]
 mtab              [  started  ]
 swap              [  started  ]
 localmount        [  started  ]
 procfs            [  started  ]
 sysctl            [  started  ]
 bootmisc          [  started  ]
 termencoding      [  started  ]
 consolefont       [  started  ]
 urandom           [  started  ]
 loopback          [  started  ]
 binfmt            [  started  ]
 keymaps           [  started  ]
 hostname          [  started  ]

These are pretty much **all** services that **need** to be started from these 2 runlevels. Do we really want calamares to even thouch them ? These runlevels are usually touched by packages or packages-scripts, and you'll find these enabled in both live and installed environment alike ... 

In a way, these are systemd's basic.target or whatever it's called. Systemd services module won't allow messing with these (systemd cant do it anyway) , why should openrc services allow it? Systemd services module plays around with multi-user.target only, so this module does the same : plays with 'default' runlevel only.

udeved:
> my argument is distro maintainers should do their part, and try not to force distro specific workarounds into the installer

Uhm, because you think distros do something you dislike, you also think its your duty to force these distros to not do what they do? 
Who is forcing you to use a workaround with the module? And why do you want to force distros to stick to a specific runlevel?

V3n3RiX:
No, I'm not trying to force you to do anything. You can put whatever crap you want in your distro. What I'm trying to prevent is that crap spreading into calamares. You may not care about sane design choices and use duct tape all over the place, but I do care. It's true that I'm not forced yo use the workarounds, I don't need them in the begin with. But you really shouldn't force **your** workarounds into upstream code.

udeved:
>But you really shouldn't force your workarounds into upstream code.

Where do I do that? Amazing, because I don't.
I simply don't get this whining about nothing.
You provide the service conf, you decide what runlevel. The code does not care for the runlevel name, it is configured in the conf.
If some distro wants to have a office runlevel because the distro is geared towards office, they can do that.

V3n3RiX:
> If some distro wants to have a office runlevel because the distro is geared towards office, they can do that.

I was never against custom runlevels which OpenRC supports, I was only against sysinit and boot runlevels. And even if a distro wants custom runlevels, that piece of code of yours won't be enough for it to boot into that custom runlevel. You need to touch other parts of calamares as well. I'll let you guess which ones.

udeved:
I am basically done going around in circles. You give no valid reason why to remove a variable in favor of a fixed runlevel. Again, this is no competition to prove an ego. 
dmcrypt is something I would not want in the service module, because its done differently on arch with initcpio.
If you configure calamares to remove essential services from a sysinit runlevel, then that is frankly your problem, and you obviously did something wrong with package deployment.
I am against trying to dictate distros how they do things, which is what your module does.

V3n3RiX:
I can say the same thing : if you need calamares to configure **boot** runlevel just for **elogind** (your comment, not mine), then you obviously did something wrong with package deployment. And if I take in consideration your latest **manual intervention required** news, then I'm sure you did just that. Fix your own broken thing, and do not try to force your workarounds in here. Even if this module is slightly more limited, it is useful for Artix (you do have a default runlevel don't you?), whereas yours is of no use for Redcore. Being Gentoo based we do things as upstream intended, we don't have our own speciffic workarounds.

And dmcrypt being done different in arch is once again, **your** problem. In OpenRC that service goes into **boot** runlevel. This is how it's done upstream in Gentoo, and over here in Redcore. 

So please, once again, keep your distro speciffic workarounds to yourself.

udeved:
> So please, once again, keep your distro speciffic workarounds to yourself.

Oh the irony, now you say, its done so on redcore and gentoo, and this must be valid on all distros using openrc.
Btw, nope, we don't need to configure any boot runlevel on artix. You are wrong.
I think you are getting way too defamatory. and I gonna leave this useless discussion now.

V3n3RiX:
If you don't need it, why try force it in here then? And if you don't need it, then this module will do just fine on Artix. I'm damn sure you do need to configure **default** runlevel. 

Default runlevel is valid for all OpenRC distros ... What you're trying to push is not. That's what I'm trying to explain to you even since the beginning ... I think I rest my case!

V3n3RiX:
Btw, if dmcrypt service is really handled different on Arch based distros, I won't try to merge the piece of code I wrote...I don't wanna **force** things ... Just wanna have **sane** defaults ... that's all

abucodonosor:
It can be handled in any OpenRC Distro equal .. no matter the initramfs impelmentation.

But is this different issue.

abucodonosor:
> I am against trying to dictate distros how they do things....

That does not matter .. Is your personal opinion and not an technically reason.

Calamares restricts all kind things already , and 'dictates' already in some places
what is allowed what is enforced and what won't be upstream , for the one or the other
reason.

> If you configure calamares to remove essential services from a sysinit runlevel, then that is frankly your problem

Again .. is not about you.. Once is upstream it will not be you who takes the bugs.
Is not you who need write some sane docs , etc ..

It just shows your bad attitude.

Technically systemd services could have support for disabling targets 
or masking any kind things.. Now guess why that is not upstream and won't be..

All about this is to find sane defaults not about what could be done.

I've told in the other PR already runlevels support can be added with
some filtering but yet you seems to ignore everything.

adriaandegroot:
Please leave a comment in #992 ; that asks for (a) civil discussion and (b) long detailed answers with examples.

adriaandegroot:
See closing comments in #992

PR #975, opened 2018-06-03
services allow 'systemctl mask <whatever>'

abucodonosor:
!!!NOT for merging , please please don't!!!
 Demo patch to show that 'could be done' should 'just not be done'
 Even it looks flexible _that should never end in any serious upstream_
 Having that in a Distribution , as own patch fine , devels *may* know what
 they are doing. Opening the pandora box exposed as option for users ohh ,
 have fun to handle the storm of bug reports.

 But wait is not a user option ?
 Well is not but be sure they'll use it..
 And be sure even devels will trap into badness with these.
 ( You think I'm wrong ? Use search button on issue tracker and figure )

### DIFF
--- a/src/modules/services/main.py
+++ b/src/modules/services/main.py
@@ -30,11 +30,11 @@ def run():
     services = libcalamares.job.configuration['services']
     targets = libcalamares.job.configuration['targets']
     disable = libcalamares.job.configuration['disable']
+    mask = libcalamares.job.configuration['mask']
 
-    # note that the "systemctl enable" and "systemctl disable" commands used
-    # here will work in a chroot; in fact, they are the only systemctl commands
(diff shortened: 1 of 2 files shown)


abucodonosor:
Jus demo code .. really do not merge that..

kkofler:
IMHO, this is a valid feature and should be added (but as a new pull request without all your negative comments). I do not see a valid reason to not support `mask`, given how trivial it is to support it.

adriaandegroot:
Please leave a comment in #992 , as this PR reflects on the more general issue of service-management outside of the package manager. And, as @kkofler says, it needs work to be mergeable.

adriaandegroot:
See closing comments in #992

PR #983, opened 2018-06-11
[modules]: fix broken logic introduced with plymouthcfg modules

abucodonosor:
plymouthcfg is optional. Also Distributions with sane packages
  won't even want to use it ( even docs suggest that ).

  However , now grubcfg literally depends on plymouthcfg to run
  bc the gs value to add 'splash' to grub is set in plymouthcfg.

  Revert to calamares 2.x logic and drop exporting gs value
  in plymouthcfg since pointless.

  Fixes ea1c8a0e5ddfc6b72311cf744e2ac514c6630009


philmmanjaro:
Looks fine for me.

adriaandegroot:
Merged after re-writing the commit message. Can we leave the editorializing out of those messages?

Issue #610, opened 2017-01-20
install latest package updates from archive


calamares-user:
Jira issue originally created by user jriddell:

a nice feature would be to offer to install the latest updates from the relevant package archive

<Kevin_Kofler> teo-, Riddell: I think installing updates is something we want to add support for to the packages module, for all the supported package managers.
<Kevin_Kofler> It can be done through the package manager's CLI just like the other operations.

adriaandegroot:
The idea would be to check for network connectivity (?), then run an update of some sort. If that fails, I don't think it should abort the installation though -- the system is installed, just not totally updated.

It's straightforward to add a method to the package manager class and add implementations where we have one.

kkofler:
The idea was to first implement the operation in the `packages` module, then offer a distro-independent UI (a checkbox somewhere) to allow the user to run it. The packager would be able to choose:
* whether to show the checkbox at all,
* whether to check or uncheck it by default, and also
* whether to require it to install packages in the `netinstall` module (because installing packages without updating the system first is pretty unreliable on some distributions such as Fedora/Kannolo). If this option is enabled and if the user opts out of updating, the `netinstall` module would just display a message instead of a list of packages, just as in the case where there is no Internet connection

adriaandegroot:
Trying to figure out the configuration space here:

 - disable
 - enabled

times (relevant only if enabled)

 - required for netinstall
 - optional netinstall

times (relevant only if enabled)

 - show in UI
 - don't show in UI (so user has no control over it)

times (only relevant if show-in-UI)

 - ticked by default
 - unticked by default in the UI (this is undesired if it is required-for-netinstall)

adriaandegroot:
The *packages* module has a configuration key *update_system* which seems to do what it wanted.

PR #989, opened 2018-06-16
Let pacman do complete system upgardes instead of partial upgardes

Foxboron:
While pacman supports partial upgrades, there are no distributions that support
this behaviour. As calamares doesn't support a "upgrade_system" routine we can't
just rely on "pacman -S" being sufficient for an installer, as it might pull in
packages that doesn't exists in the mirror. As "-Sy" might break systems because
of this, we are better off dispatching complete system upgrades instead of
leaving user systems broken.

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -242,7 +242,7 @@ class PMPacman(PackageManager):
         if from_local:
             pacman_flags = "-U"
         else:
-            pacman_flags = "-Sy"
+            pacman_flags = "-Syu"
 
         check_target_env_call(["pacman", pacman_flags,
                                "--noconfirm"] + pkgs)
@@ -251,7 +251,7 @@ class PMPacman(PackageManager):
         check_target_env_call(["pacman", "-Rs", "--noconfirm"] + pkgs)
 
     def update_db(self):
-        check_target_env_call(["pacman", "-Sy"])
+        check_target_env_call(["pacman", "-Syu"])
 
 
 class PMPortage(PackageManager):


oberon-manjaro:
The commands to install single packes with pacman are correct. `-U` for local, `-Sy` for remote.
I don't think that a full system upgrade is what we want here.
>  it might pull in
packages that doesn't exists in the mirror.

What do you mean @Foxboron ?
If a package or one of its dependency doesn't exist then somthing with the profile must be wrong. So certainly not something the installer should or could prevent by doing a full system upgrade...

oberon-manjaro:
```
def update_db(self):
check_target_env_call(["pacman", "-Syu"])
```
would certainly be wrong, as the command to update database always was and still is `pacman -Sy`

oberon-manjaro:
if I understand correctly the context how `update_db` and `install` are used here, your change would run a complete system update in the live environment in order to create package candidate lists for the target install. In some cases this would mean the useless download of hundreds of MB of packages - depending on how old the install media is, while the purpose here is merely to update the package database for installation of packages to the target.
Or am I misunderstanding something here?

Foxboron:
>would run a complete system update in the live environment in order to create package candidate lists for the target install.

Possibly me missing the context. My impression was that `pacman -Sy` is a command hitting the actual user installation, which is bad as it leads to partial upgrades. If it only touches the live ISO; It's a pass on usage.

kkofler:
While I agree with the need for a full system upgrade operation (installing packages without upgrading the full system first is not really safe on Fedora either), the `update_db` task is not the correct task to put it in. This needs a new task type in the `packages` module. See issue #610.

philmmanjaro:
Well, `-Sy` only updates the Databases and then does nothing. `-Su` will do a full update based on the given local DB. `-Syu` will do the Databases Update and then the full System Update. `-Syy` will force to do a full Database Update and not a Delta-Update on existing local Databases of given Packages. More about the switches for `S` you can find [here](https://jlk.fjfi.cvut.cz/arch/manpages/man/pacman.8#SYNC_OPTIONS_%28APPLY_TO_-S%29).

Now to the usecases:

- If you use the ISO for only installing local packages, also the same ISO it is good to keep the Database as it is. In this case the `-U` switch will be used and all dependencies should be also present on the local repository stored on the same ISO
- If you use the ISO only as Live-CD and want to unsquash the base packages from the ISO to your target System, it is needed to do a full package update on those given installed packages on that package system, to be insync with the Distro. Only then it is possible to actually install additional packages. Steps are: unsquash the ISO, update the package set installed thru the first step (`-Sy;-Su` or `-Syu`), install additional packages to the target system (`-S`)
- If you use the ISO similar to Antergos, to always install the complete target from online sources, you may want these steps: get the current database on your target system (`-Sy`), install the needed packages (`-S`). See also [pacstrap](https://git.archlinux.org/arch-install-scripts.git/tree/pacstrap.in).

Foxboron:
It is important to realize that `pacstrap` takes all packages in one go. It doesn't split it up, thus the downsides of `-Sy` is avoided.

All of the usecase mentioned by @philmmanjaro are sane for what it's worth.

Foxboron:
@oberon-manjaro It does not. Don't conflate what pacman support with what Arch Linux support.

https://gist.github.com/vodik/5660494

oberon-manjaro:
It seems you are confusing "update database" with "update system".
After running `pacman -Sy` the database will most certainly be up-to-date, no matter what packages are actually installed on the system.

Foxboron:
@oberon-manjaro I haven't confused the terms. They have been used consistently through the discussion. `pacman -Sy` and `pacman -S [newer package]` do a partial upgrade. Arch doesn't support this.

oberon-manjaro:
I think it needs to be decided based on what usecase actually calls `update_db`.
I agree that installing a single package in a not updated system can go wrong.
On the other hand, running a full system update on a live-system can also be problematic.
Also we want to avoid that in case multiple packages need to be installed that we don't run the system upgrade multiple times.

oberon-manjaro:
> pacman -Sy and pacman -S [newer package] do a partial upgrade

Not correct. Again:
* `pacman -Sy` updates the local database - it does not "do a partial upgrade
"
* 'pacman -Sy <pkgname>` updates the local db and installs the specified package only (can be problematic, yes)
* `pacman -S <pkgname> installs the package specified, assuming that the db is present at all and up-to-date.

Foxboron:
`pacman -Sy` followed by `pacman -S [package]` is equivalent to `pacman -Sy [package]`.

phillid:
I hold great concern for Manjaro one of its developers doesn't understand this fundamental concept.The gist from vodik spells the problem out very clearly.

kkofler:
`update_db` should update **only** the package database. Please do not add hacks changing the semantics of existing actions.

What we need is a **new** `update_system` action that gets implemented also for the other packaging and dependency solving programs and that may or may not be enabled by default (depending on what the majority of distros want: Arch and Fedora want system updates by default, Manjaro and I think several other distros don't).

Then it becomes very easy to make both of you happy: you just have to enable the `update_system` action on Arch and not on Manjaro, a one-line `packages.conf` change.

I can only point you again to issue #610.

oberon-manjaro:
Yes it is. Got you.
In any case what would be desireable when installing single packages on an Arch-based target install would be, running a full system upgrade **once** and then install packages with `-S`.
But running `-Syu <pkgname` multiple times is not desireble, neither is necessarily running `-Syu` for a live-environment.

Foxboron:
@kkofler Manjaro does. They are essentially blanket reusing our Arch packages from one of our mirrors and provide them to their users. Some are filtered, and they do provide some packages themselves. But this behavior is the same across both.

Foxboron:
@oberon-manjaro The commands hit the target installation, not the live environment as i understood it. Running `-Syu` multiple times isn't a big problem, albeit a slow down.

Another solution is to join join the package list and do them all in one `pacman -Syu [package list]`.

But with the current implementation it is safer for users if we blanket upgrade the system for every package.

kkofler:
Basically:
* `update_db` = `pacman -Sy`, `apt-get update`, …
* `update_system` (proposed) = `pacman -Su`, `apt-get dist-upgrade` (or `upgrade`, it will be up to the Debian maintainers to decide what makes most sense), …

Those are 2 very different actions and should **not** be conflated.

Foxboron:
A setting in the config for whether or not `update_system` should be done could be an idea. Unsure what other distributions want.

adriaandegroot:
All right, can we hit the brakes on this discussion? I don't want editorializing or off-topic comments.

We've established two things:
 - the package manager calls run in the context of the target system,
 - the meaning  of the methods `update_db` and `install` are not entirely clear.

What needs doing next is figuring out (a) where the problem lies exactly; I'd appreciate it if @foxboron could file an **issue** explaining what's going on (and explain it like you're talking to someone who uses only tar as a package manager). Then (b) whether existing semantics should be changed, or not (probably not, I'll say that in advance), (c) how issue #610 could solve your issue (running before, or after, package installs, or possibly as an addition to the packages module).

philmmanjaro:
@Foxboron: please define the exact problem with the **packages** module. With current functions it has no feature to update the target system at all. This is what currently is given:

- install local packages, which are present on the ISO on to the unsquashed packages provided by the ISO
- install remote packages from a repo on to the unsquashed packages provided by the ISO, which may fail, due no system update was performed on target installation
- install all packages from remote based on given package list to the target system. Since no packages exist on the target system, a system upgrade is not needed.

So in short: the module doesn't support option two on current version as Arch and other distros based on a rolling release model don't work with "outdated" unquashed snapshots of a set of packages to be able to install newer remote packages. So as you and @kkofler already mentioned, a new routine to update the installed target system to the current packages is needed before being able to install additional remote packages.

Therefore I **NACK** this merge request. Please work instead on an update-routine we may merge.

Foxboron:
@adriaandegroot 

* a+c) The #610 issue solves the problem, so a new issue just creates more noise.
* b) Currently in the Pacman module `install` does a `update_db` as the `-y` flag is provided. The semantics are already broken.

philmmanjaro:
@Foxboron: please tell me if this https://github.com/calamares/calamares/commit/1bd149c14e98277ae40f5d41503e5ace9a046274 is what you actually need in your case. @kkofler: you may add the needed for Fedora ...

Foxboron:
@philmmanjaro That solves the problem, yes.

eli-schwartz:
> I'd like at least some pacman-using distro people to comment on this before thinking about merging it. @philmmanjaro ?

It probably hasn't come up a lot since Arch Linux will lock and delete threads and consider banning the user, if we catch people using Calamares. We definitively don't support thirdparty installers.

That being said, I'm unaware of any Linux distro which uses pacman in which updating some, but not all, packages is safe.

And that's exactly what the -Sy without -Su command does. Or any wrapper or libalpm bindings.

>> Arch Linux, and it's derivatives, doesn't support having the database synced without a full system upgrade
> 
> It does. That's exactly what pacman -Sy does when no package name is specified.

Let's not be mendacious, please. The pacman program contains code to enable this usage, because in the grand old unix tradition, one does not restrict the user's ingenuity. And the consequence is one does not restrict the user's stupidity either.

This is a completely different topic with no common ground, than saying "the Arch Linux distribution considers this to be an officially supported action".

Because we don't. And distributions like Manjaro and Antergos, which pull packages or PKGBUILDs directly from our mirrors, are well advised to not encourage its use, since the last I checked you didn't perform the rather extensive package restructuring necessary to safely support slotted, co-installable libraries, stable sonames, and versioned soname dependencies for every single architecture-dependent package in the repos.

eli-schwartz:
I'm unsure why it is problematic to modify -Sy calls to use -Syu instead, given that
- it has been explained to only run on the target system
- it will never lead to extra work, since in all those cases you were *already* syncing the database, but anyway future calls to -Sy pkgname vs. -Syu pkgname will both only install one package.

Now, a logical replacement semantic BTW would be to switch to exclusively use -S pkgname, and if that fails then the user needs to do an update_system first. I don't see how just adding the update_system module enables the desired goal here of removing dangerous options.

Claiming the "install" semantic is -Sy pkgname is completely wrong on a technical level anyway. The pacman semantic for "installing" a remote package is -S pkgname. The code is a lie.

philmmanjaro:
@eli-schwartz: thx for your input, however I'm currently working already to ["fix this merge request"](https://github.com/calamares/calamares/tree/packages-add-update-system/src/modules/packages). Since this module is not only designed to support **pacman** it needs to be splinted. Therefore a `-Syu` is not possible.

We have to support:

- Sync DB
- Update target system (optional in some cases) **_not given yet_**
- Install local package (stored on ISO)
- Install remote package (step one and two may be needed)

Installing a remote package via `pacman Sy <package(s)>` should never had happen, as the database update happened already. Therefore for **pacman**: 1) `-Sy`, 2) `-Su`, 3) `-S <packages>`

PR #990, opened 2018-06-17
[packages] add support for target system upgrade

philmmanjaro:
Currently our **packages** module doesn't support yet a full system upgrade call. This is however needed in the following use-case:

If you use, for example an arch-based Distro, which unsquashes a snapshot of packages via the **unpackfs** module and additional adds an option to optional install additional packages via the **packages** module, a full system upgrade is needed be-forehanded. Otherwise it may result in conflicts having outdated dependencies on your target system which don't match the dependencies needed by the currently available remote packages. This also corrects the command for installing remote packages via **pacman**.

---
This makes #989 obsolete.


abucodonosor:
@philmmanjaro 

since update_system: is set to false I think update_db: should be now false too ,
and let Distributions decides to enable or disable these.

philmmanjaro:
@abucodonosor: since this is a new feature I normally disable it by default. Sure the distros should decide on how to use it.

PR #993, opened 2018-06-21
add support onLeave and onActivate to pythonqt plugins

raurodse:
PythonQt plugins don't have support to Activate and Leave events. This patch adds these events.


adriaandegroot:
That's really useful! Since you're invested in PythonQt modules, I'll do this:
 - merge this code, tidy it a little for coding style
 - re-start my fix-pythonqt branch which aims to use PythonQt also for running regular python jobs. That's a fair bit of API-wrangling, but for your use-cases, would result in just on Python interpreter instead of two inside Calamares. (And there's an issue filed requesting that, too)

PR #1030, opened 2018-09-12
Some fixes for Hungarian translation

rezso:


adriaandegroot:
Hi @rezso , the translations for Calamares are done in Transifex (see https://github.com/calamares/calamares/wiki/Translate-Guide for details), so I can't effectively use this PR. I see you have fixed a few errors in keyboard shortcuts -- that I can do-over in Transifex myself. But the actual string changes -- e.g. Telepítés instead of Telepít -- is something that the Hungarian translation coordinator should see.

adriaandegroot:
Thanks for the updates. I've done most of them by hand in Transifex now -- although it seems to me (based on looking in a dictionary) that Telepítés (noun) is not the right translation for Install (verb). But there are also other translations that use Telepítés where the verb is intended, so I'll leave that.

Issue #1046, opened 2018-11-02
Set boot flag


aliveafter1000:
https://forum.manjaro.org/t/solved-boot-loop-before-grub/63775/19

https://en.wikipedia.org/wiki/Boot_flag

"Some buggy BIOSes test if the bootflag of at least one partition is set. Otherwise they ignore the device in boot-order. So even if the bootloader does not need the flag, it has to be set to start the bootcode from BIOS.”

Currently, no boot flag is set (automated installation).
Many people have problems. Particularly old laptops.
```

    Partition* rootPartition = nullptr;
    if ( luksPassphrase.isEmpty() )
    {
        rootPartition = KPMHelpers::createNewPartition(
            dev->partitionTable(),
            *dev,
            PartitionRole( PartitionRole::Primary ),
            FileSystem::typeForName( defaultFsType ),
            firstFreeSector,
            lastSectorForRoot
        );
    }
    else
    {
        rootPartition = KPMHelpers::createNewEncryptedPartition(
            dev->partitionTable(),
            *dev,
            PartitionRole( PartitionRole::Primary ),
            FileSystem::typeForName( defaultFsType ),
            firstFreeSector,
            lastSectorForRoot,
            luksPassphrase
       );
    }
    PartitionInfo::setFormat( rootPartition, true );
    PartitionInfo::setMountPoint( rootPartition, "/" );
core->createPartition( dev, rootPartition );
```

aliveafter1000:
Also, 

```
 if ( isEfi )
    {
        qint64 efiSectorCount = bytesToSectors( uefisys_part_sizeB, dev->logicalSize() );
        Q_ASSERT( efiSectorCount > 0 );

        // Since sectors count from 0, and this partition is created starting
        // at firstFreeSector, we need efiSectorCount sectors, numbered
        // firstFreeSector..firstFreeSector+efiSectorCount-1.
        qint64 lastSector = firstFreeSector + efiSectorCount - 1;
        core->createPartitionTable( dev, PartitionTable::gpt );
        Partition* efiPartition = KPMHelpers::createNewPartition(
            dev->partitionTable(),
            *dev,
            PartitionRole( PartitionRole::Primary ),
            FileSystem::Fat32,
            firstFreeSector,
            lastSector,
            PartitionTable::FlagNone
        );
        PartitionInfo::setFormat( efiPartition, true );
        PartitionInfo::setMountPoint( efiPartition, gs->value( "efiSystemPartition" )
                                                        .toString() );
[...]

Issue #1044, opened 2018-10-29
CAL may delete partitions during disk-probe


philmmanjaro:
**Describe the bug**
During the last phase of the **v17.1** development cycle we got a forums post claiming [**CAL** may delete existing partition content](https://forum.manjaro.org/t/37298) without a user intention to do so. Seems I mislead it with https://github.com/calamares/calamares/issues/882 back then. There was a partition mounted on `/tmp/Calamares-TqYRNd`. With some talk about it a [log](https://gist.githubusercontent.com/anonymous/84ccdff59bc8b057529b03f15ae5ea49/raw) was provided. Notable line was:

> [0]: QTemporaryDir: Unable to remove "/tmp/Calamares-TqYRNd" most likely due to the presence of read-only files.

Back then I tested further and only found `/tmp/calamares-root-5aoq_pz0` mounts, which is normal during installation phase. Since I couldn't reproduce the issue back then **v17.1** got released with **CAL v3.1.11-7**.

---

In February a second user [reported](https://forum.manjaro.org/t//37298/34) a similar issue with **CAL v3.1.12-3**. In that case it was mounted on `/tmp/Calamares-EmsHcv`. In the [log](https://gist.github.com/anonymous/2fbbba9f5104aa632091ccb6d3888641) we find a similar report:

> [0]: QTemporaryDir: Unable to remove "/tmp/Calamares-EmsHcv" most likely due to the presence of read-only files.

---

Now in October a third user [reports](https://forum.manjaro.org/t//37298/36) the same issue. In this case it got mounted in `/tmp/calamares-SxYYhk` and got the same line with **CAL v3.2.2.1-3**:

> [0]: QTemporaryDir: Unable to remove "/tmp/calamares-SxYYhk" most likely due to the presence of read-only files.

Also this drive was formatted with **ext4**. This you can see on the [session log](https://forum.manjaro.org/t/37298/38). So I went ahead and did a [full lengthy regression test](https://gist.github.com/philmmanjaro/e5926529105990e05da21f449a28c9b4) with **v18.0**. Beside a small coding issue on my end I didn't found any regression. However, we seem to [found the issue](https://forum.manjaro.org/t/37298/67). Probably because it failed to umount and then QTemporaryDir deleted the contents automatically when the function returned. By default a mount is with rw rights. Auto-remove is on by default for: QTemporaryDir mountsDir; - [doc.qt](http://doc.qt.io/qt-5/qtemporarydir.html)

[...]

PR #1047, opened 2018-11-05
Set boot flag & CAL may delete partitions during disk-probe

aliveafter1000:
Set boot flag
https://github.com/calamares/calamares/issues/1046

CAL may delete partitions during disk-probe
https://github.com/calamares/calamares/issues/1044

### DIFF
--- a/src/modules/partition/core/PartUtils.cpp
+++ b/src/modules/partition/core/PartUtils.cpp
@@ -158,15 +158,14 @@ canBeResized( PartitionCoreModule* core, const QString& partitionPath )
 }
 
 
+
 static FstabEntryList
 lookForFstabEntries( const QString& partitionPath )
 {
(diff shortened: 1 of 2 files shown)


kkofler:
You should file separate pull requests for separate fixes. It is generally a good idea to not pull request from your fork's master, but to create a separate work branch in your fork for each pull request.

aliveafter1000:
Yeah, I was thoroughly confused (new at this). Oh, well.

philmmanjaro:
@aliveafter1000, do you need help with this? From the review of both patches it might work, however had no time yet to test it. Was it tested from your end?

aliveafter1000:
oops. one small mistake.

Need to check if !isEfi before setting the boot flag to root partition otherwise we now have a EFI partition marked as bootable and a root partition marked as bootable.

PR #1050, opened 2018-11-06
Prevent data destruction

aliveafter1000:
https://github.com/calamares/calamares/issues/1044

### DIFF
--- a/src/modules/partition/core/PartUtils.cpp
+++ b/src/modules/partition/core/PartUtils.cpp
@@ -163,10 +163,8 @@ lookForFstabEntries( const QString& partitionPath )
 {
     FstabEntryList fstabEntries;
     QTemporaryDir mountsDir;
-    mountsDir.setAutoRemove( false );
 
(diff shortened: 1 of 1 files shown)


philmmanjaro:
@adriaandegroot: what do you think about this PR? For me it looks merge-able ...

adriaandegroot:
- needs better commit message
 - contains a bunch of churn
 - the actual effective change is adding `-o ro,noload` ... **that** seems sensible, I think

philmmanjaro:
@adriaandegroot: I still have to test [my approach](https://gitlab.manjaro.org/applications/calamares/compare/4c73fc0...7c6ff1a3). @stikonas: do you think it is needed or does the current code we have now in **master** work already and therefore prevent data loss?

aliveafter1000:
The current code you have now mounts peoples shit into tmp.
Lotta scripts that run could destroy it. Just because QTemporaryDir doesn't destroy it doesn't mean something else can't. You only need to read the fstab.

adriaandegroot:
This is in-progress in the fix-partitioning branch.

Issue #725, opened 2017-04-29
[IMPROVE]Please add /home possibility in Auto Partitionning system


Elrondo46:
I think it's necessary for a lot of desktop distros

geckolinux:
This feature would be very useful now for use with the ZFS module, especially for running another filesystem on the rootfs but ZFS for `/home`.

Issue #901, opened 2018-02-04
[Feature Request] FS/Partition layout profiles?


polarathene:
I quite like the default one that openSUSE offers with their installer. BTRFS for the system with a rather thought out subvolume layout and XFS for `/home`. Going with the default offering also sets up Snapper integration which was another perk.

It'd be nice if Calamares could have similar "presets" or suggested layout configurations. The distro maintainer could offer these(or just a default one) as a profile, or a user could import/export a layout via Calamares to use with the installer in future installations.

---

I'm not sure what the current situation is for silent/unattended installations, being able to provide profiles for that use-case would also be nice, with constraints like min/max partition size and a % value.

abucodonosor:
dupe #725

abucodonosor:
also #525

polarathene:
@abucodonosor ? How is this a dupe of #725? That issue requests `/home` to be included in an auto partitioner.

Have I misunderstood and missed something about Calamares that allows a user or distro maintainer to have Calamares use or default to a custom partition scheme/layout? Are you familiar with the openSUSE installer and how it presents a rather advanced partition layout of BTRFS with many subvolumes and settings along with the XFS /home?

2ndly does Calamares presently support unattended/silent installs that can define said layouts to use arbitary sizes with min/max constraints and % values instead of only absolute?

#525 for Snapper mention is fine. But I disagree that this issue is a dupe of #725, it's a far more involved request.

polarathene:
I understand if the feature requested however is not something Calamares plans to support. Perhaps it is something that a third party can contribute as a module(assuming that's possible), if so and that doesn't warrant keeping this issue open for visibility, that's fine.

abucodonosor:
@polarathene 

#725 tells about /home only but is about having some sort layout/schemas.

right now only BTRFS does /home  .. also see #815 , #807 , #586

polarathene:
None of the linked issues are related to what this one requests.

I kind of get where you're coming from with #725, as it mentions to add it as a possibility/option. It is rather specific though where as this issue asks for generic partition layout/schema support with import/export(#725 does not mention anything like that, closing this issue therefore will ignore such a request?)

None of the linked issues refer to the request for min/max/% support for said FS/partition presets when doing unattended/silent installs. Closing this issue also buries that request(should I create a separate request for it?)

[...]

Issue #1043, opened 2018-10-28
Preselect already chosen drive in manual partition editor.


philmmanjaro:
**Describe the bug**
You can select the available harddrives in the selection menu of the **partition** module. When you choose manual partition mode, it automatically pre-select the first drive. Ex. `sda`. In the previous window you actually wanted to go for `sdc` instead. If you have similar setups, you might oversee this ...

**To Reproduce**
1. Select any drive, but not the first in your list
2. Select manual partitioning
3. First drive is automatically selected in drop-down on top
4. If you oversee it, you will go with the wrong drive and may delete another installation

**Expected behavior**
The pre-selected drive from the selection menu should also been pre-selected in manual partition mode. In this case `sdc` was selected and should also been pre-selected.

**Screenshots and Logs**

![image](https://user-images.githubusercontent.com/5050022/47622464-5c9e5580-db05-11e8-94f0-9d3933e40ddf.png)
![image](https://user-images.githubusercontent.com/5050022/47622471-78096080-db05-11e8-9a71-1ec194c00508.png)

Issue #1041, opened 2018-10-19
Allow for users to mount /Data partitions.


benbullard79:
**Is your feature request related to a problem? Please describe.**
I use separate partitions to store Data instead of /home. So far I call them /Data1, and /Data2.

**Describe the solution you'd like**
In older versions of Calamares I was able to add the entries in the partitioner on the blank line then select keep rather than format and the entry for the partitions was automatically created in /etc/fstab. Now this functionality is gone as there has been a change where the partitioner now has the no blank line but instead the line 'no mount point'.

**Describe alternatives you've considered**
Seems like that line could be left blank as I don't know why user would create a partition and then not have it mounted. Alternatively include a line in the partitioner for user to add a /Data partition. Doesn't seem like this could be all that difficult to do especially as the functionality was previously possible.

**Additional context**
Keeping data in a separate partition is especially helpful for multi-boot users and testers. For testers even if you bow up your test system your data is intact as it is on a separate partition.

Thanks for your consideration.

benbullard79:
Hope I made this obvious. I mean I want to have the partitioner step during Calamares install to be able to add mount points to /etc/fstab for existing Data partitions. This could or should be limited to 1 to 3 partitions to mount only? (Speculating on that.) This so the partitions are available to user after first boot in installed system. In other words a user friendly convenience or feature.

philmmanjaro:
Well, this feature is already given. Open manual partitioning. Choose any option in the mountpoint drop-down. Go into the field again and edit it as you need/want:

![cal](https://user-images.githubusercontent.com/5050022/47244742-e871fc80-d3f6-11e8-989f-05d314c39173.png)

benbullard79:
This is in OpenMandriva using Calamares 3.2.2. What you say was true until Calamares changed the partitioner from having the first line blank to the first line saying 'No mount point'. In fact I've been doing exactly as you said every since OpenMandriva switched to Calamares until now with most recent Calamares version where the first line was changed from blank to 'No mount point'.

If I delete 'No mount point' and add say for instance /Data1 no entry is listed under mount point in the partitioner so I presume it is not working. Worse is if I drop down to / and then add Data1 so it says again /Data1. What that does, at least in OpenMandriva, is somewhere during the install process all root partitions get mounted and installer won't format the requested partitions. 

Which having said all that brings up the point: Is this a Calamares issue or an OpenMandriva issue. If you could just help me confirm that it isn't Calamares that would be great.

[...]

PR #1068, opened 2019-01-08
Add the ability to support custom partition layouts

a-wai:
Hi,

For a specific use-case (system with read-only rootfs and atomic upgrades with A/B system partitions -- /, /boot & /var), I needed to automate the partitioning.
This patch series implements support for using a distribution-defined partition layout. The layout is defined in the `partition` module config file, and is automatically applied to the selected disk space:

- When user chooses to erase the whole drive, the partition layout is applied to the drive
- When another partition is replaced, or an existing partition is resized to make room for the new system, the layout is applied to the newly freed disk space

Additionally, for this use-case, the user should not be able to manually partition the drive, therefore I added an optional config option in order to disable manual partitioning.

Solves #901 

Cheers,
Arnaud


adriaandegroot:
That's a really good idea. Have you considered also handling the other FS-partitioning-schemes issues that are mentioned in the linked issue, or one of #1043 or #1041, also in the same general area? (This is me encouraging you to do more, more, more!)

I just released 3.2.3, and I'd like to get 3.2.4 out the door shortly; I might not have time to look at this very closely just now.

src/modules/partition/core/PartitionLayout.h
```diff
@@ -0,0 +1,70 @@
...
+        SizeUnit partMinSizeUnit;
+    };
+
+    PartitionLayout();
+    PartitionLayout( PartitionEntry entry );
+    PartitionLayout( const PartitionLayout& layout );
+    ~PartitionLayout();
+
+    void addEntry( PartitionEntry entry );
+    void addEntry( QString mountPoint, QString size, QString min = "" );
```
adriaandegroot: `min = QString()` is more idiomatic nowadays. Also one typically passes `const QString&`, even though they're implicitly shared and this doesn't make a lot of difference performance-wise.

adriaandegroot: .. I might have some stylistic comments, but overall at-a-quick-glance this looks amazingly well-done.

kkofler: The point of the const reference is to avoid pointless atomic reference count operations.

And `min = QString()` is not just "more idiomatic", but also faster than `min = ""` that explicitly constructs a non-null empty QString.

a-wai: OK, will fix it, thanks for the explanations.
(my C++ is a bit rusty, I've been doing C-only work for the last 10 years or so)

I'll leave you some time to review the whole thing and will submit a batch of fixes to take everything into account, if you're fine with it.

a-wai:
@adriaandegroot This PR should fix #725 & #1028 too, I believe my solution is generic enough to be used in these cases. Regarding the BtrFS-related issues (such as #525 ), I lack hands-on experience with the filesystem, but it is an improvement I could work on in the future.

#1041 & #1043 look like relatively easy fixes, I should be able to address them :smiley:

abucodonosor:
Very interesting idea.

Maybe some prepartion for https://github.com/calamares/calamares/issues/847 
could be made here too ?

PR #1075, opened 2019-01-23
Add a raw filesystem copy module

a-wai:
In some cases, we might want to copy a filesystem as if we were using a
simple 'dd' command, in order to create an exact copy, down to the block
level.
This can be useful in particular when working with dm-verity for
checking the rootfs integrity: that way, we can make a direct copy of
the rootfs and its verity partition and keep the system usable.

This patch adds a new 'rawfs' module to calamares, making possible to
block-copy a filesystem to a block device.

### DIFF
--- a/src/modules/rawfs/main.py
+++ b/src/modules/rawfs/main.py
@@ -1,0 +1,184 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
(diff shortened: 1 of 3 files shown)


a-wai:
@philmmanjaro I started with the `dummypython` module and therefore kept the authors list, but I get how this can be confusing, I can change it.

PR #1077, opened 2019-02-01
Fix partition layout sizes parsing

a-wai:
This PR fixes 2 bugs in partition size parsing, both related to the custom partition layout additions.


philmmanjaro:
code wise it LGTM

PR #1078, opened 2019-02-01
[rawfs] Allow use of symlinks in configuration

a-wai:
In its current state, the 'rawfs' module requires the source partition 
to be identified either by its mount point or device name, but using a 
symlink to either one (e.g '/dev/disk/by-uuid/...') would fail.

This patch fetches the real path of source partition, allowing the use 
of symlinks.


philmmanjaro:
LGTM

PR #1094, opened 2019-02-22
Fix default fs

### LINKED ISSUE #1093: defaultFileSystemType is ignored
**Describe the bug**

After https://github.com/calamares/calamares/pull/1068 , defaultFileSystemType is ignored.

Old code inserted defaultFsType while new code will always insert Ext4 when custom layout
is not used.

https://github.com/calamares/calamares/blob/master/src/modules/partition/core/PartitionLayout.cpp#L118

**To Reproduce**
( Distribution does not matter here )
1. Use any calamares version / calamares from git with merged code from https://github.com/calamares/calamares/pull/1068
2. Set defaultFileSystemType to something else than ext4 in partition.conf ( xfs as example )
3. Be sure you don't have custom layout enabled
4. Start the installation and use Erase as example and you should notice the following:

....
.. Partition-module setting *defaultFileSystemType* "xfs"
..

and a bit later:

     mtpoint: :   "/" 
     fs: :   "ext4" 
     fsname :   "ext4" 
     uuid :   "" 

...

**Expected behavior**

Working defaultFileSystemType with or without custom layouts enabled.

a-wai:
When using the default partition layout (only a `/` partition), the
filesystem used was ext4, ignoring the `defaultFileSystemType`
configuration option.

This commit fixes this bug, so that any supported filesystem can now be
used for the default partitioning scheme.

Fixes #1093

### DIFF
--- a/src/modules/partition/core/PartitionLayout.cpp
+++ b/src/modules/partition/core/PartitionLayout.cpp
@@ -18,6 +18,9 @@
  *   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "GlobalStorage.h"
+#include "JobQueue.h"
+
 #include "core/PartitionLayout.h"
 
 #include "core/KPMHelpers.h"
@@ -28,17 +31,32 @@
 #include <kpmcore/core/partition.h>
 #include <kpmcore/fs/filesystem.h>
 
+static int
+getDefaultFileSystemType()
+{
+    Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
+    int defaultFs = FileSystem::typeForName( gs->value( "defaultFileSystemType" ).toString() );
+
+    if ( defaultFs == FileSystem::Unknown )
+        defaultFs = FileSystem::Ext4;
+
+    return defaultFs;
+}
+
 PartitionLayout::PartitionLayout()
 {
+    defaultFsType = getDefaultFileSystemType();
 }
 
 PartitionLayout::PartitionLayout( PartitionLayout::PartitionEntry entry )
 {
+    defaultFsType = getDefaultFileSystemType();
     partLayout.append( entry );
 }
 
 PartitionLayout::PartitionLayout( const PartitionLayout& layout )
     : partLayout( layout.partLayout )
+    , defaultFsType( layout.defaultFsType )
 {
 }
 
@@ -115,7 +133,7 @@ PartitionLayout::addEntry( const QString& mountPoint, const QString& size, const
     PartitionLayout::PartitionEntry entry( size, min );
 
     entry.partMountPoint = mountPoint;
-    entry.partFileSystem = FileSystem::Ext4;
+    entry.partFileSystem = defaultFsType;
 
     partLayout.append( entry );
 }
@@ -128,6 +146,8 @@ PartitionLayout::addEntry( const QString& label, const QString& mountPoint, cons
     entry.partLabel = label;
     entry.partMountPoint = mountPoint;
     entry.partFileSystem = FileSystem::typeForName( fs );
+    if ( entry.partFileSystem == FileSystem::Unknown )
+        entry.partFileSystem = defaultFsType;
 
     partLayout.append( entry );
 }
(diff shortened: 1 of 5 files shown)


philmmanjaro:
Lgtm




Kind regards


Philip Müller


Sent by mobile 







On Fri, Feb 22, 2019 at 1:14 PM +0100, "Arnaud Ferraris" <notifications@github.com> wrote:












When using the default partition layout (only a / partition), the

filesystem used was ext4, ignoring the defaultFileSystemType

configuration option.


This commit fixes this bug, so that any supported filesystem can now be

used for the default partitioning scheme.


Fixes #1093



You can view, comment on, or merge this pull request online at:


  https://github.com/calamares/calamares/pull/1094

Commit Summary

  [partition] use configured default filesystem type instead of ext4
  [partition] fix naming of PartitionLayout class member variables


File Changes

  
    M
    src/modules/partition/core/PartitionLayout.cpp
    (34)
  
  
    M
    src/modules/partition/core/PartitionLayout.h
    (3)
  


Patch Links:

  https://github.com/calamares/calamares/pull/1094.patch
  https://github.com/calamares/calamares/pull/1094.diff

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -28,17 +31,32 @@
 #include <kpmcore/core/partition.h>
 #include <kpmcore/fs/filesystem.h>
 
+static int
```
adriaandegroot: That looks poorly-typed to me: you're going from FileSystem::Type to int. It should be the enum type, and clients somewhere else can cast it to int if needed (besides, this will break if FileSystem::Type changes to an enum class).

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -28,17 +31,32 @@
...
+getDefaultFileSystemType()
+{
+    Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
+    int defaultFs = FileSystem::typeForName( gs->value( "defaultFileSystemType" ).toString() );
```
adriaandegroot: This looks kind of fragile: it's dependent on that key being already-inserted into the GlobalStorage, It **is**, in the current code-setup, by PartitionViewStep::setConfigurationMap(), but ..

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -128,6 +146,8 @@ PartitionLayout::addEntry( const QString& label, const QString& mountPoint, cons
     entry.partLabel = label;
     entry.partMountPoint = mountPoint;
     entry.partFileSystem = FileSystem::typeForName( fs );
```
adriaandegroot: All the calls to FileSystem::typeForName() now use **localised** names; that means that sometimes you write "btrfs" and sometimes "Btrfs" depending on the language the user has selected. You can pass a localisation in, basically QStringList( {"C"} ). See findFS() in PartitionViewStep.cpp .. this also applies to other calls to typeForName(). (The function findFS() does some work to make sure it ends up with a canonical name -- maybe we should also just put the enum value into globalstorage, then you don't need to do any re-parsing of the name here)

src/modules/partition/core/PartitionLayout.cpp
a-wai: Right, I'll check it and fall back to ext4 if the key is missing

src/modules/partition/core/PartitionLayout.cpp
a-wai: I'm pretty sure I had a good reason to use int back then. As I can't even remotely remember why, I guess it wasn't so good, after all...

src/modules/partition/core/PartitionLayout.cpp
a-wai: Oh OK, thanks for the tip ! I'm gonna export findFS into PartUtils then, and build on top of it.
Putting the enum value into globalstorage would be helpful for seeking the default filesystem, but I have to use findFS for every partition anyway, so, once more shouldn't harm (plus it would a one-shot use case, not necessarily worth it).

abucodonosor:
Code is working fine for me now.

Tested installs with defaultFileSystemType set to xfs and btrfs and all worked fine.

PR #1117, opened 2019-04-08
[partition] Add max size parameter

a-wai:
When using a custom partition layout with partition sizes in %, it can
be useful to set an upper limit to the partition size.

For instance, using a 20% size for the `/` partition will create a 24GB
partition on a 120GB drive, but a 200GB partition on a 1TB drive, which
is not useful, and could be avoided by setting a maximum partition size.

This commit adds the `maxSize` parameter (with a default value of 100%).

### DIFF
--- a/src/modules/partition/core/PartitionCoreModule.cpp
+++ b/src/modules/partition/core/PartitionCoreModule.cpp
@@ -774,6 +774,7 @@ PartitionCoreModule::initLayout( const QVariantList& config )
 {
     QString sizeString;
     QString minSizeString;
+    QString maxSizeString;
 
     m_partLayout = new PartitionLayout();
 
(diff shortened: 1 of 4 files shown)


abucodonosor:
@a-wai 

Thx for the work on this :)

Maybe I am missing something but 100%  seems wrong for me by default.
That means if you do not set max size it eats all the disk ?:)

For me ,the logic would be when max size is missing to default to 'size' case that 
is now max size for that partition..

Also I think you need add one more check in case min size = max size ( yes stupid
thing to do but users do all kind strange things all the time ) 

Thx

a-wai:
Thanks for your review :)

> Maybe I am missing something but 100% seems wrong for me by default.
> That means if you do not set max size it eats all the disk ?:)

No, the *reference* parameter remains `size`, `maxSize` is only used when it's smaller than `size` (once converted to a disk sectors count in `PartitionLayout::execute`).
This comes down to the logic you describe, only the implementation details differ ;)

> Also I think you need add one more check in case min size = max size ( yes stupid
> thing to do but users do all kind strange things all the time )

I don't think this would be a problem: if both values are equal, the partition would first be shrunk to `minSize`, then the check against `maxSize` will pass.
However, I should indeed add a check in case `minSize` > `maxSize` (really stupid, but if one is expressed in % and the other one in GB for example, this could easily happen).

Come to think of it, I should also add a hard check on mandatory parameters (name, size, fs type & mountpoint) which would throw an error if any of them is missing, this would be safer.

PR #1119, opened 2019-04-10
[partition] Fixing LVM scanning according to new kpmcore API

cjlcarvalho:
To scan LVM devices, it was needed to call LvmDevice::scanDevices from kpmcore before. However, there is a change in this API, as we need to force the scanning of the other types of volume groups before scanning LVM, because it needs to check for PVs in other devices and some clients may call this API without considering these other types of volume groups. This scanning process now is done using VolumeManagerDevice::scanDevices to force the scanning of all the types of volumes groups before LVM.

### DIFF
--- a/src/modules/partition/core/PartitionCoreModule.cpp
+++ b/src/modules/partition/core/PartitionCoreModule.cpp
@@ -55,6 +55,7 @@
 #include <kpmcore/core/device.h>
 #include <kpmcore/core/lvmdevice.h>
 #include <kpmcore/core/partition.h>
+#include <kpmcore/core/volumemanagerdevice.h>
 #include <kpmcore/backend/corebackend.h>
 #include <kpmcore/backend/corebackendmanager.h>
 #include <kpmcore/fs/filesystemfactory.h>
@@ -686,12 +687,13 @@ PartitionCoreModule::scanForLVMPVs()
         }
     }
 
-    // Update LVM::pvList
-    LvmDevice::scanSystemLVM( physicalDevices );
-
 #ifdef WITH_KPMCOREGT33
+    VolumeManagerDevice::scanDevices( physicalDevices );
+
     for ( auto p : LVM::pvList::list() )
 #else
+    LvmDevice::scanSystemLVM( physicalDevices );
+
     for ( auto p : LVM::pvList )
 #endif
     {
--- a/src/modules/partition/core/PartitionCoreModule.cpp
+++ b/src/modules/partition/core/PartitionCoreModule.cpp
@@ -551,26 +551,22 @@ PartitionCoreModule::lvmPVs() const
 bool
 PartitionCoreModule::hasVGwithThisName( const QString& name ) const
 {
-    for ( DeviceInfo* d : m_deviceInfos )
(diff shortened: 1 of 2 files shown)


abucodonosor:
@cjlcarvalho 

That API change is for 3.3.x ?

cjlcarvalho:
@abucodonosor Yes.

stikonas:
> @cjlcarvalho
> 
> That API change is for 3.3.x ?

Well, to be precise, API changed between 3.3.x and 4.0

src/modules/partition/core/PartitionCoreModule.cpp
```diff
@@ -686,12 +683,13 @@ PartitionCoreModule::scanForLVMPVs()
         }
     }
 
-    // Update LVM::pvList
-    LvmDevice::scanSystemLVM( physicalDevices );
-
 #ifdef WITH_KPMCOREGT33
```
stikonas: Can you fix this ifdef so that it would work properly with kpmcore 3.3.1 (i.e. old API for kpmcore 3.3.1 but new for kpmcore 4)

adriaandegroot:
`-DWITH_KPMCOREGT33` means "KPMCore **greater than** 3.3.0", which has been master and/or whatever KDE neon ships for some time now. Since (for instance) `softwareraid.h` was added to master after 3.30 it is only used when something newer than 3.3.0 is found.

If you release a KPMCore 3.3.1, is that 3.3.0-API or master-API? Because that would have a real impact on the Calamares code (er .. a minimal fix would be to change the CMake check that sets KPMCOREGT33, a better fix would change the name of the define).

stikonas:
Kpmcore 3.3.1 would have the same API/ABI as kpmcore 3.3.0, I think whoever added that ifdef (could be me) made a poor assumption that >3.3.0 means master.

Well, what needs to be out in released version is fix for luks1 forcing when creating luks partitions.

By the way, kpmcore 4 is ready for release too provided that we solve this issue. I don't want to release something that does not work with Calamares.

On 11 April 2019 14:37:39 BST, Adriaan de Groot <notifications@github.com> wrote:
>`-DWITH_KPMCOREGT33` means "KPMCore **greater than** 3.3.0", which has
>been master and/or whatever KDE neon ships for some time now. Since
>(for instance) `softwareraid.h` was added to master after 3.30 it is
>only used when something newer than 3.3.0 is found.
>
>If you release a KPMCore 3.3.1, is that 3.3.0-API or master-API?
>Because that would have a real impact on the Calamares code (er .. a
>minimal fix would be to change the CMake check that sets KPMCOREGT33, a
>better fix would change the name of the define).
>
>-- 
>You are receiving this because you commented.
>https://github.com/calamares/calamares/pull/1119#issuecomment-482117365

-- 

stikonas:
@cjlcarvalho so @adriaandegroot  suggested to rename ifdef to `KPMCORE4API`

adriaandegroot:
Fixes to handle the various API versions are incoming in master, so we'll be ok then.

cjlcarvalho:
@stikonas Ok. Sorry for the delay. I am in the end of the semester now and I was a little busy. Thanks for merging, @adriaandegroot :)

PR #1123, opened 2019-04-15
Better LightDM Autologin Support

feren:
This change fixes a few issues and adds a few improvements to the LightDM Autologin configuration process:

Fixes:

- Fixes malforming of configuration file out of the box, as without `[SeatDefaults]`, `[Seat:*]` or similar in the configuration file LightDM will break on a lot of distributions using LightDM

- Preserves the intended lightdm.conf file settings outside of `autologin-user` if the distribution has an /etc/lightdm/lightdm.conf file of its own

Misc. changes:

- Small spelling fix


src/modules/displaymanager/main.py
```diff
@@ -487,31 +487,45 @@ def set_autologin(self, username, do_autologin, default_desktop_environment):
...
+        addseat = False
+        loopcount = 0
 
         if os.path.exists(lightdm_conf_path):
             with open(lightdm_conf_path, 'r') as lightdm_conf:
                 text = lightdm_conf.readlines()
+                # Check to make sure [SeatDefaults] or [Seat:*] is in the config,
+                # otherwise we'll risk malforming the config
+                if not '[SeatDefaults]' in text and not '[Seat:*]' in text:
+                    addseat = True
```
adriaandegroot: `addseat = '[SeatDefaults]' not in text and  '[Seat:*]' not in text` is shorter and more typically Pythonic

src/modules/displaymanager/main.py
```diff
@@ -487,31 +487,45 @@ def set_autologin(self, username, do_autologin, default_desktop_environment):
...
 
             with open(lightdm_conf_path, 'w') as lightdm_conf:
+                if addseat:
+                    # Append Seat line to start of file rather than leaving it without one
```
adriaandegroot: That's *prepend* Seat line to file.

src/modules/displaymanager/main.py
```diff
@@ -487,31 +487,45 @@ def set_autologin(self, username, do_autologin, default_desktop_environment):
...
+                    text = ["[Seat:*]\n"] + text
+                loopcount = 0
                 for line in text:
                     if 'autologin-user=' in line:
                         if do_autologin:
                             line = "autologin-user={!s}\n".format(username)
                         else:
                             line = "#autologin-user=\n"
+                        text[loopcount] = line
+                    loopcount += 1
```
adriaandegroot: Not sure why you're replacing the loop which writes line-by-line by one that updates the *text* array, then writes it in one go? That seems like a superfluous change.

feren: Originally I thought the loop was overwriting the file each time with the current line it was looped to, hence I came up with that loop, but now the established loop technique seems to be working for me as it was designed to so I restored it in the latest commit.

src/modules/displaymanager/main.py
feren: Fixed in the next commit.

src/modules/displaymanager/main.py
feren: Cheers, shortened it to that in the code.

feren:
I'll go ahead and commit in a bit to add myself to the credits files you mentioned in the previous comment, and then the pull should be good-to-merge hopefully.

feren:
Alright, done. I added '(feren)' to the CHANGES credit for my name since it otherwise is not so easy to associate my name with my GH account due to not having set that as my name yet... is that ok or should I remove the brackets or just change my name in there to what it appears to be on my GitHub Profile, being 'The Feren OS Dev'?

adriaandegroot:
Merged, thanks!

Issue #1122, opened 2019-04-15
Add option to disable cancel ONLY when the installation is in progress


feren:
**Is your feature request related to a problem? Please describe.**
It's kind of odd that Calamares lets you Cancel out of the installation procedure while the OS is installing, especially given that the damage of having begun the installation procedure is already done by the time you try and cancel out of installing the OS at that moment as a person who isn't aware that cancelling it won't undo the changes that have already happened.

**Describe the solution you'd like**
The option to disable cancel ONLY when in the 'Install' phase would be more than appreciated. This is not to be confused with the Pull Request for disabling cancel globally in the installer, however.

**Describe alternatives you've considered**
I've considered making use of the global cancel disabler, but to be honest globally disabling cancel for the purpose of preventing freak accidents from prematurely finishing the installation phase by cancelling it part-way through the phase is a little overkill.

**Additional context**
Pretty much just how Ubiquity and such do it where the effects of disabling cancel only hit Calamares once the Installation process has begun for real and then get enabled again at the end for the post-Install screens (usually just the 'finish' module)...

abucodonosor:
maybe with some option .. 

I don't see why force that on all users while some peoples may 
not know what they are doing or what is means to cancel an installation.

Maybe one changes his mind , want to install something different and just cancel,
or realized he forgot something or what do it different .. So perfectly fine to 'cancel'
in such a cases ..

While I get your point , I don't think we should force everyone by defalt to wait the installation finished to 'cancel' , for whatever reason the user may have doing that.

Thx

feren:
It'd be nice to have it as an option to do so, ala how the Cancel Button Global Disabling now works to be honest.

kkofler:
One use case where it makes sense to cancel a running installation is if you realize that you are about to format (or already formatting) the wrong partition. The sooner you cancel, the higher your chances are to recover at least **some** of the data. If you have to wait until the end of the installation, it will be all gone for sure!

abucodonosor:
@feren 

sure but there is still reason to cancel, like pointed out from @kkofler.

Maybe something like this will do it:

https://crazy.dev.frugalware.org/maybe-something-like-this.patch

adriaandegroot:
That patch looks pretty good, actually; I'd call it something else, like *no-cancel-during-exec*, and it needs double-checking that when you come **out** of exec, the cancel is restored (put an extra view step after the first exec block in sequence to try it, eg. a dummypythonqt or some such). But that'd do the job for @feren

[...]

PR #1126, opened 2019-04-16
Patch to add 'dont-cancel' setting to disable Cancel only during the installation phase

feren:
Implementation of #1122 using the patch by @abucodonosor

### DIFF
--- a/src/libcalamares/Settings.cpp
+++ b/src/libcalamares/Settings.cpp
@@ -185,6 +185,7 @@ Settings::Settings( const QString& settingsFilePath,
     , m_doChroot( true )
     , m_promptInstall( false )
     , m_disableCancel( false )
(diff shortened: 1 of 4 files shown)


src/libcalamares/Settings.cpp
```diff
@@ -1,6 +1,7 @@
 /* === This file is part of Calamares - <https://github.com/calamares> ===
  *
  *   Copyright 2019, Dominic Hayes <ferenosdev@outlook.com>
+ *   Copyright 2019, abucodonosor
```
abucodonosor: if you add me there then : Gabriel Craciunescu <crazy@frugalware.org> :)

adriaandegroot:
Merged, thanks! I'll polish it a little for the 3.2.6 release later this week.

PR #1127, opened 2019-04-18
Add various checks for partition layout configuration

a-wai:
In order to avoid undefined or strange behavior when the partition layout configuration is misformed, this PR adds various checks regarding each partition's mandatory fields, their sizes and consistency of sizes (e.g. we don't want max size > min size).
It aims at implementing a consistent and reliable default behavior if an error is encountered.

This PR also introduces a new class for parsing partition sizes, based on `NamedSuffix`, which replaces the functions previously used. This should help maintain consistency across various modules, and might possibly be moved to libcalamares with a bit more work in the future, replacing a similar class in the `resizefs` module.


adriaandegroot:
Thanks! (Moving to libcalamares is, indeed, a good next step, dunno if that fits your schedule in the next week-and-a-half).

PR #1144, opened 2019-05-08
Factor partition size classes in libcalamares

a-wai:
As there are similar classes, used to store partition sizes, in both the `partition` and `fsresizer` modules, this patch series introduces a generic PartitionSize class in libcalamares, replacing the previous module-specific implementations.


a-wai:
@adriaandegroot I'm not sure putting `PartitionSize` in the `Calamares` namespace is the best option here, the decision is up to you as you have better knowledge of the global architecture and what you expect for the future.
If you prefer going for a new namespace (`CalamaresPartition` ?), or any other option, please let me know and I'll change it, first thing in the morning :wink:

adriaandegroot:
I'm good with namespace CalamaresUtils for now. THis is an important first step in sharing code betweeen modules that have similar needs **and** a first step in having libcalamares provide more "services" in general to C++ modules -- so let's get it done first, and make it perfect later.

PR #1177, opened 2019-06-17
make directory before copying

apachelogger:
otherwise clean build dirs fail because copying into a nonexistent target
isn't supported.


apachelogger:
Oh lol, @adriaandegroot got to it first. Closing

PR #1181, opened 2019-06-22
Enhancement of initcpio config supporting cases where squashfs image doesn't contain kernel image

mipimipi:
Hi,
the initcpio module doesn't work for ISO's where the squashfs image doesn't contain a kernel image (that's the case, for example, for archiso).
I propose to enhance the module by some new  optional config parameters (and the corresponding logic in the main.py). With that, the module also works for the ISO's mentioned above. Since the added config is completely optional, users don't have to change their existing config.
With the new config, users can set a source and a target for the kernel image. The image is then copied from the source to the target.
mipi

### DIFF
--- a/src/modules/initcpio/initcpio.conf
+++ b/src/modules/initcpio/initcpio.conf
@@ -1,3 +1,4 @@
 # Run mkinitcpio(8) with the given preset value
 ---
-kernel: linux312
+kernel: linux
+image_source: /run/archiso/bootmnt/arch/boot/x86_64/vmlinuz
--- a/src/modules/initcpio/initcpio.conf
+++ b/src/modules/initcpio/initcpio.conf
@@ -1,4 +1,22 @@
 # Run mkinitcpio(8) with the given preset value
 ---
 kernel: linux
-image_source: /run/archiso/bootmnt/arch/boot/x86_64/vmlinuz
+
+# For some ISO's (e.g. with archiso generated ISO's) the squashfs image doesn't
+# contain a kernel image file. Therefore, the kernel image file needs to be
+# copied to the corresponding target. For this, the source and the target path
+# can be specified with the configuration parameters image_source and
+# image_target.
+#
+# Both parameters are optional. If image_source is not set, the system assumes
+# that the squashfs image already contains a kernel image file, and image_target
+# is ignored (if it is set).
+# If image_source is set, the kernel image is copied to the target specified
+# with image_target. If image_target is not set, /boot/vmlinuz-linux is
+# assumed to be the target path of the image. The image_target needs to be
+# consistent with the kernel image path in the mkinitcpio preset.
+#
+# For ISO's generated with archiso, the following configuration is needed:
+#
+# image_source: /run/archiso/bootmnt/arch/boot/x86_64/vmlinuz
+
--- a/src/modules/initcpio/initcpio.conf
+++ b/src/modules/initcpio/initcpio.conf
@@ -1,6 +1,6 @@
 # Run mkinitcpio(8) with the given preset value
 ---
-kernel: linux
+kernel: linux312
 
 # For some ISO's (e.g. with archiso generated ISO's) the squashfs image doesn't
 # contain a kernel image file. Therefore, the kernel image file needs to be
--- a/src/modules/initcpio/main.py
+++ b/src/modules/initcpio/main.py
@@ -22,8 +22,6 @@
 
 import gettext
 import libcalamares
-from libcalamares.job import configuration
-from lincalamares.globalstorage import value
 from subprocess import CalledProcessError
 from subprocess import check_call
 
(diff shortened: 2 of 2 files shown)


adriaandegroot:
Can you squash this down to fewer commits? In particular, also commits with meaningful commit messages? Please follow https://github.com/calamares/calamares/blob/master/ci/HACKING.md#commit-messages . (I can do this for you if you don't know how, but that would at least **partly** break your fork).

Have you added documentation of the new optional settings (in `initcpio.conf`)? I can't spot that quickly in these commits.

adriaandegroot:
(er .. **before** you start on this) .. what is the actual problem? Is your target filesystem not being set up properly to place a kernel into the target system? If so, isn't an addition to the `unpackfs` module a more likely place to do this? I can't see if there is **more** to this patch than copying a kernel file into the target at this point, but if there isn't the solution probably lies elsewhere.

Feel free to drop in on IRC (#calamares on Freenode, often most-active during European evenings).

mipimipi:
I can do the squash - no problem. And yes, there is documentation about the new initcpio options.
What is the problem? For archiso, for example, the squashfs does not contain a kernel image. I.e., the kernel image that is needed by initcpio must be copied from wherever it is located originally to the place where initcpio expects it - and you're completely right, it's just a simple file copy. As I found out in the WWW, people are solving this currently by modifying initcpio in a way that is very specific to there ISO (I found examples for archiso). I thought that Calamares should offer a more generic solution for that which doesn't require the users to do code changes but which only requires configuration. For sure this could be done in unpackfs. In this case I would need some guidance  where to place the file copy command and where in the config to add the necessary config options.
IRC channel: I was already there with 2 other questions, but the experience was not that great, I have to admit.

adriaandegroot:
Ok, so let's back alllll the way up to the original problem:

 - initcpio module fails because the required kernel isn't in the target system

The obvious solution to that is

 - put the required kernel into the target system

There are a couple of different ways of getting to that obvious solution:

 - put the kernel into a squashfs or ext4 fs that can be copied to the target
 - copy the file separately into the target
 - copy the file as part of the initcpio module

The first comes down to "fix your squashfs", which **may** imply "fix your tooling", so if that's impractical, let's look at the other two. Adding specialized configuration and code for copying a file in the initcpio module itself is tempting, like in this patch. Apparently people are doing this in various places (and not upstreaming it, or telling us about the problem, so for that at the very least I'm grateful to @mipimipi ). But then, hypothetically, we get another module that needs a file copied, and it gets its own code and config, and then another .. there's then a lot of code going on that does the same thing.

So I'd really like to look at the more general approach, which is "copy the file separately". (Oh, **do** make sure you put that module in the sequence before initcpio). There's three ways we could go about that:

 - process job which runs "cp" (this doesn't have good support for finding the target mount point, though).
 - shellprocess job (e.g. set the job to be `cp /path/to/live-iso/kernel @@ROOT@@/boot/kernel`).
 - unpackfs job

That last module **is** responsible for filling up target filesystems, so I would call that, of the three, the most natural fit for copy-a-file-into-the-target. The shellprocess job is a natural second, since it's configurable, designed to handle the mountpoints nicely, and the recommended way to do "randome shell commands during installation".

Still, let's look at the unpack module, since that's the populate-target-filesystems Swiss-army-knife.

adriaandegroot:
(comment part 2) The unpackfs module configuration is a list of sources, with a type and a destination. Types "squashfs" and "ext4" are generally supported, but anything you can mount will do.

I would propose adding the type "file" there. Maybe also "tar". Then we can use those to drive non-filesystem-related file copies to the target system, and you have one place showing what goes into the target and in what order.

Unfortunately, the unpackfs module is a mess.

adriaandegroot:
So, to sum up the summary: I'm not **against** this, and the overall code is nice, but I think it would be better in a different spot, so as to be more general.

mipimipi:
I think, adding the copy activity to unpackfs is a very good idea. I could do this, but since I am currently working on some other projects, it'll take a few months until I will be ready with that. If it's OK for you, I will send a new PR once I am done with the unpackfs approach.

erikdubois:
Just discovered this topic and it is exactly what we do in ArcoLinux (based on Arch Linux) to get our kernel into the installed system.

We change the initcpio our selves
https://github.com/arcolinux/arcolinux-pkgbuild-3party/blob/master/calamares-3.2.10-3/initcpio-arcolinux/main.py

line 40

Would be an improvement if we could have a more general approach for all Linux systems.

arcolinuxz:
Am I correct in saying that Calamares 3.2.11 does not support the copying over of the vmlinux file any more. We used to have a main.py file in initcpio module on 3.2.10. I do not see it anymore.
What is today the proper way (you think) to get this file over again? 

https://github.com/arcolinux/arcolinux-pkgbuild-3party/blob/master/calamares-3.2.10-3/initcpio-arcolinux/main.py

kkofler:
The `initcpio` module has been ported from Python to C++ (as part of the security fixes), so you have to modify the C++ code of the module now.

adriaandegroot:
See #1188 .. copying a file is an extension that some people hacked into the initcpio module; as you can see from the discussion  that functionality really belongs elsewhere, so it'll be added to unpackfs.

arcolinuxz:
When can we test this functionality?

adriaandegroot:
The *unpackfs* module has been extended with a special *sourcefs* value of *file*. Using that, you can copy a single file (or a directory) into the target system directly from the source system. That means you can copy the kernel in with a suitable configuration there, rather than patching the initcpio module.

Issue #1182, opened 2019-06-25
Feature request : button to send calamares log file to pastebin.com or other services


arcolinuxz:
During conversations on discord we discovered the need in the ArcoLinuxD iso (which does not contain any browser at all) to have the ability to put logs online in order to investigate issues with calamares.

I came up with the application pastebinit on Arch Linux as a solution to put documents online on pastebin.com.

Then it occurred to us that having a way to share issues/information about calamares in Calamares could be in everyone's interest.

What if  there was a button to quickly send over all the data of the installation?
What if you would have a battery of logs to analyze issues?
How many times does this error occurr etc. A way to prioritize.
Just brainstorming out loud.

Keep at it people.

adriaandegroot:
That's an intriguing idea. Long ago (?) I think Calamares had a log-upload-on-crash, which went to calamares.io. That went away with the switch to KCrash and when the heavyweight calamares.io site was phased out.

A common (?) XML-RPC interface for pastebins is documented at http://paste.debian.net/rpc-interface.html .. so an implementation of that in C++ is possible. OTOH, `pastebinit` is **also** common, supports lots of bins and is already widely configurable, so calling that as an external process is probably much, much easier. Main issue then is getting the pastebin URL back to the user so they can report it.

kkofler:
I guess a C++/Qt client library for paste services would be a candidate for KDE Frameworks. I can imagine, e.g., DrKonqi getting support for it.

adriaandegroot:
KF5Purpose **might** be the thing to do this; it's hard to tell (it has a generic HTTPPost job).

But I'm told that @bill-auger already has this , so I'll just wait from input from that side.

bill-auger:
yes, its dead simple with termbin.com - and that has the extra
advantage of the back-end (fiche) being freely licensed and
easy to self host, if one wants the posts to be private, for example

arcolinuxz:
I had an idea you might incorporate easily in the debug part of calamares.
In one of the tabs at the end you have buttons. An extra button to send the log text to pastebin.com?
I use **pastebinit** to send code of installations. Maybe this can be implemented.
The url can be written down and later you can make an issue about it.

JoelStienlet:
Is there also support planned for connections through "socks" proxy servers? I'm pretty sure no liveCD has proxychains4 installed by default.

bill-auger:
JoelStienlet - if that is unrelated to paste uploads, it would be best to open a specific ticket about it

[...]

PR #1186, opened 2019-06-28
implement post log to paste server

bill-auger:
this addresses feature request #1182

this commit adds a button to the failure MessageBox which posts the install log to termbin.com - this approach has the extra advantage of the back-end (fiche) being freely licensed and easy to self host, if one wants the posts to be private, for example

### DIFF
--- a/src/libcalamaresui/ViewManager.cpp
+++ b/src/libcalamaresui/ViewManager.cpp
@@ -238,65 +238,73 @@ quint16 fichePort = 9999;           // TODO: config var
              [msgBox, ficheHost, fichePort, pasteUrlFmt, pasteUrlTitle] ( QAbstractButton* button )
     {
         if ( button->text() != tr( "&Yes" ) )
-            QApplication::quit();
-        else
         {
-            QFile pasteSourceFile( Logger::logFile() );
-            if ( !pasteSourceFile.open( QIODevice::ReadOnly | QIODevice::Text ) )
-                cError() << "Could not open log file";
-            else
-            {
-                QByteArray pasteData;
-                while ( !pasteSourceFile.atEnd() )
-                {
-                    pasteData += pasteSourceFile.readLine();
-                }
-
-                QTcpSocket* socket = new QTcpSocket(msgBox);
-                socket->connectToHost( ficheHost, fichePort );
-
-                if ( !socket->waitForConnected() )
-                    cError() << "Could not connect to paste server";
-                else
-                {
-                    cDebug() << "Connected to paste server";
-
-                    socket->write( pasteData );
-
-                    if ( !socket->waitForBytesWritten() )
-                        cError() << "Could not write to paste server";
-                    else
-                    {
-                        cDebug() << "Paste data written to paste server";
-
-                        if ( !socket->waitForReadyRead() )
-                            cError() << "No data from paste server";
-                        else
-                        {
-                            cDebug() << "Reading response from paste server";
-
-                            char resp[1024];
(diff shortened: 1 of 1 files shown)


src/libcalamaresui/ViewManager.cpp
```diff
@@ -189,28 +192,106 @@ ViewManager::insertViewStep( int before, ViewStep* step )
 void
 ViewManager::onInstallationFailed( const QString& message, const QString& details )
 {
+bool    shouldOfferWebPaste = true; // TODO: config var
```
abucodonosor: Btw at least that TODO need be implemented.

Without that we force that on all the time.

bill-auger:
yes, this was offered for discussion - it includes some naught
bits in its current form that should not be merged

it is also not as optimal as it could be - it is not my original
implementation - i re-wrote it almost entirely yesterday
specifically for this, to avoid shelling out to netcat; so it is
not well-tested

for examples:

* i havent looked, but the log data may actually be in memory
already (or it could be made to be); so the filesystem access
could be eliminated

* it may be possible to avoid spawning the second QMessageBox*
pasteUrlMsgBox; though it was not obvious how

* with all of the error checking, the code style is violated due
to so many indentation levels; so it may be better to refactor
that lambda into a new method

* in its current form, it segfaults intermittently due to a
MUTEX issue; though i suspect that is due to forcing the method
to fire from the ViewManager constructor, at such an early stage
while the requirements check is running

bill-auger:
yes, this was offered for discussion - it includes some naughty bits, in its current form, that should not be merged

it is also not as optimal as it could be - it is not my original implementation - i re-wrote it almost entirely yesterday specifically for this, to avoid shelling out to netcat; so it is not well-tested

examples for discussion:

* it inttroduces several new strings that need translations:
    * QString pasteMsg = tr( "Would you like to paste the install log to the web?" );
    * QString pasteUrlFmt = tr( "Install log posted to: %1" );
    * QString pasteUrlTitle = tr( "Install Log Paste URL" );
    * tr( "&Yes" ) )

* i havent looked, but the log data may actually be in memory already (or it could be made to be); so the filesystem access could be eliminated

* it may be possible to avoid spawning the second QMessageBox, for presenting the paste URL, by preventing the existing one from closing; though it was not obvious how to - that may entail overriding the 'onClicked' handler in a custom widget

* one error check was omitted - socket->readLine() returns the number of bytes read - a result of 0 or -1 indicates that nothing was read; in which case it should probably indicate the failure, rather than presenting the "Install log posted to: %1" message - also, it should return a valid URL; so that could be validated - (UPDATE: i added the data length check)

* it is not obvious if the "new QTcpSocket" and/or the "new QMessageBox" should be deleted explicitly; though the existing QMessageBox is not - perhaps some magic handles that; but of course, the program exits immediately afterward

* with all of the error checking, the code style is violated due to so many indentation levels; so it may be better to refactor that lambda into a new method (UPDATE: i just re-worked this - it is much more readable with early returns; though im not sure if that is acceptable style)

* in its current form, it segfaults intermittently due to a MUTEX issue; though i suspect that is due to forcing the method to fire from the ViewManager constructor, at such an early stage while the requirements check is running

bill-auger:
UPDATE:

* added the socket read data length check and a crude URL validation

* unwrapped the nested error checks - it is much more readable with early returns; though im not sure if that is acceptable style)

adriaandegroot:
Thanks, @bill-auger . This one is going to wait for the next release cycle, early august.

Issue #1176, opened 2019-06-16
Improve the sliders timing when the filesystem is copied over to the harddisk


arcolinuxz:
I realize this is a strange request.

We have been getting this response from our users for like months.
And I never dared to ask it since it is not important at all.

So any calamares version applies.

When you install the system on the harddisk  the slider and the percentage do not change for like minutes and people assume it is broken or crashed.

Could you make it so that the slider does not jump from 22% to 50 % but give people the impression that it is still installing by showing more progress in the slider and the percentages - unsquash process
![slider](https://user-images.githubusercontent.com/36679406/59561594-9aafa180-9022-11e9-863d-4bd5bc09134f.png)

adriaandegroot:
Not a strange request at all: the **appearance** of progress is important, even if it slows the actual progress down :) The unpackfs module is supposed to report progress as it happens, so I guess there's something wrong there (might be that it only reports proper progress with one of the source options, though).

d3signr:
It appears the unpackfs module is designed to only update progress every 100 files, maybe this vale could be reduced in order to update the progress more frequently?

adriaandegroot:
**wow**, that's a beautiful distro during installation. Shame Calamares is ugly :( .. there's probably a lot to gain in the desktop-integration and theming part of the story.

Anyway, I see Calamares reporting progress like **mad** during unpack operations. There's about 500k files to read, and progress goes from 0 to 1 (0 to 100%) as it grinds through them, but that only translates into going from 25% to 26% as it chugs along. So the issue is scaling in the overall progress, **or** as @d3signr points out, a per-module progress could help as well.

adriaandegroot:
I hear @bill-auger has something along these lines, so I'm going to park this issue for a bit and wait for a PR.

bill-auger:
yep, ill show it soon - i need to refresh my memory a bit from
about 18 months ago so i can untangle it for a merge request

my solution adds an m_jobWeight instance var and getter to the
Job class, allowing each sub-class to define a weighted share of
the total 100% progress, against which the jobPercent as passed
into the JobThread::emitProgress() handler is compared, and added
to all previous job weights, rather than the current linear
calculation (m_jobIndex / jobPercent)

for any sub-class not implementing the m_jobWeight getter, the
default value will reproduce the original current behavior as
1/n_jobs

the caveat for the implementer, is that it requires the long
running command to be run in its own thread; so that 'emit
progress' can be pumped in fine grained intervals - and of
[...]

PR #1187, opened 2019-06-28
implement definable job weights

bill-auger:
this addresses feature request #1176 

this commit adds a getter method to the Job class,
allowing each sub-class to define a weighted share of
the total 100% progress, and an instance var to the
JobThread class that caches them

the jobPercent as passed into the JobThread::emitProgress()
handler is compared to the defined weight f the current job,
and added to all previous job weights, rather than the current linear
calculation (m_jobIndex / jobPercent)

for any Job sub-class not implementing the getter, the
default value will reproduce the original current behavior as
1/n_jobs

the caveat for the implementer, is that it requires the long
running command to be run in its own thread; so that 'emit
progress' can be pumped in fine grained intervals - and of
course there needs to be some way to determine what is the
current progress from the parent thread, while waiting for
the long-running thread to exit

### DIFF
--- a/src/libcalamares/Job.cpp
+++ b/src/libcalamares/Job.cpp
@@ -97,6 +97,13 @@ Job::~Job()
 {}
 
 
+qreal
+Job::getJobWeight() const
+{
+    return qreal( 1.0 );
+}
+
+
 QString
 Job::prettyDescription() const
 {
--- a/src/libcalamares/Job.h
+++ b/src/libcalamares/Job.h
@@ -83,6 +83,7 @@ public:
     explicit Job( QObject* parent = nullptr );
     virtual ~Job();
 
+    virtual qreal getJobWeight() const;
     virtual QString prettyName() const = 0;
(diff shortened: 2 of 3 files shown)


adriaandegroot:
Thanks, @bill-auger . This one is going to wait for the next release cycle, early august.

(As for threading: many long-running jobs already do something along these lines, such as unpackfs which reports rsync progress)

bill-auger:
yes, i remember some discussion on IRC about the rsync progress,
with one suggestion being to add a new spinner widget while the
rsync process was running - i think this feature would serve
that purpose as well

also, because the Job->prettyStatusMessage() of the current Job
is queried during each firing of JobThread->emitProgress(), in
addition to overriding the new Job->getJobWeight(), Job
subclasses can add a 'statusMsg' QString instance var as the
value of the local Job->prettyStatusMessage() override, allowing
transient details to enhance the numerical percentage in the
existing GUI

PR #1206, opened 2019-07-29
Esp mount options

### LINKED ISSUE #1198: Add possibility to define mount options for the ESP
**Use case:** Distributor can define mount options for the ESP.

**Current behaviour**
Installing Lubuntu 19.10 with Calamares in EFI mode results in a mount entry in /etc/fstab, which looks like:
```
UUID=ABCD-1234    /boot/efi    vfat    defaults    0    2
```

Installing Xubuntu 19.04 produces a mount entry like:
```
UUID=ABCD-1234    /boot/efi    vfat    umask=0077    0    2
```
and Fedora:
```
UUID=ABCD-1234    /boot/efi    vfat    umask=0077,shortname=wnnt    0    2
```

**Possible solution**
Add an optional configuration `efiMountOptions` to _fstab.conf_.
If the the distributor does not define special mount options for the ESP, fall back to the default `mountOptions`.

**Other**
The distributor could define the fstab.conf as follows:
```
mountOptions:
    default: defaults
    vfat: umask=0077
```
This "works" as desired only if the ESP is the only partition with FAT.

apt-ghetto:
This should fix #1198 

The first commit fixes some flake8 errors.
The second commit adds the new configuration option.

### DIFF
--- a/src/modules/fstab/fstab.conf
+++ b/src/modules/fstab/fstab.conf
@@ -12,6 +12,11 @@ mountOptions:
     default: defaults,noatime
     btrfs: defaults,noatime,space_cache,autodefrag
 
+# Mount options to use for the EFI System Partition. If not defined, it
+# takes either the value of *vfat*, or if also missing, *default* from
+# *mountOptions*
+efiMountOptions: umask=0077
+
 # If a filesystem is on an SSD, add the following options. If a specific
 # filesystem is listed here, use those options, otherwise no additional
 # options are set (i.e. there is no *default* like in *mountOptions*).
--- a/src/modules/fstab/main.py
+++ b/src/modules/fstab/main.py
@@ -162,7 +162,6 @@ class FstabGenerator(object):
             return None
 
         mapper_name = partition["luksMapperName"]
-        mount_point = partition["mountPoint"]
         luks_uuid = partition["luksUuid"]
         if not mapper_name or not luks_uuid:
             return None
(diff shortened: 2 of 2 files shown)


src/modules/fstab/fstab.conf
```diff
@@ -12,6 +12,11 @@ mountOptions:
     default: defaults,noatime
     btrfs: defaults,noatime,space_cache,autodefrag
 
+# Mount options to use for the EFI System Partition. If not defined, it
+# takes either the value of *vfat*, or if also missing, *default* from
```
adriaandegroot: This mentions *vfat*, but the code doesn't do anything like that. To match the code:

```
If not defined, it uses the settings for *default*, above.
```

apt-ghetto: If you have an (existing) configuration without the new `efiMountOptions` like:
```
mountOptions:
    default: defaults
    btrfs: defaults,noatime,space_cache,autodefrag
    vfat: umask=0007
ssdExtraMountOptions:
    ext4: discard
    jfs: discard
    xfs: discard
    swap: discard
    btrfs: discard,compress=lzo
crypttabOptions: luks,keyscript=/bin/cat
```
It will then execute [line 305-306](https://github.com/calamares/calamares/pull/1206/files#diff-a321688334e9f490c6444e4d85aadb81R305) with the variable `filesystem=vfat`. 

`vfat: umask=0007` is just an illustrative example. I don't know, if any distributor has defined a `vfat` value.

The comment should reflect this behaviour (English is not my strength).

Another concern I have: I don't know what happens, if a distributor sets an empty value: `efiMountOptions: `. But I guess, that this is very unlikely.

PR #1219, opened 2019-08-12
mount: Make sure extra mounts are mounted right after /

a-wai:
When working with a read-only rootfs, we might need to mount a tmpfs (defined as an 'extra mount') where we'll create the mountpoints for other partitions. Unfortunately, the extra mounts are only mounted after each partition has been successfully, which might not be possible in this case.

The `mount` module is therefore modified so that any extra mount gets mounted right after the rootfs.

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -37,7 +37,7 @@ _ = gettext.translation("calamares-python",
 def pretty_name():
     return _("Mounting partitions.")
 
-def mount_part(root_mount_point, partition, partitions):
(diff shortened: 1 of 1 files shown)


adriaandegroot:
This seems like a surprising change to me -- for existing configurations that need extramounts and might have them set up to be mounted *after* all the mounts.

I like the refactoring you've done .. maybe we can sort the entire list of mounts based on mount-path and then do them in that order instead? It would mean merging the two lists, then calling the mount-one-thing function for each (rather than special-casing when the extramounts are called).

PR #1221, opened 2019-08-12
Add 'base 10' partition size multiples

a-wai:
In order to improve compatibility with third-party software, we could use a way to define partition sizes in 'base 10' multiples, namely kilo-, mega- & giga- (only 'base 2' multiples kibi-, mibi- & gibi- are supported for now).

Additionally, this PR modifies the `bytesToSectors()` function, so that partitions get aligned on sector boundaries, not necessarily on 1MiB blocks.


kkofler:
The alignment thing is a tradeoff between compactness and performance. Some hardware benefits highly from stronger than sector alignment. Modern hard disk drives use 512e sectors, i.e., 4096-byte physical sectors and 512-byte logically emulated sectors, so they benefit from 8-sector alignment. Some SSDs benefit from even higher alignment.

adriaandegroot:
The KB, MB, GB change is fine.

The alignment thing is there for a reason -- which isn't sufficiently documented in the code, I guess. Partitions do have a minimum size (e.g. an ext4 won't work in 16 blocks), and some partitions have surprising alignment requirements -- ISTR that EFI needs to start on a 1MB boundary. 59fea041b601af942833880506822a50ec755407 is suggestive like that.

So I'd prefer to just do the extra names, and discuss use-cases for tiny-partitions separately.

a-wai:
Got it, thanks for the explanations, I'll revert this commit for now.

PR #1226, opened 2019-08-23
[unpackfs] Enable to use "/" as a source

a-wai:
If we don't have/need an image for the rootfs, we might want to
configure the `/` directory as a source for unpackfs. Unfortunately,
this raises an error:
  - unpackfs first creates a temporary directory
  - it then creates a subdirectory for each source, using the source
path's basename
  - when the source is `/`, the basename is an empty string, therefore
the module tries to create an already existing directory

In order to prevent this error, we use the `os.makedirs` function with
parameter `exist_ok=True` instead of `os.mkdir`.

### DIFF
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -208,7 +208,7 @@ class UnpackOperation:
                 imgbasename = os.path.splitext(
                     os.path.basename(entry.source))[0]
                 imgmountdir = os.path.join(source_mount_path, imgbasename)
-                os.mkdir(imgmountdir)
+                os.makedirs(imgmountdir, exist_ok=True)
 
                 self.mount_image(entry, imgmountdir)
 


agaida:
if i understand you right you want to install a real live system - maybe you should consider to mount your / to an path like /tmp/$foo and install from /tmp/$foo - i do this in siduction for several years without problems. (ok, there will be a cleanup needed after installation - one don't want certain files from the live system in the final installation)

adriaandegroot:
It's a special case for live-CD-is-also-exactly-what-you-want-on-the-installed-system

kkofler:
IMHO, rsyncing the running live system is a very bad idea, because there may be files such as databases that are open and that are not necessarily in a consistent state for an external tool such as `rsync`. A reliable live image installer always installs the raw read-only base image without the writable overlay.

a-wai:
@agaida @kkofler you're both right, though I have to admit that simply copying `/` can be quite helpful during development. The release version probably will do things dfferently, but for now it really speeds up my workflow :smiley:

PR #1230, opened 2019-08-28
add esperanto branding translations, fix mis-configured ones

bill-auger:
all of these translations other than english were declared as 'nl'


adriaandegroot:
Didn't know you were proficient in Esperanto, good job.

The others were hand-translated by me, they're not very good translations. And I don't think it's integrated into the Transifex workflow at all, so they don't get regular updates; this was mostly to demonstrate how branding translations would work in general.

Issue #1202, opened 2019-07-27
Help strings in user ui page is too long for the labels


fhdk:
I have noted the labels for help text on the user setup page is too small for the provided text **or** the help text is too long.

Maybe the labels could be extended to fit horizontally to the window size or the help texts could be shortened to fit the label size.

I have noted it on Manjaro installers and have opened an issue there as well.

https://gitlab.manjaro.org/applications/calamares/issues/42

An example for Manjaro installer can be seen in [this forum comment](https://forum.manjaro.org/t/iso-openbox-19-07-11-unstable/94461/21?u=linux-aarhus) on Manjaro Forum.

adriaandegroot:
I'm minded to change them all to (?) icons with a tooltip, rather than putting the text into the page  itself.

PR #1235, opened 2019-08-28
add checkbox to disable password validations

bill-auger:
this is useful for grown-up users (and devs) to be able to use weak passwords if they so choose, while keeping the constraints enabled by default - it currently only guards the libpwquality checks; and is hidden when libpwquality is not linked in

### DIFF
--- a/src/modules/users/UsersPage.cpp
+++ b/src/modules/users/UsersPage.cpp
@@ -88,6 +88,13 @@ UsersPage::UsersPage( QWidget* parent )
              this, &UsersPage::onRootPasswordTextChanged );
     connect( ui->textBoxVerifiedRootPassword, &QLineEdit::textChanged,
              this, &UsersPage::onRootPasswordTextChanged );
+    connect( ui->checkBoxValidatePassword, &QCheckBox::stateChanged,
+             this, [this]( int checked )
+    {
+        onPasswordTextChanged( ui->textBoxUserPassword->text() );
+        onRootPasswordTextChanged( ui->textBoxRootPassword->text() );
+        checkReady( isReady() );
+    } );
     connect( ui->checkBoxReusePassword, &QCheckBox::stateChanged,
              this, [this]( int checked )
     {
@@ -105,6 +112,8 @@ UsersPage::UsersPage( QWidget* parent )
 
     setWriteRootPassword( true );
     ui->checkBoxReusePassword->setChecked( true );
+    ui->checkBoxValidatePassword->setChecked( true );
+    ui->checkBoxValidatePassword->setVisible( false );
 
     // Don't expand the explanations to "stupid wide", but keep them vaguely as-wide-as
     // the things they are explaining.
@@ -387,6 +396,7 @@ UsersPage::onPasswordTextChanged( const QString& )
 {
     QString pw1 = ui->textBoxUserPassword->text();
     QString pw2 = ui->textBoxUserVerifiedPassword->text();
+    m_readyPassword = true;
 
     // TODO: 3.3: remove empty-check and leave it to passwordRequirements
     if ( pw1.isEmpty() && pw2.isEmpty() )
@@ -403,23 +413,24 @@ UsersPage::onPasswordTextChanged( const QString& )
     }
     else
     {
-        bool ok = true;
-        for ( auto pc : m_passwordChecks )
+        if ( ui->checkBoxValidatePassword->isChecked() )
         {
-            QString s = pc.filter( pw1 );
-            if ( !s.isEmpty() )
+            for ( auto pc : m_passwordChecks )
             {
-                labelError( ui->labelUserPassword, ui->labelUserPasswordError, s );
-                ok = false;
-                m_readyPassword = false;
-                break;
+                QString s = pc.filter( pw1 );
+
+                if ( !s.isEmpty() )
+                {
+                    labelError( ui->labelUserPassword, ui->labelUserPasswordError, s );
+                    m_readyPassword = false;
+                    break;
+                }
             }
         }
 
-        if ( ok )
+        if ( m_readyPassword )
         {
             labelOk( ui->labelUserPassword, ui->labelUserPasswordError );
-            m_readyPassword = true;
         }
     }
 
@@ -431,6 +442,7 @@ UsersPage::onRootPasswordTextChanged( const QString& )
 {
     QString pw1 = ui->textBoxRootPassword->text();
     QString pw2 = ui->textBoxVerifiedRootPassword->text();
+    m_readyRootPassword = true;
 
     // TODO: 3.3: remove empty-check and leave it to passwordRequirements
     if ( pw1.isEmpty() && pw2.isEmpty() )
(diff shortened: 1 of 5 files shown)


src/modules/users/CheckPWQuality.cpp
```diff
@@ -31,20 +31,23 @@
 #include <memory>
 
 PasswordCheck::PasswordCheck()
-    : m_message()
+    : m_message( [](){ return ""; } )
```
adriaandegroot: That lambda should probably be a free function (so it's easier to add APIDOX to it explainng that it's a MessageFunc that doesn't complain), and return QString() rather than "".

src/modules/users/CheckPWQuality.h
```diff
@@ -38,10 +38,12 @@ class PasswordCheck
     using AcceptFunc = std::function<bool( const QString& )>;
     using MessageFunc = std::function<QString()>;
 
+    enum CheckId { DEFAULT, LIBPWQ };
```
adriaandegroot: Needs dox, and I'm not a big fan of these names; perhaps InternalCheck and LibPQWCheck, and rename the *CheckId* (since it's **not** an indentity) to *CheckFamily* to emphasise that it applies to a family of checks?

src/modules/users/CheckPWQuality.h
```diff
@@ -55,9 +57,19 @@ class PasswordCheck
         return m_accept( s ) ? QString() : m_message();
     }
 
+    /** Retturns a unique identifier of the check.
```
adriaandegroot: Typo "retturns"

src/modules/users/CheckPWQuality.h
```diff
@@ -55,9 +57,19 @@ class PasswordCheck
...
+     *  Consumers of this class may use this for special handling of specific checks.
+     *  Currently the only handled option is 'libpwquality'.
```
adriaandegroot: "some families of checks", "handled option" -> "family that is handled specially", `libpwquality` should match the name of the family / Id from CheckId.

src/modules/users/CheckPWQuality.h
```diff
@@ -38,10 +38,12 @@ class PasswordCheck
     using AcceptFunc = std::function<bool( const QString& )>;
     using MessageFunc = std::function<QString()>;
 
+    enum CheckId { DEFAULT, LIBPWQ };
+
     /** Generate a @p message if @p filter returns true */
-    PasswordCheck( MessageFunc message, AcceptFunc filter );
+    PasswordCheck( MessageFunc message, AcceptFunc filter, CheckId id = DEFAULT );
```
adriaandegroot: Please don't use default values in constructors. If needed, add them explicitly where the constructors are called.

src/modules/users/UsersPage.cpp
```diff
@@ -88,6 +88,13 @@ UsersPage::UsersPage( QWidget* parent )
              this, &UsersPage::onRootPasswordTextChanged );
     connect( ui->textBoxVerifiedRootPassword, &QLineEdit::textChanged,
              this, &UsersPage::onRootPasswordTextChanged );
+    connect( ui->checkBoxValidatePassword, &QCheckBox::stateChanged,
+             this, [this]( int checked )
+    {
+        onPasswordTextChanged(ui->textBoxUserPassword->text());
+        onRootPasswordTextChanged(ui->textBoxRootPassword->text());
+        checkReady( isReady() );
+    } );
```
adriaandegroot: Please apply `ci/calamaresstyle` as well.

src/modules/users/UsersPage.cpp
```diff
@@ -403,23 +417,25 @@ UsersPage::onPasswordTextChanged( const QString& )
     }
     else
     {
-        bool ok = true;
         for ( auto pc : m_passwordChecks )
         {
             QString s = pc.filter( pw1 );
+
+            if ( ! ui->checkBoxValidatePassword->isChecked() &&
```
adriaandegroot: I'd suggest factoring this out into a `isCheckEnabled( pc )` method of UsersPage so the special cases can be handled in one place.

src/modules/users/UsersPage.cpp
```diff
@@ -447,30 +464,39 @@ UsersPage::onRootPasswordTextChanged( const QString& )
...
```
adriaandegroot: .. then you can use `isCheckEnabled( pc )` here again.

src/modules/users/UsersPage.cpp
```diff
@@ -447,30 +464,39 @@ UsersPage::onRootPasswordTextChanged( const QString& )
...
-            m_readyRootPassword = true;
         }
     }
 
     emit checkReady( isReady() );
 }
 
 
+void
+UsersPage::setValidatePasswordDefault( bool checked )
```
adriaandegroot: the "Default" part here seems weird -- you're actually doing the setting, not changing a default somewhere.

src/modules/users/UsersViewStep.cpp
```diff
@@ -170,6 +170,10 @@ UsersViewStep::setConfigurationMap( const QVariantMap& configurationMap )
             i != pr_checks.constEnd(); ++i)
         {
             m_widget->addPasswordCheck( i.key(), i.value() );
+#ifdef CHECK_PWQUALITY
+            if ( i.key() == "libpwquality" )
+                m_widget->setValidatePasswordDefault( true );
```
adriaandegroot: I think this should be done in UsersPage::addPasswordCheck() .. that too is called only during configuration, and already handles the libpwquality case, so you could just move the setValidatePasswordDefault() call to there.

src/modules/users/page_usersetup.ui
```diff
@@ -460,6 +460,13 @@
      </property>
     </spacer>
    </item>
+   <item>
+    <widget class="QCheckBox" name="checkBoxValidatePassword">
+     <property name="text">
+      <string>Require strong passwords.</string>
```
adriaandegroot: I expect -- without having compiled this to see what it looks like -- that this is going to add clutter to an already cluttered form. Can we get a designer to take a look at this? I suspect we might end up with a dots-menu about passwords.

bill-auger:
ok ill make some changes, taking your suggestions into consideration

regarding the GUI, it looks fine on my screen - here is a screen-shot

Screenshot from 2019-08-29 10-27-07

kkofler:
One behavior change in this pull request is that, if `libpwquality` is found at build time, it is always enabled by default, even if not configured.

Currently, it is possible to build Calamares against `libpwquality`, but to disable the checks in the default configuration (by just commenting out the `libpwquality:` section from `users.conf`). I ship Calamares in Fedora configured that way, so that derivative distributions may or may not require password strength. (Kannolo doesn't.)

As far as I can tell, your pull request as currently submitted breaks that: the checkbox is enabled from 2 places: `UsersViewStep.cpp`, where you actually check that the `libpwquality:` configuration section exists, and `UsersPage.cpp`, where you don't.

bill-auger:
i re-worked this into a simpler design that addresses the comments so far - to make the passwords checkbox implementation maximally configurable, and more uniform with the other checkboxes, and to satisfy kevin's concern, i added a new config key: 'doPasswordChecks' analog to 'doAutologin' and 'doReusePassword' - that governs the all password validations (othr than the "Your passwords do not match!", which is always enabled); and a second one 'allowWeakPasswords' to control the visibility of the checkbox (and the user's ability to enter a weak password) via a new 'setPasswordCheckboxVisible()' method - the default behaviour when neither 'allowWeakPasswords' nor 'doPasswordChecks' are defined, is for the 'checkBoxValidatePassword' checkbox to be checked (all checks defined in the 'passwordRequirements' hash are enabled), and for the 'checkBoxValidatePassword' checkbox to be hidden (not possible for the user to disable the checks) - by not discriminating the libpwquality checks from the internal ones, this simpler design also eliminates all changes to the 'PasswordCheck' class

regarding the semantics of 'setValidatePasswordDefault()': that is as it is because i implemented the 'checkBoxValidatePassword' to be consistent with the other two: ('checkBoxAutoLogin' , 'checkBoxReusePassword') - for those, the setter methods ('setAutologinDefault()' and 'setReusePasswordDefault()') are called only once, in order to over-ride the hard-coded default when parsing the config file, and only if the key is present; so the semantics is the same for all three - its really more like 'setInitialState()' - likewise, i noticed that the semantics of the 'doAutologin' and 'doReusePassword' config keys are really more like 'initialDoAutologin' and 'initialDoReusePassword'; but it would be a breaking change to rename those

bill-auger:
regarding the GUI layout: i noticed the feature request for adding some option to the user-setup view to select the login shell; so i looked into redesigning the user-setup view - doing so, realized that my suggestion of putting both realname and username in a new Hbox would not save any space because of the error texts

i also realized that the hostname field is the one field on this view that is not clearly within the scope of user creation - it is rather a miscellaneous environmental setting - that field is auto-filled anyways, and many (or most) users would probably be content with (or never even notice) if the default hostname: 'mylogin-pc' was not configurable - so another idea to un-clutter the user-setup view would be to add a new optional 'environment' view module to present the hostname option, and any future miscellaneous additions that do not fit neatly into any existing modules - that would either make space for the login shell selector, or the login shell selector could be put into the 'environment' module also, keeping the existing hard-coded fallback: /bin/bash; so that the 'environment' module could be entirely optional - there are some non-trivial caveats to that feature that would need addressing though; such as the existing 'userShell' config key being a single value rather than a list as with 'availableShells' in PR #280; so its outside the scope of this feature - the new 'environment' module would be fairly easy to add to this though, albeit the hostname field would be lonely there on its own

adriaandegroot:
I'm working on integrating this now (and dealing with #1202 at the same time)

PR #1243, opened 2019-09-08
QLatin1Literal->QLatin1String.

stikonas:


adriaandegroot:
.. because?

adriaandegroot:
Or, while we're at it, why not QStringLiteral? I don't have a particular attachment to any one form of literals, which is in turn probably why it's inconsistent across the codebase.

stikonas:
Because it's gone from Qt 5.14. QStringliteral is not and is still useful.

On 9 September 2019 15:45:21 BST, Adriaan de Groot <notifications@github.com> wrote:
>.. because?
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/calamares/calamares/pull/1243#issuecomment-529510582

-- 

adriaandegroot:
My question actually meant "why not port to QStringLiteral", but that was amazingly ambiguous. And now I've read up on things, I'm not sure QStringLiteral would be the answer anyway. 

Calamares is kind of inconsistent in dealing with character-literals. Does it even compile with NO_ASCII_CAST? If not this branch could be a lot longer with making that happen.

That said, this MR no longer applies, I think I changed locale/ things underneath you. Can you fix that up?

stikonas:
I guess it won't compile with NO_ASCII_CAST if you don't actually test with that flag.

Anyway, I've updated my merge request now.

PR #1259, opened 2019-10-23
WIP: Use localized strings for region and timezone

balasankarc:
So that combo-boxes for selecting region and timezone also show localized strings.

### DIFF
--- a/src/modules/locale/LocalePage.cpp
+++ b/src/modules/locale/LocalePage.cpp
@@ -154,7 +154,7 @@ LocalePage::init( const QString& initialRegion, const QString& initialZone, cons
 
     foreach ( const QString& key, keys )
     {
-        m_regionCombo->addItem( LocaleGlobal::Location::pretty( key ), key );
+        m_regionCombo->addItem( tr(LocaleGlobal::Location::pretty( key ).toUtf8().constData()), key );
     }
 
     m_regionCombo->blockSignals( false );
@@ -420,7 +420,7 @@ LocalePage::regionChanged( int currentIndex )
     const QList< LocaleGlobal::Location > zones = regions.value( m_regionCombo->currentData().toString() );
     for ( const LocaleGlobal::Location& zone : zones )
     {
-        m_zoneCombo->addItem( LocaleGlobal::Location::pretty( zone.zone ), zone.zone );
+        m_zoneCombo->addItem( tr(LocaleGlobal::Location::pretty( zone.zone ).toUtf8().constData()), zone.zone );
     }
 
     m_zoneCombo->model()->sort( 0 );


kkofler:
Uh, normally, it is not a good idea to call `tr` on a variable string, because the strings will not actually show up in the translation templates. Where do the strings come from and where should their translations come from?

balasankarc:
> Uh, normally, it is not a good idea to call `tr` on a variable string, because the strings will not actually show up in the translation templates. Where do the strings come from and where should their translations come from?

I agree, and I don't have the answers yet. Hence the `WIP` in the title. :)

I see we read this info from `TZ_DATA_FILE` (which points at `/usr/share/zoneinfo/zone.tab`). In other projects I have contributed to, (take GNOME Evolution for example - https://gitlab.gnome.org/GNOME/evolution/blob/master/src/calendar/zones.h), this information is copied to the project source so that `gettext` picks them, and hence they have a fixed source for strings. However, we are using the system provided one, which is in theory a variable source.

balasankarc:
@adriaandegroot @kkofler How do you feel about dropping the dependency on system provided `/usr/share/zoneinfo/zone.tab` file in favor of maintaining a copy of that file (or the same information in a suitable format) in the repo, so that we have a single source of truth for strings to be localized? The downside is we will have to remember to periodically update it.

adriaandegroot:
If there's some shared effort (freedesktop, maybe?) then we should really try to use that, because this is going to be a pretty big kind of database (I suppose we could translate column four from zone.tab)

balasankarc:
@adriaandegroot

> If there's some shared effort (freedesktop, maybe?) then we should really try to use that,

A shared effort to maintain the localizations of timezone data, specifically? I will look around, but I haven't seen something like that yet.

> I suppose we could translate column four from zone.tab

For it to be picked up by gettext (and to show up in translation templates), we need to store them somewhere, right? Did you mean extracting last column of zone.tab file, and keeping it in source (similar to what I shared above that is done by GNOME - https://gitlab.gnome.org/GNOME/evolution/blob/master/src/calendar/zones.h) ?

adriaandegroot:
I've followed the GNOME approach with an extracted list of names. That's step 1. Step 2 will be to use them; I don't think the approach in the commit here is usable because you translate only once, when the items are inserted. They need to have a model with a proper name-lookup that respects the locale in use at the time the item is displayed.

adriaandegroot:
Work continues in the *zone-data* branch; it has a list of region and zone names; I'm mostly wondering if we can't source this from somewhere else instead.

balasankarc:
With 2034dd62a4da779fe8166dedffaea499c0a7877e, this PR is outdated. Hence closing.

Thanks for driving this forward @adriaandegroot

adriaandegroot:
This is now implemented. I've chosen to make a separate translation file / domain of the timezone data because it's 450 strings and that would kill everyone's translation statistics if I added it in one go. Besides, most of the names **don't** need translation.

You can contribute a `lang/tz_<locale>.ts` file, or wait until next week when I push it to Transifex.

balasankarc:
> or wait until next week when I push it to Transifex.

That should be fine. Thanks for implementing it. :)

PR #1272, opened 2019-11-12
esperanto translations for nav buttons

bill-auger:
keyboard navigation is broken now with the esperanto translation; because the "cancel" button has a translation "Nulligi", which makes 'N' the hotkey, conflicting with the 'next' button, which is not translated - this PR translates all hotkeys


adriaandegroot:
I can't merge this because the translations are pulled from Transifex -- so these changes will get overwritten. I'll paste these strings into TX though, so **effectively** you've updated the translations.

adriaandegroot:
Duh, I can merge this, then upload the file to TX for the same effect, which would make it clearer where the translations come from. In any case, with the next TX pull these are applied.

PR #1275, opened 2019-11-12
housekeeping - logging for CreateUser Job

bill-auger:


src/modules/users/CreateUserJob.cpp
```diff
@@ -74,6 +74,8 @@ CreateUserJob::exec()
 
     if ( gs->contains( "sudoersGroup" ) && !gs->value( "sudoersGroup" ).toString().isEmpty() )
     {
+        cDebug() << QString("[CREATEUSER]: preparing sudoers") ;
```
adriaandegroot: That doesn't need to be a QString() at all; generally you'd use QStringLiteral() for this kind of literal (it saves a few cycles and is better with const) but here, you can log a C++ string literal: just drop QString() from there and leave the bare string literal.

src/modules/users/CreateUserJob.cpp
```diff
@@ -96,6 +98,8 @@ CreateUserJob::exec()
             return Calamares::JobResult::error( tr( "Cannot chmod sudoers file." ) );
     }
 
+    cDebug() << QString("[CREATEUSER]: preparing groups") ;
```
adriaandegroot: Ditto (and the one below, too)

PR #1276, opened 2019-11-12
pre-select default item in packagechooser

bill-auger:
while evaluating the 'packagechooser' module, i noted some things that i did not like - one of them being that there was no way to have a static description/explanation of the module - the other more importanlty thing was that there is no way to pre-select a default list item - that would be quite useful, especially in "required" modes -

this MR accomplishes both of those features, while retaining the original behavior, if desired; by adding two new config keys to over-ride the existing null item method, if desired

the documentation for the 'item['id]' indicates that it is not intended to be used in 'required' modes

> Only include this if the mode allows selecting none.

for that reason, this MR also removes the null item if present and in one of the 'required' modes, regardless of the new over-rides

### DIFF
--- a/src/modules/packagechooser/PackageChooserViewStep.cpp
+++ b/src/modules/packagechooser/PackageChooserViewStep.cpp
@@ -144,6 +144,12 @@ PackageChooserViewStep::isAtEnd() const
     return true;
 }
 
+void
+PackageChooserViewStep::onActivate()
+{
+    if ( !m_widget->hasSelection() )
+        m_widget->setSelection( m_defaultIdx );
+}
 
 void
 PackageChooserViewStep::onLeave()
@@ -198,6 +204,9 @@ PackageChooserViewStep::setConfigurationMap( const QVariantMap& configurationMap
         }
     }
 
+    QString default_item_id = CalamaresUtils::getString( configurationMap, "default" );
+    m_defaultIdx = QModelIndex();
+
     bool first_time = !m_model;
     if ( configurationMap.contains( "items" ) )
     {
@@ -207,6 +216,22 @@ PackageChooserViewStep::setConfigurationMap( const QVariantMap& configurationMap
     if ( first_time && m_widget && m_model )
     {
         hookupModel();
+    }
+
+    // find default item
+    if ( first_time && m_model && !default_item_id.isEmpty() )
+    {
+        for (int item_n = 0; item_n < m_model->packageCount(); ++item_n)
+        {
+            QModelIndex item_idx = m_model->index( item_n, 0 );
+            QVariant item_id = m_model->data( item_idx, PackageListModel::IdRole );
+
+            if ( item_id.toString() == default_item_id )
+            {
+                m_defaultIdx = item_idx;
+                break;
+            }
+        }
     }
 }
 
--- a/src/modules/packagechooser/PackageChooserViewStep.h
+++ b/src/modules/packagechooser/PackageChooserViewStep.h
@@ -50,6 +50,7 @@ public:
     bool isAtBeginning() const override;
     bool isAtEnd() const override;
 
+    void onActivate() override;
     void onLeave() override;
 
     Calamares::JobList jobs() const override;
@@ -67,6 +68,7 @@ private:
     PackageChooserMode m_mode;
     QString m_id;
     CalamaresUtils::Locale::TranslatedString* m_stepName;  // As it appears in the sidebar
+    QModelIndex m_defaultIdx;
 };
 
 CALAMARES_PLUGIN_FACTORY_DECLARATION( PackageChooserViewStepFactory )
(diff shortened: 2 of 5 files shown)


src/modules/packagechooser/PackageChooserViewStep.h
```diff
@@ -67,6 +68,8 @@ class PLUGINDLLEXPORT PackageChooserViewStep : public Calamares::ViewStep
     PackageChooserMode m_mode;
     QString m_id;
     CalamaresUtils::Locale::TranslatedString* m_stepName;  // As it appears in the sidebar
+    QString m_description;
+    int m_defaultIdx;
```
adriaandegroot: Perhaps this one should be a string, rather than an int, so it can be matched to the *id* of one of the items rather than their presentation order?

src/modules/packagechooser/PackageChooserViewStep.cpp
```diff
@@ -199,9 +213,28 @@ PackageChooserViewStep::setConfigurationMap( const QVariantMap& configurationMap
...
 
     bool first_time = !m_model;
-    if ( configurationMap.contains( "items" ) )
+    QVariantList list_items = configurationMap.value( "items" ).toList();
+    if ( !list_items.isEmpty() )
     {
-        fillModel( configurationMap.value( "items" ).toList() );
+        // remove null item if unused
+        if ( m_mode == PackageChooserMode::Required ||
+             m_mode == PackageChooserMode::RequiredMultiple || m_defaultIdx >= 0)
```
adriaandegroot: I'm not really convinced that having a preset-set should mean that the null item is removed.

src/modules/packagechooser/PackageChooserViewStep.cpp
```diff
@@ -188,6 +196,12 @@ PackageChooserViewStep::setConfigurationMap( const QVariantMap& configurationMap
         m_id = moduleInstanceKey().split( '@' ).last();
     }
 
+    m_description = ( configurationMap.contains( "description" ) ) ?
+                      configurationMap.value( "description" ).toString() : QString();
```
adriaandegroot: `CalamaresUtils::getString( map, key )`

src/modules/packagechooser/PackageChooserViewStep.cpp
```diff
@@ -199,9 +213,28 @@ PackageChooserViewStep::setConfigurationMap( const QVariantMap& configurationMap
...
+
+                if ( list_item_id.isEmpty() )
+                {
+                    list_items.removeAt( list_item_n );
+                    break;
+                }
+            }
+        }
+
+        fillModel( list_items );
```
adriaandegroot: Perhaps instead of removing the item at all, a FilterProxyModel would work?

src/modules/packagechooser/PackageChooserViewStep.cpp
```diff
@@ -188,6 +196,12 @@ PackageChooserViewStep::setConfigurationMap( const QVariantMap& configurationMap
         m_id = moduleInstanceKey().split( '@' ).last();
     }
 
+    m_description = ( configurationMap.contains( "description" ) ) ?
```
adriaandegroot: Maybe we can get this from the empty-id item if it's there, rather than from a new key? It seems to me that you're taking the description from the `.conf` file too literally and that we can make "" work for all kinds of module description.

bill-auger:
i agree that matching to the 'id' is better - i made that change
and split that feature into its own commit

the reason for removing the null item, was mainly that the
documentation for the 'item['id]' indicates that the null item
was not intended to be used in 'required' modes - yet it is
allowed and used regardless of the mode

the reason for adding a new 'description' key to replace the
null item description, was that the null item description is
being confusingly over-loaded for that purpose

all of the other item descriptions are describing the purpose
of that item; but the null item description is describing the
purpose of the entire module, and not its own purpose - it would
be more sensible to use the null item description to describe the
purpose of that item; such as: "No additional packages are
selected. Only the basic system will be installed." - or, in the
specific case of the example .conf file, it would be
"Command-line system only. No desktop." - when seen that way, it
is obvious that the null item should not be visible with the
'required' modes - i.e. it would not be meaningful to show "No
desktop will be installed" if choosing a desktop is mandatory

also, by using the null item description to fill the static
module description, then the same text would appear in two places
whenever the null item is selected, which will always be the
initial state if 'default' is undefined

alternatively, if the null item description were used always,
for the purpose of the static module explanation. as you
suggested, then the documentation should change to make it clear
that it is applicable to and will be used for 'required' modes,
and is actually necessary in order to have a static explanation,
even when the null item itself is to be hidden - but, to me, that
has the design smell of "shoe-horning", mixed or over-loaded
semantics

adriaandegroot:
Makes sense, thanks for the explanation

bill-auger:
oh, i should have mentioned that i did not plan to update this
PR - the other part was really unrelated and doesnt really add
any new functionality - the work-around is just to treat the
empty string in global-storage as meaningful - it just seems like
the semantics would be better as i described it - the next time i
compile, i planned to re-work it and open a different ticket so
maybe others would discuss it or refute it

PR #1290, opened 2019-12-16
[initcpiocfg] update plymouth check

philmmanjaro:
We should use the same **plymouth check** as we use with `plymouthcfg` module. Using just a binary file check might cause issues when the rootfs was not formatted and for what ever reason the binary of plymouth exists. Using `which` makes it more solid.

---

https://forum.manjaro.org/t/installation-failed-mkinitcpio-finished-with-errors/113921


adriaandegroot:
Merged with minor polish, thanks!

PR #1292, opened 2019-12-19
Fix issue 1286

berolinux:
automoc in cmake 3.16.1 adds "-p subdirectory" to moc commands run for
files in a subdirectory, breaking inclusion of utils/moc-warnings.h the
way calamares currently does it for clang users.


adriaandegroot:
This was fixed in 1bc8a28eac026f1e3140681f6339f9b00deb4154

PR #1298, opened 2020-01-06
i18n: Replace Kiev with Kyiv. Kiev is incorrect spelling

satorym:
Official statement of the Ministry of Foreign Affairs of Ukraine
https://mfa.gov.ua/en/page/open/id/5418

### DIFF
--- a/lang/tz_en.ts
+++ b/lang/tz_en.ts
@@ -1283,7 +1283,7 @@
     </message>
     <message>
         <location filename="../src/libcalamares/locale/ZoneData_p.cxxtr" line="253"/>
-        <source>Kiev</source>
+        <source>Kyiv</source>
         <comment>tz_names</comment>
         <translation type="unfinished"></translation>
     </message>
--- a/lang/tz_nl.ts
+++ b/lang/tz_nl.ts
@@ -1283,7 +1283,7 @@
     </message>
     <message>
         <location filename="../src/libcalamares/locale/ZoneData_p.cxxtr" line="253"/>
-        <source>Kiev</source>
+        <source>Kyiv</source>
         <comment>tz_names</comment>
         <translation type="unfinished"></translation>
     </message>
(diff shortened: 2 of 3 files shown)


philmmanjaro:
Normally it just needs to be changed in `src/libcalamares/locale/ZoneData_p.cxxtr` as **ts**-files get generated automatically by **transifex**.

adriaandegroot:
The zonedata file is a bit unusual: it's **not** in Transifex yet, because it's a whole lot of strings that I didn't want to bury the translators under. I need to figure out how to make 400 extra strings not kill all the translation stats. In the end I might just leave it in git and leave timezone translation to other mechanisms.

The purpose of the zonedata file is to present the names of timezones in a localized fashion. The authoritative source of zone names is `/usr/share/zoneinfo/zone.tab` and as long as it says *Europe/Kiev* that will be the **source** name. So changing the C++ code (in the file marked `GENERATED FILE DO NOT EDIT`) will just break things (the label of the timezone will then be untranslated Latin-1 *Kiev* for everyone).

It's possible to change the translations into various languages, though. The reason there are (only) English and Dutch translations is that I have a few simple examples that apply there -- Paris and Berlin being the main ones. Those are an example -- the real purpose here is for non-Latin-1 languages to be **able** to present, say, Europe/Kiev timezone as Київ when you pick Ukranian as installer language. Whether it makes sense to present Antarctica/Troll in any other form than Latin-1 *Troll*, I don't know.

In Dutch, the place along the Dniepr is still spelled *Kiev*, like the place on the Seine is spelled *Parijs*. (Source: Dutch ministry for foreign affairs maps, updated 2019). In English (which is used much more widely) the spelling is shifting towards *Kyiv*, so I'd be willing to accept that single change (which would mean that clicking on the world map near Zaporozhye would display *Kyiv* both on the map and in the *Zone:* drop-down).

adriaandegroot:
(There's a `lang/tz_uk.ts` now, please *do* apply translations there as appropriate)

PR #1300, opened 2020-01-14
adding configure option dummyqml

demmm:
sidebar entry can be configured and translated
adding a more elaborate qml example
keeping this in dummyqml for now, another commit will follow with
continuation of dummyqml in a more aptly named module

### DIFF
--- a/src/modules/dummyqml/DummyQmlViewStep.cpp
+++ b/src/modules/dummyqml/DummyQmlViewStep.cpp
@@ -19,6 +19,9 @@
 
 #include "DummyQmlViewStep.h"
 
+#include "utils/PluginFactory.h"
+#include "viewpages/QmlViewStep.h"
+
 #include <QVariant>
 
 DummyQmlViewStep::DummyQmlViewStep( QObject* parent )
--- a/src/modules/dummyqml/DummyQmlViewStep.cpp
+++ b/src/modules/dummyqml/DummyQmlViewStep.cpp
@@ -1,6 +1,7 @@
 /* === This file is part of Calamares - <https://github.com/calamares> ===
  *
  *   Copyright 2020, Adriaan de Groot <groot@kde.org>
+ *   Copyright 2020, Anke Boersma <demm@kaosx.us>
  *
  *   Calamares is free software: you can redistribute it and/or modify
  *   it under the terms of the GNU General Public License as published by
@@ -18,6 +19,8 @@
 
 #include "DummyQmlViewStep.h"
 
+#include <QVariant>
+
 DummyQmlViewStep::DummyQmlViewStep( QObject* parent )
     : Calamares::QmlViewStep( "dummyqml", parent )
 {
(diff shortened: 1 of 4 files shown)


src/modules/dummyqml/DummyQmlViewStep.h
```diff
@@ -19,20 +20,27 @@
 #ifndef DUMMYQMLVIEWSTEP_H
 #define DUMMYQMLVIEWSTEP_H
 
+#include <PluginDllMacro.h>
```
adriaandegroot: This one with "" as well (it's not a system header).

src/modules/dummyqml/DummyQmlViewStep.h
```diff
@@ -19,20 +20,27 @@
...
+#include "locale/TranslatableConfiguration.h"
```
adriaandegroot: Yes, sure. It's shorter than declaring the class (which is all you'd need for that one pointer member).

src/modules/dummyqml/DummyQmlViewStep.h
```diff
@@ -19,20 +20,27 @@
...
+#include "utils/CalamaresUtilsSystem.h"
+#include "utils/Variant.h"
```
adriaandegroot: Those two should be in the .cpp

src/modules/dummyqml/DummyQmlViewStep.h
```diff
@@ -19,20 +20,27 @@
...
     DummyQmlViewStep( QObject* parent = nullptr );
     virtual ~DummyQmlViewStep() override;
 
-    virtual QString prettyName() const override;
+    QString prettyName() const override;
+    
+    void setConfigurationMap( const QVariantMap& configurationMap ) override;
 
 private:
+    CalamaresUtils::Locale::TranslatedString* m_notesName;  // As it appears in the sidebar
```
adriaandegroot: Needs a destructor to clean it up (even if right now, Calamares module destructors are run .. only sometimes)

src/modules/dummyqml/DummyQmlViewStep.cpp
```diff
@@ -28,8 +31,23 @@ DummyQmlViewStep::~DummyQmlViewStep() {}
...
 }
 
+void
+DummyQmlViewStep::setConfigurationMap( const QVariantMap& configurationMap )
+{ 
+    Calamares::QmlViewStep::setConfigurationMap( configurationMap ); // call parent implementation
+    
+    bool qmlLabel_ok = false;
+    auto qmlLabel = CalamaresUtils::getSubMap( configurationMap, "qmlLabel", qmlLabel_ok );
+    if ( qmlLabel_ok )
```
adriaandegroot: If it's not ok, the returned map is empty, so it won't contain "notes". you can remove this check (or fold it with the nested if).

adriaandegroot:
Changes requested, but the whole looks really good. This makes it one step closer to having a generic "informative QML page". We should keep watching for useful things to fold "upstream" into the QmlViewStep base class, so that they become useful for other QML-UI consumers as well.

demmm:
Will try and resolve all.  Start a new PR and remove this one?
And for 
>We should keep watching for useful things to fold "upstream" into the QmlViewStep base class

I have opening an outside file (not part of Calamares, can be local or on web) about working, just can't figure a workaround yet to not need to have the file part of qrc.

adriaandegroot:
- Keep this PR and update it
 - Check 750465153f4d6a329b97b2ed9786423b618411d9, that should load externals properly (there's some trickery with schemes, I think)

(Sorry about being terse, at conference far away from home)

src/modules/dummyqml/DummyQmlViewStep.h
demmm: Fails to build if I move #include "utils/CalamaresUtilsSystem.h" and #include "utils/Variant.h" from the header to cpp file.

src/modules/dummyqml/DummyQmlViewStep.h
demmm: No idea what to do here

demmm:
Besides my new comments, other issues have been changed, build & tested as working, but since I can't fix all that is needed, I will close this pull request in 1-2 days.

adriaandegroot:
I've merged this into the *qml* branch, needs a little more testing, and would be good to combine with the new descriptor and instancekey work (that's my job, though, so I'm closing the PR as "merged").

PR #1331, opened 2020-02-27
[packages] Add apk (Alpine Linux package manager) support

PureTryOut:
Yay for easy Python scripting!

Since we're looking into using Calamares as on-device installer for postmarketOS, let's add support for the Alpine Linux (which postmarketOS is based on) package manager.

Is this the only required bit as far as packages go, or is there some other things that need specific Alpine Linux support to work?

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -364,6 +364,24 @@ class PMPisi(PackageManager):
         pass
 
 
+class PMApk(PackageManager):
+    backend = "apk"
+
+    def install(self, pkgs, from_local=False):
+        for pkg in pkgs:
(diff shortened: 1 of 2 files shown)


philmmanjaro:
Calamares is highly customizable. It is always good to support more package managers for the **packages** module. The more the merry. You commit looks good to me. Feel free to ping me if you need a guideline on how to start with Calamares.

PureTryOut:
First I'm going to try get it working properly with Alpine Linux. Once that's done I'll look into integrating it in postmarketOS. I already joined #calamares on Freenode so I'll ask for help there (already did just now for an issue ;)

PR #1341, opened 2020-03-17
[partition] Add support for filesystem-specific features

gportay:
_KPMcore_ is now supporting for filesystem-specific features since the version _4.1.0_.

The file-system __features__ can be set using the module partition by specifying the dictionary `features` for any partition in `partitionLayout`. The feature is specific to the filesystem; the feature as a name and a value that can be either a boolean flag or an integer value or a string.

The example below tests the boolean and integer values feature. It:

- sets the both _FAT32_ features `sector-size` and `sectors-per-cluster` that take an integer value to the partition _efi_, and
- sets the _ext4_ feature `casefold` and resets the _ext4_ feature `64bit` that are two boolean flags to the partition _rootfs_. 

``` yaml
$ cat partition.conf
(...)
partitionLayout:
  - name: "efi"
    filesystem: "fat32"
    mountPoint: "/efi"
    size: 128M
    minSize: 128M
    features:
      sector-size: 4096
      sectors-per-cluster: 128
  - name: "rootfs"
    filesystem: "ext4"hg
    mountPoint: "/"
    size: 1G
    minSize: 1G
    features:
      casefold: true
      64bit: false
```

KPMcore generates the following commands:

-  `/usr/bin/mkfs.vat -S4096 -s124 -F32 -I -v /dev/sda1`
- `/usr/sbin/mkfs.ext4 -O ^64bit,casefold -qF /dev/sda2`

This commit uses the changes in the [commit] to add that support in `calamares`.

I hope you will like that contribution.

[commit]: https://github.com/KDE/kpmcore/commit/d24191ebd8ee4a1792003e5bd280cb7679f5e834

### DIFF
--- a/src/modules/partition/core/PartitionCoreModule.cpp
+++ b/src/modules/partition/core/PartitionCoreModule.cpp
@@ -834,6 +834,7 @@ PartitionCoreModule::initLayout()
 void
 PartitionCoreModule::initLayout( const QVariantList& config )
 {
+    bool ok;
     QString sizeString;
     QString minSizeString;
     QString maxSizeString;
@@ -884,6 +885,7 @@ PartitionCoreModule::initLayout( const QVariantList& config )
         if ( !m_partLayout->addEntry( CalamaresUtils::getString( pentry, "name" ),
                                       CalamaresUtils::getString( pentry, "mountPoint" ),
                                       CalamaresUtils::getString( pentry, "filesystem" ),
+                                      CalamaresUtils::getSubMap( pentry, "features", ok ),
                                       sizeString,
                                       minSizeString,
                                       maxSizeString ) )
--- a/src/modules/partition/core/PartitionLayout.cpp
+++ b/src/modules/partition/core/PartitionLayout.cpp
@@ -120,6 +120,7 @@ bool
 PartitionLayout::addEntry( const QString& label,
                            const QString& mountPoint,
                            const QString& fs,
+                           const QVariantMap& features,
                            const QString& size,
                            const QString& min,
                            const QString& max )
@@ -144,6 +145,7 @@ PartitionLayout::addEntry( const QString& label,
     {
         entry.partFileSystem = m_defaultFsType;
     }
+    entry.partFeatures = features;
 
     m_partLayout.append( entry );
 
@@ -238,6 +240,15 @@ PartitionLayout::execute( Device* dev,
         if ( !part.partLabel.isEmpty() )
         {
             currentPartition->fileSystem().setLabel( part.partLabel );
+        }
+        if ( !part.partFeatures.isEmpty() )
+        {
+#if defined( WITH_KPMCORE42API )
+            for ( const auto& k : part.partFeatures.keys() )
+                currentPartition->fileSystem().addFeature( k, part.partFeatures.value(k) );
+#else
+            cWarning() << "Ignoring features; requires KPMcore >= 4.2.0.";
+#endif
         }
         // Some buggy (legacy) BIOSes test if the bootflag of at least one partition is set.
         // Otherwise they ignore the device in boot-order, so add it here.
--- a/src/modules/partition/partition.conf
+++ b/src/modules/partition/partition.conf
@@ -106,9 +106,15 @@ defaultFileSystemType:  "ext4"
 #       mountPoint: "/home"
 #       size: 3G
 #       minSize: 1.5G
+#       features:
+#         64bit: false
+#         casefold: true
 #     - name: "data"
 #       filesystem: "fat32"
 #       mountPoint: "/data"
+#       features:
+#         sector-size: 4096
+#         sectors-per-cluster: 128
 #       size: 100%
 #
 # There can be any number of partitions, each entry having the following attributes:
@@ -120,6 +126,8 @@ defaultFileSystemType:  "ext4"
 #           % of the available drive space if a '%' is appended to the value
 #   - minSize: minimum partition size (optional parameter)
 #   - maxSize: maximum partition size (optional parameter)
+#   - features: filesystem features (optional parameter; requires KPMCore >= 4.2.0)
+#       name: boolean or integer or string
 
 # Checking for available storage
 #
--- a/src/libcalamares/CMakeLists.txt
+++ b/src/libcalamares/CMakeLists.txt
@@ -128,7 +128,12 @@ if ( KPMcore_FOUND )
     find_package( Qt5 REQUIRED DBus )  # Needed for KPMCore
     find_package( KF5 REQUIRED I18n WidgetsAddons )  # Needed for KPMCore
 
-    if( KPMcore_VERSION VERSION_GREATER_EQUAL "4.0" )
+    if( KPMcore_VERSION VERSION_GREATER_EQUAL "4.2" )
+        add_definitions(
+            -DWITH_KPMCORE42API
+            -DWITH_KPMCORE4API
+        ) # kpmcore 4.2 with new API
+    elseif( KPMcore_VERSION VERSION_GREATER_EQUAL "4.0" )
         add_definitions( -DWITH_KPMCORE4API ) # kpmcore 4 with new API
     elseif( KPMcore_VERSION VERSION_GREATER "3.3.70" )
         message( FATAL_ERROR "KPMCore beta versions ${KPMcore_VERSION} not supported" )
--- a/src/modules/fsresizer/CMakeLists.txt
+++ b/src/modules/fsresizer/CMakeLists.txt
@@ -8,7 +8,9 @@ set( _partition_defs "" )
 if ( KPMcore_FOUND AND Qt5DBus_FOUND AND KF5CoreAddons_FOUND AND KF5Config_FOUND )
     include_directories( ${KPMCORE_INCLUDE_DIR} ${CMAKE_SOURCE_DIR}/src/modules/partition )
 
-    if( KPMcore_VERSION VERSION_GREATER_EQUAL "4.0" )
+    if( KPMcore_VERSION VERSION_GREATER_EQUAL "4.2" )
+        list( APPEND _partition_defs WITH_KPMCORE42API WITH_KPMCORE4API ) # kpmcore 4.2 with new API
+    elseif( KPMcore_VERSION VERSION_GREATER_EQUAL "4.0" )
         list( APPEND _partition_defs WITH_KPMCORE4API ) # kpmcore 4 with new API
     elseif( KPMcore_VERSION VERSION_GREATER "3.3.70" )
         message( FATAL_ERROR "KPMCore beta versions ${KPMcore_VERSION} are not supported" )
--- a/src/modules/partition/CMakeLists.txt
+++ b/src/modules/partition/CMakeLists.txt
@@ -33,6 +33,9 @@ if ( KPMcore_FOUND AND Qt5DBus_FOUND AND KF5CoreAddons_FOUND AND KF5Config_FOUND
     if ( KPMcore_VERSION VERSION_GREATER "3.90")
         list( APPEND _partition_defs WITH_KPMCORE4API) # kpmcore 4 with new API
     endif()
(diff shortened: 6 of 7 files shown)


src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -239,6 +241,26 @@ PartitionLayout::execute( Device* dev,
...
+                {
+                case QMetaType::Bool: 
+                    currentPartition->fileSystem().addFeature( { k, v.toBool() } );
+		    break;
+                case QMetaType::LongLong:
+                    currentPartition->fileSystem().addFeature( { k, v.toInt() } );
+		    break;
+                case QMetaType::QString:
+                    currentPartition->fileSystem().addFeature( { k, v.toString() } );
+                    break;
```
adriaandegroot: I think this should have a `default:` for all the cases not-handled (possibly logging that the configuration looks broken).

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -144,6 +145,7 @@ PartitionLayout::addEntry( const QString& label,
     {
         entry.partFileSystem = m_defaultFsType;
     }
+    entry.partFeatures = features;
```
adriaandegroot: This suggests to me that *features* should be part of the *PartitionEntry* constructor.

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -239,6 +241,26 @@ PartitionLayout::execute( Device* dev,
         {
             currentPartition->fileSystem().setLabel( part.partLabel );
         }
+        if ( !part.partFeatures.isEmpty() )
+        {
+            for ( const auto &k : part.partFeatures.keys() )
+            {
+                const auto &v = part.partFeatures.value(k);
```
adriaandegroot: Nitpick: `const auto& v` is the official-Calamares-style for that. (You can use ci/calamaresstyle to apply astyle+clang-format with the right bits set)

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -239,6 +241,26 @@ PartitionLayout::execute( Device* dev,
...
+
+                switch (v.type())
+                {
+                case QMetaType::Bool: 
+                    currentPartition->fileSystem().addFeature( { k, v.toBool() } );
```
adriaandegroot: Since `addFeature()` is a KPMCore 4.1.0 API addition, this should be properly `#ifdef`fed. Calamares is compatible with KPMCore 3 and 4.0 and .. so this can't be used just-like-that.

src/modules/partition/partition.conf
```diff
@@ -120,6 +126,8 @@ defaultFileSystemType:  "ext4"
 #           % of the available drive space if a '%' is appended to the value
 #   - minSize: minimum partition size (optional parameter)
 #   - maxSize: maximum partition size (optional parameter)
+#   - features: filesystem features (optional parameter)
+#       name: boolean or integer or string
```
adriaandegroot: "This is only available with KPMCore 4.1.0 and later" should be a documentation somewhere.

src/modules/partition/core/PartitionCoreModule.cpp
```diff
@@ -881,12 +882,18 @@ PartitionCoreModule::initLayout( const QVariantList& config )
             maxSizeString = CalamaresUtils::getString( pentry, "maxSize" );
         }
 
+        if ( pentry.contains( "features" ) )
+        {
+            featureMap = pentry.values( "features" ).at(0).toMap();
```
adriaandegroot: You might want to use just `value()` there, or also `CalamaresUtils::getSubMap()` -- the latter is supposed to be a convenience function, although I don't know how convenient it is.

src/modules/partition/core/PartitionCoreModule.cpp
```diff
@@ -881,12 +882,18 @@ PartitionCoreModule::initLayout( const QVariantList& config )
             maxSizeString = CalamaresUtils::getString( pentry, "maxSize" );
         }
 
+        if ( pentry.contains( "features" ) )
```
adriaandegroot: Since features are only available to KPMCore 4.1.0 and later, this should have an `#ifdef`fed `cWarning()` if the configuration isn't going to have an effect.

src/modules/partition/core/PartitionLayout.cpp
gportay: I added the default  case; it says:

```
15:49:14 [2]: WARNING: Ignoring feature "casefold" of type QVariant::QVariantList ; requires type QVariant::bool, QVariant::qlonglong or Variant::QString. 
```

src/modules/partition/core/PartitionLayout.cpp
gportay: Oops, my bad. Fixed!

src/modules/partition/core/PartitionLayout.cpp
gportay: Done.04

src/modules/partition/core/PartitionCoreModule.cpp
gportay: I used `getSubMap()`. I change the code a bit to remove the `if` statement.

src/modules/partition/partition.conf
gportay: Done.

src/modules/partition/core/PartitionLayout.cpp
gportay: Well, I guess this should apply to the attributes `partLabel`, `partMountPoint` and `partFileSystem`; but if so, the structure `PartitionEntry` is unless.

I preferred to left it as is. I guess it can be done in a second commit.

gportay:
Merci @adriaandegroot for the review; you have responded to all the questions I wanted to ask.

I have marked the point as resolved; tell me if I did it wrong. But, one conversation remains.

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -239,6 +241,33 @@ PartitionLayout::execute( Device* dev,
...
+                    currentPartition->fileSystem().addFeature( { k, v.toInt() } );
+		    break;
+                case QMetaType::QString:
+                    currentPartition->fileSystem().addFeature( { k, v.toString() } );
+                default:
+                    cWarning() << "Ignoring feature" << k << "of type" << v.type() << ";"
+                               << "requires type QVariant::bool, QVariant::qlonglong or QVariant::QString.";
+                    break;
+                }
+            }
```
gportay: Also, I wonder if I can make a change in KPMcore to replace the class [FSFeature](https://github.com/KDE/kpmcore/commit/d24191ebd8ee4a1792003e5bd280cb7679f5e834#diff-4a568e881c09c1f8b464ebde9506d94d) by QVariant; I would simplify the code to the simple call `currentPartition->fileSystem().addFeature( v );`

gportay:
Sorry, I have to fix indentation violations...

src/modules/partition/core/PartitionLayout.cpp
adriaandegroot: @stikonas opinions about this suggestion for API? That would be a KPMCore 4.2, I guess (or 4.1.1 if you add it, rather than replacing the existing API).

src/modules/partition/core/PartitionLayout.cpp
adriaandegroot: Don't worry about this one: it applies polishing to other code, and doesn't have to be part of this MR. If you have time in your project, it's probably worth chasing some technical debt (like this one) but otherwise leave it and I'll get around to it eventually (that's what `TODO:` is for, right?)

src/modules/partition/core/PartitionLayout.cpp
andriuss-xilinx: If you can keep it ABI compatible, it's fine to go to 4.1.1. See
https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

Otherwise, it should go to 4.2 (and SOVERSION needs a bump.)

stikonas: If you can keep it ABI compatible, it's fine to go to 4.1.1. See
https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

Otherwise, it should go to 4.2 (and SOVERSION needs a bump.)

andriuss-xilinx:
Does this compile with https://invent.kde.org/kde/kpmcore/-/merge_requests/6/diffs ?

stikonas:
Does this compile with https://invent.kde.org/kde/kpmcore/-/merge_requests/6/diffs ?

gportay:
@stikonas, @adriaandegroot I made a [MR](https://invent.kde.org/kde/kpmcore/-/merge_requests/6) about the change of the API.

If it is accepted, it will bump the KPMcore to `4.2.0` and soversion to `10`. See my fixup [https://github.com/gportay/calamares/commit/df89e6625fce7ae08b5cc60ee0caafe25e97b94e#diff-f5fe5f5f221a5ff09da3821dfea0c449](commit) on my branch `wip`.

Sorry, I have lost the thread. I do not know how GitHub works... it seems that force-pushing lost the reviews or make them unreachable :(

stikonas:
> @stikonas, @adriaandegroot I made a [MR](https://invent.kde.org/kde/kpmcore/-/merge_requests/6) about the change of the API.
> 
> If it is accepted, it will bump the KPMcore to `4.2.0` and soversion to `10`. See my fixup [https://github.com/gportay/calamares/commit/df89e6625fce7ae08b5cc60ee0caafe25e97b94e#diff-f5fe5f5f221a5ff09da3821dfea0c449](commit) on my branch `wip`.
> 
> Sorry, I have lost the thread. I do not know how GitHub works... it seems that force-pushing lost the reviews or make them unreachable :(

That's expected. Force-pushing removes reference from the old commits, so git considers them eligible for garbage collection. It's not really a Github issue. Just github doesn't try to workaround them. You can keep old commits by adding some kind of reference to previous revision (branch, tag, or some custom git reference).

gportay:
> That's expected. Force-pushing removes reference from the old commits, so git considers them eligible for garbage collection. It's not really a Github issue. Just github doesn't try to workaround them. You can keep old commits by adding some kind of reference to previous revision (branch, tag, or some custom git reference).

Thanks for that fast response.

kkofler:
I still see the old comments (for the commits replaced by the force push) on the pull request web page (https://github.com/calamares/calamares/pull/1341), they are just hidden behind a "Show resolved" link.

gportay:
> I still see the old comments (for the commits replaced by the force push) on the pull request web page (#1341), they are just hidden behind a "Show resolved" link.

Indeed, thanks for the tip!

stikonas:
> I still see the old comments (for the commits replaced by the force push) on the pull request web page (#1341), they are just hidden behind a "Show resolved" link.

Well, old comments, and snippets of code around them are available. But the old commits themselves might be gardened by git.

src/modules/partition/core/PartitionLayout.cpp
gportay: I can do it in another commit ;)

Furthermore, I would like to rename things. Like removing the `part` prefixes from the attributes that are redundant: `PartitionLayout::PartitionEntry::partLabel`. The `Partition` from `PartitionEntry` is redundant as well.

Also, the attribute `QString partLabel` should also be renamed to `QString partName` as it refers to the YAML attribute `PartitionLayout.name` in the file `partition.conf`.

gportay:
I have update de PR since the [MR](https://invent.kde.org/kde/kpmcore/-/merge_requests/6) is merged upstream.

I assume the next version of KPMcore is `4.2.0`, but I may be wrong.

stikonas:
> I assume the next version of KPMcore is `4.2.0`, but I may be wrong.

That's almost certainly the case. If I need to release a bugfix release, I'll branch from earlier point before soversion bump. Only in the very unlikely case of Qt6, and KF6 releases coming first, it will be KPMcore 5.0.0.

PR #1357, opened 2020-03-25
Increase Manual Partition instructions

demmm:


apt-ghetto:
@adriaandegroot 
Is it possible to revert this commit?

- It is shown also in BIOS mode.
- The mountpoint is distro specific and defined with `efiSystemPartition`.
- It is shown also without GPT. And why is GPT even mentioned?
- It is the wrong place to explain, what can go wrong with manual partitioning. It should only explain the option, not a little part of all possible problems.

PR #1364, opened 2020-04-01
About option re-added

demmm:


demmm:
Welcomeq page look with this PR:
Screenshot_20200401_160840
Screenshot_20200401_160700

PR #1368, opened 2020-04-04
[localeq] Repeater with zonesModel crashes cala

demmm:
probably due to dynamically loading items
regionModel now lists, zonesModel only lists one delegate, but
working on QML modules can now continue without crashing cala

### DIFF
--- a/src/modules/localeq/localeq.qml
+++ b/src/modules/localeq/localeq.qml
@@ -27,9 +27,9 @@ RowLayout
         Layout.fillWidth: true
         ColumnLayout {
             id: zones
-            Repeater {
+            ListView {
                 model: config.zonesModel
-                Text {
+                delegate: Text { 
                     text: label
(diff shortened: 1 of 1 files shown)


demmm:
State of loacaleq with this PR:
Screenshot_20200404_184704

adriaandegroot:
No-crashes (but ugly) is preferable to crashes (and also still ugly). I imagine the previous QML code was better in this regard. It's probably worth looking at it (somewhere inside the ResponsiveBase and other stuff that isn't available).

demmm:
If you'd like something a bit prettier and no crashes now, I can do that.  That means layout first, getting the regionsModel & zonesModel to display fully second.
The original localeq.qml works with a stack view, I don't see that as needed, prefer to implement a version with the marble widget instead (should be in a very old bug report/request for cala.

PR #1371, opened 2020-04-09
[welcomeq] add languagebar

demmm:
data filled by languagesModel
better spacing/layout buttons


demmm:
Current state of welcomeq and localeq:

Screenshot_20200409_122604
Screenshot_20200409_122647

demmm:
State of localeq as of April 14, converting country/city names to coordinates works, pin moving to new, user location, map centering on new pin location, Map.qml connected to locale data, though regions aren't supported in QtLocation, so coordinates are wrong, start of adding text for the map:
Screenshot_20200414_174640

src/modules/localeq/Map.qml
```diff
@@ -101,7 +114,13 @@ Column {
                     marker.coordinate = coordinate
                     map.center.latitude = coordinate.latitude
                     map.center.longitude = coordinate.longitude
-                    console.log(coordinate.latitude, coordinate.longitude)
+                    
+                    geocodeModel.query = QtPositioning.coordinate(coordinate.latitude, coordinate.longitude)
```
demmm, line 118: missed in commit message, this is a start of implementing ReverseGeocoding, console log shows output, but translated to degrees, not text yet

src/modules/localeq/localeq.qml
```diff
@@ -119,9 +79,18 @@ Page
...
+                        onClicked: load.source = "i18n.qml"
+                    }
                 }
             }
         }
+        
+        
+        
     }
+    Loader {
```
demmm, line 92: Current state of localeq with this commit:
Screenshot_20200418_171001
Screenshot_20200418_171021

Issue #1222, opened 2019-08-14
[Support Question] How to change default partition scheme?


kevinsmia1939:
I want to modify the "Erase Disk" option when partitioning to be divided into root and home which is both btrfs.
I modified partition.conf in usr/share/calamares/modules to look like this:

partitionLayout:
     - name: "root"
       filesystem: "btrfs"
       mountPoint: "/"
       size: 20%
     - name: "home"
       filesystem: "btrfs"
       mountPoint: "/home"
       size: 80%
     
But no option in partitioning in calamares show this at all. Did I miss understand partition.conf?  
Thanks.

abucodonosor:
There is no UIX support for this. It will do what you set partitionLayout.
Please note, with btrfs enabled calamares is adding @home already.

kkofler:
To be clear, Calamares automatically adds `/home` as a subvolume on `btrfs`. The logic creating that subvolume is hardcoded. On the other hand, the configuration @kevinsmia1939 posted would, if it worked, create it as a separate partition. I suspect it will not work as intended because it will conflict with the hardcoded subvolume creation logic.

kevinsmia1939:
> To be clear, Calamares automatically adds `/home` as a subvolume on `btrfs`. The logic creating that subvolume is hardcoded. On the other hand, the configuration @kevinsmia1939 posted would, if it worked, create it as a separate partition. I suspect it will not work as intended because it will conflict with the hardcoded subvolume creation logic.

Will the hard code become configurable anytime soon?

PR #1375, opened 2020-04-14
 Add subvolumes @cache and @log to the btrfs configuration 

Chrysostomus:
Most useful feature of a btrfs root on desktop system is snapshotting. Mounting /var/cache and /var/log on separate subvolumes excludes them from snapshots. This desirable, because

- Otherwise restoring a snapshot will cause you to lose your logs which detail the situation that caused you to need to roll back to snapshot
- /var/cache contains a lot of rapidly changing and low value data such as package manager cache. Including it in snapshots causes the snapshots to bloat rapidly and unnecessarily.

These are part of the subvolume layout that is recommended by the Arch wiki.


abucodonosor:
Nope, sorry that has no place in calamares like this.
I fact what we do on @home should be removed that is broken too.

If someone wants such a setup he'll for sure add needed subvolumes, forcing this on everyone is a no-no, unless you come up with a GUI for all this so users can choose what they want.

Chrysostomus:
In what way is the current system broken? Having @ and @home subvolumes is the common standard for flat subvolume structure, and some applications like timeshift (shipped with manjaro by default) do not even work with other subvolume structures. Removing the @home subvolume would make calamares practically unviable installer for btrfs systems, considering that calamares doesn't allow altering the subvolume structure. Forcing people to a nonstandard setup that causes personal data loss on rollback is not an improvement. 

I agree that being able to manually select the subvolume structure would be much better of course. Second best option would be to have distro-configurable subvolume structure.

abucodonosor:
> In what way is the current system broken? Having @ and @home subvolumes is the common standard for flat subvolume structure, and some applications like timeshift (shipped with manjaro by default) do not even work with other subvolume structures. Removing the @home subvolume would make calamares practically unviable installer for btrfs systems, considering that calamares doesn't allow altering the subvolume structure. Forcing people to a nonstandard setup that causes personal data loss on rollback is not an improvement.

There is NOTHING like a standard for <put_here_random_fs_you_wish>. 
It is all a user choice, it always was and it always will be.

TBH it doesn't matter what <random distribution> installs by default or not, besides
subvolumes can be managed after installation too, or have a distribution tool doing that on post-install.

Why is wrong? Well obvious we force stuff on people and since we have now a profile
based fs layout that badly conflicts too. See #1222 as an example.


> I agree that being able to manually select the subvolume structure would be much better of course. Second best option would be to have distro-configurable subvolume structure.

Feel free to figure something configurable on top fs layout.

Chrysostomus:
Very well, I'll look into this.

philmmanjaro:
Looks good to me. Based on our current internal discussions we will bring those features also to upstream. More Arch-based Distros might take use of it. @abucodonosor if it is not working for you it might work for others. I will keep a closer look on issues or MRs closed when not discussed more openly. Will reopen it now to see what might need to be changed, as Manjaro and others might switch to BTRFS as default file system. That means that we will port more of our distro specific modules to the upstream code for easier maintenance.

philmmanjaro:
@Chrysostomus we might need to make it more configurable as some distros might not use this feature as we might do. Let's see what @adriaandegroot and others may say.

abucodonosor:
> Looks good to me. Based on our current internal discussions we will bring those features also to upstream. More Arch-based Distros might take use of it. @abucodonosor if it is not working for you it might work for others. 

Then make it configurable if your Arch based distro is making use of such a thing,
as-is is not acceptable, period.

Besides, I never suggested that to be closed.

>I will keep a closer look on issues or MRs closed when not discussed more openly. 

So what kind of discussion do you want?

>Will reopen it now to see what might need to be changed, as Manjaro and others might switch to BTRFS as default file system. That means that we will port more of our distro specific modules to the upstream code for easier maintenance.

Yes, that is the keyword *distro-specific*. While so, why do you think it is acceptable to force that on everyone?

abucodonosor:
> @Chrysostomus we might need to make it more configurable as some distros might not use this feature as we might do. Let's see what @adriaandegroot and others may say.

There is really no 'we might' there is for sure a 'we have too'.

abucodonosor:
Oh, @philmmanjaro and before you give bogus ACKs you should really review the code right?

Chrysostomus:
When I originally opened the pull request it was not yet ready to be merged. I was thinking we might have a conversation about this feature and how to best implement it. First I wasn't sure if this would be best to implement on distro level or upstream, but @adriaandegroot said he might be interested in upstreaming it, so I figured it would be best to move the conversation here. 

I got some good ideas from the conversation, but I also figured that it was probably yet too incomplete, so I closed it for now. I'm currently working on implementation that allows setting the subvolumes with a configuration file, so each distro can have their own settings. But it's going to take a while before it is ready. 

For upstream it is understandable that the solution should be generic.

Chrysostomus:
I'm sorry if I haven't followed the proper protocol with this pull request. My intention was not to cause strife.

abucodonosor:
> I'm sorry if I haven't followed the proper protocol with this pull request. My intention was not to cause strife.

Don't worry, everything is fine.

Your current code has some typos and will break with luks and if someone want
to use such as fs layout. But if you really need that code right now upstreamed, with the hardcoded names of @cache & @log etc we can make it work but we need some option to have if off by default. So you have time to implement something else but still have a way to enable that for now.

Ideally, we should not need any hardcoded names, that include @home, instead of an option like

```
btrfs_create_volumes: a b c d e f g

```

Which then is ignored in case FS != BTRFS or ignored if not set.

Chrysostomus:
I think it is better not to hurry with this. The generic implementation is better for manjaro too, because then we can have more specific mount points (/var/cache/pacman/pkg) instead of using a general solution like /var/cache.

abucodonosor:
@Chrysostomus sure, as you wish.

Chrysostomus:
Okay, getting a little closer now... I'll start testing if this thing actually runs to find what other typos I made...

Conan-Kudo:
Note that Fedora's Btrfs layout is similar, but different from Ubuntu, and quite different from openSUSE. We default to `root` (for `/`), `home` (for `/home`) and eventually `boot` (for `/boot`) top level subvolumes. We also permit custom names during installation with Anaconda, and I'd expect the same in Calamares.

adriaandegroot:
This bears directly also on #1544 (and has the same kind of issues -- expanding hardcoded lists, rather than introducing configuration and a bit of software architecture)

PR #1380, opened 2020-04-18
[branding] add documentation for WIP 'productWallpaper' branding

bill-auger:


adriaandegroot:
The *productWallpaper* is mostly added for QML (but isn't used in any of the QML we have), because it gives an easy way to offer a generic background -- this has the same kinds of trade-offs as more configurations of the welcome page: in QML you could just hard-code stuff and dump it in the branding directory, but it's nice if we have a generic solution that also follows useful configuration variables.

**Really** this PR is a hint that there needs to be an issue "widgets pages do not support *productWallpaper* settings"

bill-auger:
food for though

PR #1381, opened 2020-04-18
[welcome] add optional branding banner to welcome page

bill-auger:


bill-auger:
i also have made another patch which makes the height customizible - it follows the plumbing of 'languageIcon' and the others, with their getters and setters in the Config class, and the special handling in the post-instatiatoin init() method, modifying all of these files:

 src/modules/welcome/Config.cpp
 src/modules/welcome/Config.h
 src/modules/welcome/WelcomePage.cpp
 src/modules/welcome/WelcomePage.h
 src/modules/welcome/WelcomeViewStep.cpp
 src/modules/welcome/welcome.conf

IMHO, that is very messy and over-kill for the small benefit - i am not planning to carry that patch; but i will add it to this PR if you ask - that, plus a CSS ID would be necessary for the CSS customization though

src/modules/welcome/WelcomePage.cpp
```diff
@@ -77,11 +77,25 @@ WelcomePage::WelcomePage( Config* conf, QWidget* parent )
...
-        2 * QSize( CalamaresUtils::defaultFontHeight(), CalamaresUtils::defaultFontHeight() ) ) );
+        2 * QSize( defaultFontHeight, defaultFontHeight ) ) );
     connect( ui->aboutButton, &QPushButton::clicked, this, &WelcomePage::showAboutBox );
 
+    // insert system-check widget below welcome text
     int welcome_text_idx = ui->verticalLayout->indexOf( ui->mainText );
     ui->verticalLayout->insertWidget( welcome_text_idx + 1, m_checkingWidget );
+
+    // insert optional logo banner image above welcome text
+    Calamares::Branding::ImageEntry bannerImage = Calamares::Branding::ProductBanner;
```
adriaandegroot, line 88: I don't see the point here: you only use this enum value once, and the line doesn't get too long.

src/modules/welcome/WelcomePage.cpp
```diff
@@ -77,11 +77,25 @@ WelcomePage::WelcomePage( Config* conf, QWidget* parent )
...
+    QString bannerPath = Calamares::Branding::instance()->imagePath( bannerImage );
```
adriaandegroot, line 89: I think you should add a `if (!bannerPath.isEmpty()) {}` around the next block, if only to avoid typing-to-load-a-QImage that is obviously bogus.

src/libcalamaresui/Branding.h
```diff
@@ -67,10 +67,11 @@ class UIDLLEXPORT Branding : public QObject
 
     enum ImageEntry : short
     {
-        ProductLogo,
+        ProductBanner,
```
adriaandegroot, line 70: I get the impression there's a corrsponding `Branding.cpp` change that is missing here.

src/branding/default/branding.desc
```diff
@@ -106,6 +106,11 @@ strings:
 
 # These images are loaded from the branding module directory.
 #
+# productBanner is an optional image, which if present, will be shown
+#       on the welcome page of the application, above the welcome text.
+#       It is intended to have a width much greater than height.
+#       It is displayed at 64px height (also on HiDPI).
```
adriaandegroot, line 112: The code says `64 + 2 * defaultFontHeight` which is probably 88 in total?

adriaandegroot:
I think for more welcome-page customization we should turn to QML -- if the C++ code becomes too convoluted, it's just time to use a different technology that allows you to just chuck a text file into the branding directory and have that work. @demmm has done really good work on the example QML there (I still need to merge it) and it turns into "you want another image in your welcome page? Add an `Image {}` to the QML file."

There's more percentage in that -- even if we do need to do some extra work on Config (I'm torn there: a custom QML can pull in images from wherever, but it might be nice to offer generic solutions that can be modified through the `welcome.conf` file)

demmm:
>a custom QML can pull in images from wherever, but it might be nice to offer generic solutions that can be modified through the welcome.conf file) 

You can easily have both, set to load `Custom.qml` page anywhere in the main QML page, define in welcome.conf path to `Custom.qml` (it can be as simple as just an Image qml class).

src/modules/welcome/WelcomePage.cpp
bill-auger: i dont understand the dispute - the LOC referencing 'bannerImage' is nearly 90 chars now, even with the variable substituted

src/libcalamaresui/Branding.h
bill-auger: yes there is - i forgot to include it - i re-implemented this feature by hand onto the master branch without building it, rather than cherry-picking the buried commits

src/branding/default/branding.desc
bill-auger: your right - that exposed a think-o - i intended (2 * defaultFontHeight) as padding - my image actually is 64 px - i should have surrounded it with explicit "spacer" widgets

src/modules/welcome/WelcomePage.cpp
bill-auger: the problem with checking the string, is that the image file may not be present, regardless of whether the branding key is defined or not - so if not checking for a null image, it would need instead to check for a nonexistent file, before instantiating an image

src/modules/welcome/WelcomePage.cpp
```diff
@@ -62,7 +61,8 @@ WelcomePage::WelcomePage( Config* conf, QWidget* parent )
              &CheckerContainer::requirementsProgress );
     ui->setupUi( this );
 
-    ui->verticalLayout->insertSpacing( 1, CalamaresUtils::defaultFontHeight() * 2 );
+    int defaultFontHeight = CalamaresUtils::defaultFontHeight();
```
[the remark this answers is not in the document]
bill-auger: ok - i also added const to the 'welcome_text_idx' declaration, accordingly

bill-auger:
if the plan is put (or migrate) everything under the QML/CSS system, i dont know what that entails - this is the exact patch that i have ben carrying - its just the most KISS approach, which does the job - that is the reason why i hard-coded the height, and did not include the next patch which makes it customizable, though i did implement it, anticipating that it may be requested

generally, i would not want to over-implement without a use-case - i have no use for the QML/CSS system myself; so i would need to become familiar with it, if thats the direction calamares is headed - maybe put this one on hold to see what demmm comes up with

bill-auger:
if QML would allow the welcome page to be fully customizable, like the slide-show; thats probably for the best - after adding this banner with the default calamares code, the layout is quite bad by default, at the smaller end of the window size constraints - at the smallest allowed window size, the language image nearly disappears to smaller than my mouse cursor

i also have these adjustments, which makes the layout fit small screens much better:

```
+            // adjust layout
+            QString defaultFontFamily = CalamaresUtils::defaultFont().defaultFamily();
+            ui->mainText->setFont( QFont( defaultFontFamily, 8, QFont::Bold ) );
+            ui->verticalLayout->removeItem( ui->verticalSpacer );
+            ui->verticalLayout->removeItem( ui->verticalSpacer_4 );
```

src/modules/welcome/WelcomePage.cpp
adriaandegroot: That makes sense -- my instinct was "if it's not set, don't do anything" but of course if it's not set, then the image-loading bits won't trigger and your code already does the right thing there. So it's good.

PR #1382, opened 2020-04-18
[i18n] esperanto translations

bill-auger:


kkofler:
I thought translations have to go through Transifex, don't they?

bill-auger:
ive sent them as a patch once before - ade tried it last time and
noticed that it is equivalent

adriaandegroot:
The preferred approach is through Transifex, indeed -- that means that translators can work together and updates are reviewed. Patches **also** work, though:

 - merge patch to Calamares via GH
 - git pull on development machine
 - log in to Transifex
 - Resources -> (pick the resource that was updated in the merge) -> pick language -> *upload translations*
 - Pick the newly-patched .ts file and click ok

It's more work for me, along the lines of "two minutes of button clicking", but that's fine as long as not all of the translations start doing that. Esperanto is special, anyway.

bill-auger:
pk, i was under the crazy presumption that the translation
service had git integration, and that direct commits into the
master repo would actually save you some effort - i will use the
web thing in the future

PR #1386, opened 2020-04-21
Add support for bash-completion

gportay:


adriaandegroot:
This seems like something **amazingly** specialised which I wouldn't want to inflict on distro's under normal circumstances. I'm going to munge it somewhat and merge the modified form.

adriaandegroot:
Merged as ded923fb9545b8191ae84a091e3fdbabcc6bde5a

PR #1389, opened 2020-04-21
[partition] Update debug output

gportay:
The mount-point is known as mountPoint in the global storage.

This commit updates the trace to reflect its real name.


adriaandegroot:
True that (I remember being misled by this at least once)

PR #1394, opened 2020-05-05
[bootloader] Set the EFI Boot Next

gportay:
- This forces the EFI firmware to boot the loader that was just
   created, whatever the boot order set in the firmware setup (USB,
   CD/DVD, HD...).
 - It is safe to use the first Boot Entry listed in BootOrder as the
   previous command creates the new entry and adds it to the first place
   of the BootOrder.

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -403,6 +403,21 @@ def install_secureboot(efi_directory):
         "-p", efi_partition_number,
         "-l", install_efi_directory + "/" + install_efi_bin])
 
+    boot_entry = None
+    efi_bootvars = subprocess.check_output(
+        [libcalamares.job.configuration["efiBootMgr"]], text=True)
+    for line in efi_bootvars.split('\n'):
+        if not line:
+            continue
+        words = line.split()
+        if len(words) >= 2 and words[0] == "BootOrder:":
+            boot_entry = words[1].split(',')[0] 
+            break
+    if boot_entry:
+        subprocess.call([
+            libcalamares.job.configuration["efiBootMgr"],
+            "-n", boot_entry])
+
     # The input file /etc/default/grub should already be filled out by the
     # grubcfg job module.
     check_target_env_call([libcalamares.job.configuration["grubMkconfig"],


src/modules/bootloader/main.py
```diff
@@ -403,6 +403,22 @@ def install_secureboot(efi_directory):
         "-p", efi_partition_number,
         "-l", install_efi_directory + "/" + install_efi_bin])
 
+    boot_entry = None
+    ret = subprocess.run([libcalamares.job.configuration["efiBootMgr"]],
```
abucodonosor, line 407: subprocess.run() is Python >=3.5, we want to stay compatible with Python 3.3, so you cannot use it.

abucodonosor:
> TBH, I have tested that code in another home-made module.
> 
> This is the first time I write things in python, thus I am opened to better solutions to parse the output.

Is not about a better solution, .run() is no available for Py 3.3, so maybe use
subprocess.check_output() ?

gportay:
> Is not about a better solution, .run() is no available for Py 3.3, so maybe use
subprocess.check_output() ?

Ah ah, comment crossing :) I was not replying to your comment.

I will have a look at `check_output`. Thanks

src/modules/bootloader/main.py
gportay: So I guess, the code should be:

```
    boot_entry = None
    efi_entries = subprocess.check_output([libcalamares.job.configuration["efiBootMgr"]], text=True)
    for line in efi_entries.split('\n'):
        if not line:
            continue
        words = line.split()
        if len(words) >= 2 and words[0] == "BootOrder:":
            boot_entry = words[1].split(',')[0] 
            break
    if boot_entry:
        subprocess.call([
            libcalamares.job.configuration["efiBootMgr"],
            "-n", boot_entry])
```

adriaandegroot:
Merged and then slightly refactored, thanks. I do see that it is **only** called in the secureboot path -- have you considered also running it from the other installation (systemd, grub) paths when EFI is in use? I moved the relevant code to a function to make that easier to re-use.

PR #1398, opened 2020-05-08
start of adding requirements to welcomeq

demmm:
just QML, no real connections yet
hardcoded to not visible
not clear if requirementsModel is available to welcomeq


demmm:
Screenshot when Recommended switch is set to not match:
Screenshot_20200508_132015

PR #1404, opened 2020-05-12
Update calamares_fa.ts

MRazian:
This update includes 115 translated strings.

### DIFF
--- a/lang/calamares_fa.ts
+++ b/lang/calamares_fa.ts
@@ -1,22 +1,20 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!DOCTYPE TS>
-<TS language="fa" version="2.1">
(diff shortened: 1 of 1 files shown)


adriaandegroot:
I'm vaguely amused by translations like "پارتیشن سیستمی" because that's a transliteration more than anything else. Is there no word like قسم in Farsi? (No, of course not, it's indo-european, not semitic)

Regardless of my amusement in Farsi (I only just got a 1962 copy of *teach yourself modern Persian* and it probably won't mention computers) I've uploaded your translations to Transifex, then merged translations into Calamares. Thank you for helping out! If you can build Calamares from source, then you can check that the translations have survived the round-trip.

I'm closing this PR **without merging** because the workflow for translations is different. Thanks again! Or تشكر, as the case may be.

adriaandegroot:
(To be perfectly clear: closed, because the new translations were imported into Transifex)

PR #1415, opened 2020-05-22
[partition] Partition table type settings

gportay:
Hello,

This PR ~~is an _RFC_ to~~ restricts the installation on disks that use a specified partition table.

For example, since the `systemd` 211and the service [systemd-gpt-auto-generator](https://www.freedesktop.org/software/systemd/man/systemd-gpt-auto-generator.html), it is possible to discover and mount various partitions based on the _GPT_ [type](https://www.freedesktop.org/wiki/Specifications/DiscoverablePartitionsSpec/).

Therefore, we want to make sure the operating system is installed on a _GPT_ disk.

For now, the option **Erase disk** creates a new _GPT_ partition table if the booted system is _EFI_; otherwise, it creates an _MBR/MSDOS_ partition table (legacy _BIOS_).

The options **Replace a partition** and **Install alongside** left untouched the partition table type that exists on disk.

The option **Manual partitioning** may reuse the existing partition or let the user pick up the one he wants.

This PR introduces the two new settings in `partition.conf`:
- `defaultPartitionTableType` that forces the use of a partition table type, whatever if the firmware (_EFI_ or _BIOS_); see the first commit. That setting takes a string.
- `requiredPartitionTableType` that specifies the type of partition table on which the operating system can be installed. If the disk partition type does not match, the options **Replace a partition**, **Install alongside** and **Manual partitioning** are hidden; only the option **Erase disk** is made available to the user; see the second commit. That setting takes a string or a list of strings.

Example:

``` yaml
defaultPartitionTableType: gpt
requiredPartitionTableType:
  - msdos
  - gpt
```

This one forces the use of GPT:
``` yaml
defaultPartitionTableType: gpt
requiredPartitionTableType: gpt
```

Both commits are independent.

This is out of scope, but I have noticed that the UI label message is hidden since 48d0de2e08d66d5c9f38b87fed66512f241f79bb. I have reused it in case of the partition table type mismatch.

But if that element is not used anymore, we can remove it entirely that will simplify the [code](https://github.com/calamares/calamares/blob/69c2d089f29338b0d48724a092c82149721bd091/src/modules/partition/gui/ChoicePage.cpp#L1266-L1370).

### DIFF
--- a/src/modules/partition/core/Config.cpp
+++ b/src/modules/partition/core/Config.cpp
@@ -239,6 +239,20 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
     Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
     gs->insert( "allowManualPartitioning",
                 CalamaresUtils::getBool( configurationMap, "allowManualPartitioning", true ) );
+
+    if ( configurationMap.contains( "requiredPartitionTableType" ) &&
+        configurationMap.value( "requiredPartitionTableType" ).type() == QVariant::List )
+    {
+        m_requiredPartitionTableType.clear();
+        m_requiredPartitionTableType.append( configurationMap.value( "requiredPartitionTableType" ).toStringList() );
+    }
+    else if ( configurationMap.contains( "requiredPartitionTableType" ) &&
+        configurationMap.value( "requiredPartitionTableType" ).type() == QVariant::String )
+    {
+        m_requiredPartitionTableType.clear();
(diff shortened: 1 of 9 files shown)


gportay:
squash fixup and rebased on to branch `calamares`.

gportay:
Rebased on `calamares/calamaces`.

adriaandegroot:
Sorry, both your PRs now clash with the swapfile work; if you don't get around to it, I'll fix and merge these on friday.

PR #1420, opened 2020-05-28
Update SPDX License Identifier

Cal8031:
Use of GPL-3.0+ is deprecated as of SPDX 2.0-rc2, replaced by GPL-3.0-or-later

LGPL-2.1 also is now LGPL-2.1-only


adriaandegroot:
This is good work. Could you, from here on out on this PR, use the standard Calamares commit message format *[modulename] description*? There's no tooling dependent on it, but it helps distinguish the CMakeLists.txt in various modules from each other. In particular the last two changes to ImageRegistry files would be labeled *[libcalamaresui] Update SPDX identifiers*.

I'm also fine with bulk commits updating identifiers in a single module -- those two ImageRegistry commits could be combined into one. Doing libcalamares can be 1, or a handful, of commits, rather than 115 individual ones.

Speaking of libcalamares: that **doesn't** have SPDX markers in it at all, even though it contains installed headers. It probably should. I remember way back talking with Max Mehl about the SPDX bits but after a couple of rounds of REUSE compliance, didn't feel like sitting down to do all the drudge-work. Hence the very, *very* partial coverage of SPDX in the Calamares codebase. So if you feel like chasing the rest of the files, that would be appreciated.

Cal8031:
@adriaandegroot Yep cool, I'll use that standard. I've opened a new branch Cal8031/calamares:libcalamares to do this. Is this the standard that should be used at the start of all files in libcalamares?

```
# === This file is part of Calamares - <https://github.com/calamares> ===
#
#   Copyright 2014-2015, Teo Mrnjavac <teo@kde.org>
#   Copyright 2016, Kevin Kofler <kevin.kofler@chello.at>
#   Copyright 2020, Adriaan de Groot <groot@kde.org>
#
#   Calamares is free software: you can redistribute it and/or modify
#   it under the terms of the GNU General Public License as published by
#   the Free Software Foundation, either version 3 of the License, or
#   (at your option) any later version.
#
#   Calamares is distributed in the hope that it will be useful,
#   but WITHOUT ANY WARRANTY; without even the implied warranty of
#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
#   GNU General Public License for more details.
#
#   You should have received a copy of the GNU General Public License
#   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
#
#   SPDX-License-Identifier: GPL-3.0-or-later
#   License-Filename: LICENSE
#
###
```

adriaandegroot:
That's a shell / CMake-style header, so I wouldn't put it in a .cpp or .h file :)

The SPDX part looks right, though. Everything in there is GPL-3.0-or-later as stated. REUSE suggests also putting in SPDX-FileCopyrightText lines, which can duplicate the copyright lines in the files (three, in the example header you have here).

Cal8031:
Accidental commit.

Cal8031:
@adriaandegroot Gonna close will open new when I am done.

PR #1422, opened 2020-05-30
remove unnecessary period mark since it's a title

andy1247008998:


adriaandegroot:
This will land after the next release so as not to disturb too many translators.

PR #1423, opened 2020-05-31
Update and add SPDX identifiers

Cal8031:
Updated existing SPDX license idents. And, added SPDX FileCopyright lines and SPDX License identifiers to libcalamares.

Added BSD2 license (due to src/libcalamares/locale/cldr-extractor.py and src/libcalamares/locale/zone-extractor.py) for SPDX

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 #   You should have received a copy of the GNU General Public License
 #   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
 #
-#   SPDX-License-Identifier: GPL-3.0+
+#   SPDX-License-Identifier: GPL-3.0-or-later
 #   License-Filename: LICENSE
 #
 ###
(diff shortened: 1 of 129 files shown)


adriaandegroot:
There's sporadic issues here.

 - REUSE, or [Matija](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change) isn't a fan of bumping Copyright statements by year. Calamares has a bunch of ranged statements, so we shouldn't change that, but:
 - I don't mind changing 2017,2020 to 2017-2020 because insofar as there's going to be a copyright issue, likely the most interesting bit is the first year, the last yeast is the latest possible contribution.
 - Changes from 2017 to 2017- or 2020 to 2020- (that seems to happen with my Copyright statements) I don't like at all. They leave an open-ended range so they **increase** ambiguity. I'd need to go through and "close" the range if I stop contributing. With Matija's sentiments (above) I'd just drop the - and make it a single year, no range. Or for cases where it was 2017-2019, leave it at that.
 - The changes where I misspelled my own name and you fixed it are good :+1:
 - I'm not convinced of the change to `CountryData_p.cpp`. That's a generated file, and it's entirely derived from something else. The tool used to generate it has its own copyright, as does the data it's generated from.

So, good, but there's some fiddly details (that's always the case, eh, with licensing)

Cal8031:
@adriaandegroot Should be better.

PR #1436, opened 2020-06-16
[partition] Strip extra file after the at sign

gportay:
- os-proper may return an extra file after the device:
  /dev/sda1:Ubuntu 19.10 (19.10):Ubuntu:linux
  /dev/sdb1@/EFI/Microsoft/Boot/bootmgfw.efi:Windows Boot Manager:Windows:efi

### DIFF
--- a/src/modules/partition/core/OsproberEntry.h
+++ b/src/modules/partition/core/OsproberEntry.h
@@ -41,6 +41,7 @@ struct OsproberEntry
 {
     QString prettyName;
     QString path;
+    QString file;
     QString uuid;
     bool canBeResized;
     QStringList line;
--- a/src/modules/partition/core/PartUtils.cpp
+++ b/src/modules/partition/core/PartUtils.cpp
@@ -393,17 +393,25 @@ runOsprober( DeviceModel* dm )
                 prettyName = lineColumns.value( 2 ).simplified();
             }
 
-            QString path = lineColumns.value( 0 ).simplified();
+            QString file, path = lineColumns.value( 0 ).simplified();
             if ( !path.startsWith( "/dev/" ) )  //basic sanity check
             {
                 continue;
+            }
+
+            // strip extra file after device: /dev/name@/path/to/file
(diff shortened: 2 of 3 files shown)


gportay:
I am not sure about the second part of the patch because we lose some information.

``` diff
             osproberEntries.append(
                 { prettyName, path, QString(), canBeResized( core, path ), lineColumns, fstabEntries, homePath } );
-            osproberCleanLines.append( line );
+            osproberCleanLines.append( lineColumns.join( ':' ) );
         }
     }
```
We lost the part `/EFI/Microsoft/Boot/bootmgfw.efi:Windows Boot Manager:Windows:efi` in the example below:

```
/dev/sda1:Ubuntu 19.10 (19.10):Ubuntu:linux
/dev/sdb1@/EFI/Microsoft/Boot/bootmgfw.efi:Windows Boot Manager:Windows:efi
```
Maybe, the users of the global storage `osproberLines` should strip it themself before using it.

@adriaandegroot, any though?

adriaandegroot:
I wonder why the replace widget is using the (unstructured) os-prober lines, rather than the structured data that has been collected from them? Possibly we should put the @-part into the OspropberEntries -- it **looks** like they're related to the EFI-boot-executable that is used for them, so worth saving.

I'd prefer something to get rid of the big stringlist, overall. Are there any other users of that stringlist than replacewidget?

adriaandegroot:
@gportay is this still in the RFC stage? Functionally it seems sound to me (you have systems that return this kind of data, so you can tell me if consumers of the os-prober data are affected better than I can).

gportay:
I would say it is not an RFC anymore.

Do you know an existing use case? I made the change to suppress a warning iirc. If not, I can test it test and tells the changes.

I do not have the time to test it right now as I am on holiday. Let me have a look at it on September.

gportay:
Rebased on `calamares/calamares`.

gportay:
Fixed indentation violations.

PR #1435, opened 2020-06-16
Partition Attributes

gportay:
Hello,

This PR adds the attribute `attributes` that sets the GPT attributes to the GPT partition.

This PR requires the [MR](https://invent.kde.org/system/kpmcore/-/merge_requests/9) to be merged in KPMCore. Merged.

This PR is similar to #1409 for GPT partition UUID (`PARTUUID`).

Regards,
Gaël

PR #1437, opened 2020-06-16
[libcalamares] Handle integers prefixed with 0 or 0x

gportay:
- QString to-integer members detect if an integer string begins with
  "0x" (base 16) or "0", base 8; but QVariant members do not.
- QString: the C language convention is used is base is set to 0.
- Convert to QString and use its member toLongLong() and set base to 0
  to detect integer strings begin with a prefix.

### DIFF
--- a/src/libcalamares/utils/Variant.cpp
+++ b/src/libcalamares/utils/Variant.cpp
@@ -72,14 +72,7 @@ getInteger( const QVariantMap& map, const QString& key, qint64 d )
     if ( map.contains( key ) )
     {
         auto v = map.value( key );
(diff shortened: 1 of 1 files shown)


gportay:
I have introduced the function `getUnsignedInteger()` with #1435.

This function was initially based on the function `getInteger()`. Because I wanted support for hexadecimal values (for readability in the configuration file, i.e. prefixed with `0x`), I had to convert the `QVariant` to `QString` to use its `QString::toULongLong(ok, base)` that detects the base prefix if `base` is set to 0.

This commit applies the same trick to `getInteger()`.

PR #1441, opened 2020-06-21
[packages] add pamac support

codesardine:


adriaandegroot:
The `check_db_lock()` function is a bit weird:
 - The `os.exists()` check checks for the file's existence relative to the **host** system, and then runs `rm` in the **target* system. That's not going to help if the root filesystems don't overlap.
 - Why not just `rm -f` the lock file (in the target system)? Then you don't get a failure if the file doesn't exist.

adriaandegroot:
I should add: looks good overall, seems like a sensible addition. You should definitely address the issues above,  and it would be nice if you also:

- added pamac to the list of supported package managers in the example configuration file
- added pamac to the list of supported package managers in the schema file (you'll need to rebase for that, since the schema is being updated to actually work)

codesardine:
I have updated let me know if I miss something.

adriaandegroot:
Merged, thanks! This will be in 3.2.27 which is scheduled for this week.

PR #1444, opened 2020-06-22
[partition] Add the GPT type and attributes to global storage

gportay:
Hello,

This PR adds the GPT partition type `parttype` and `partattrs` to the global storage (as for `partlabel` and `partuuid` since commit 3d2b9053b08b41ee9dab70c61a9542324442b35a).

I have no opinion on with format to use to serialize the attributes to the global storage as is a 64bit integer, and should be expressed in hexadecimal as it is flags. String is fine? Any idea?

Regards,
Gaël


adriaandegroot:
Merging, though it might need some KPMCore API checks.

PR #1464, opened 2020-07-24
Revert Manual Partition instructions

apt-ghetto:
With PR calamares/calamares#1357 the label of the "Manual partitioning" option
was changed, which introduced several downsides:
  * The label is shown for UEFI and for BIOS installations.
  * The mountpoint of the ESP is and should be distro specific.
  * The label always mentioned GPT, which is irrelevant.
  * The label should explain, what the option does, and not, what
    problems can occur under certain circumstances.

### DIFF
--- a/src/modules/partition/gui/ChoicePage.cpp
+++ b/src/modules/partition/gui/ChoicePage.cpp
@@ -332,9 +332,7 @@ ChoicePage::setupChoices()
 
     CALAMARES_RETRANSLATE(
         m_somethingElseButton->setText( tr( "<strong>Manual partitioning</strong><br/>"
-                                            "You can create or resize partitions yourself."
-                                            " Having a GPT partition table and <strong>fat32 512Mb /boot partition "
-                                            "is a must for UEFI installs</strong>, either use an existing without formatting or create one." ) );
+                                            "You can create or resize partitions yourself." ) );
         updateSwapChoicesTr( m_eraseSwapChoiceComboBox );
     )
 }


apt-ghetto:
@demmm FYI

demmm:
Regardless of reverting (irrelevant for local use, since this has been in place since 2014), the label clearly mentions GPT is a must for UEFI, certainly relevant for UEFI systems.
But /boot is not right for many cases, so probably best to revert, it will cause more questions then answers now for systems not using /boot for ESP.

apt-ghetto:
> the label clearly mentions GPT is a must for UEFI, certainly relevant for UEFI systems.

Why is GPT a must for UEFI?

adriaandegroot:
I'd much rather fix-it-right than just revert.

We know what it **should** say:
- on a BIOS system, no additional message
- on an EFI system, something about a suitable EFI partition, mounted somewhere (that is distro specific)

Whether the message needs to be right there, or somewhere else (e.g. on the partitioning page, or when leaving the module, or ..) is still up for debate. Probably should go on the manual partitioning page.

I did get a bug report elsewhere about exactly this message, though -- it is causing confusion in some places where the "for EFI" part isn't noticed when the user has a BIOS system.

kkofler:
IMHO, this revert makes complete sense. What is really needed is actual error-checking, not a message at the beginning stating "you need to do this and that or the installation will be silently screwed up". The message is just a workaround for missing error checking, and we have been sitting on that workaround for way too long.

apt-ghetto:
I am not against showing a warning or error. But in my opinion, the right place is in the manual partitioning screen and is only shown, if needed.

It makes sense to do error checking on the manual partitioning screen. Checking, that a ESP is specified in EFI mode makes absolutely sense and is something I wish to have. The question is more, how to handle a missing ESP?

- Should it disable the "Next" button and prevent to continue the installation?

- Or should it show only a warning?

- Or show a warning and when the user clicks on the "Next" button, a confirmation dialog appears, where the user has to confirm and after that, the installation continues?

adriaandegroot:
There's a precedent of leaving-the-partitioning-page showing a pop-up warning, but I don't have an example at hand just now.

demmm:
>There's a precedent of leaving-the-partitioning-page showing a pop-up warning, but I don't have an example at hand just now.

https://github.com/calamares/calamares/blob/calamares/src/modules/partition/gui/PartitionViewStep.cpp#L473

PR #1487, opened 2020-08-11
[mkinitfs] New module

PureTryOut:
This module allows the generation of the initramfs in Alpine Linux based
systems (excluding postmarketOS). Very bare bones, but then again it
doesn't need much. It uses the Alpine Linux tool "mkinitfs" to do the
job.

### DIFF
--- a/src/modules/mkinitfs/main.py
+++ b/src/modules/mkinitfs/main.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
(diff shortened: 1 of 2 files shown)


adriaandegroot:
Since this is just running a single command in the target system, why not use *shellprocess*? You don't get nice naming and translation then, though .. I suppose that's reason enough.

PureTryOut:
Hmm, yeah I guess that could've worked. I just checked how the other systems generated their initramfs and it seems they did it the same, so I basically copied it. `dracut` for example does the same thing.

PR #1491, opened 2020-08-14
retrieving string list fix

asif-mahmud:
previous implementation of type checking variable `v` was never true, as a string list in the map is actually of type `QVariantList`. its better to use the framework provided method `canConvert` to check if this can yield a string list and act on it. this may also fix the issue with `users` module's group bug. users module uses `getStringList` method to retrieve the config group list, which never returns the actual group list found in the config because of this bug.

### DIFF
--- a/src/libcalamares/utils/Variant.cpp
+++ b/src/libcalamares/utils/Variant.cpp
@@ -67,7 +67,7 @@ getStringList( const QVariantMap& map, const QString& key, const QStringList& d 
     if ( map.contains( key ) )
     {
         auto v = map.value( key );
-        if ( v.type() == QVariant::StringList )
+        if ( v.canConvert( QMetaType::QStringList ) )
         {
             return v.toStringList();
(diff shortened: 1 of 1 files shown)


gportay:
It sounds good but I guess this change can be generalized to all get methods, right?

asif-mahmud:
I think so. `QVariant` has many getters and runtime type checking methods.

adriaandegroot:
Thanks! I added a test before merging this, to demonstrate that there was a problem, and that the PR fixes it. And then fixed the test, since there's now an additional case where a single-string can be converted into a stringlist.

adriaandegroot:
.. and added some more tests to the users module, showing that there was a problem there **too** and that it's fixed. Also, I should have read this PR before releasing 3.2.28.3 :(

asif-mahmud:
i found another issue in the network api. will create a pull request for that later today.

PR #1492, opened 2020-08-17
Fix #1471 on more CPU's

### LINKED ISSUE #1471: hostinfo test fails on non-Intel CPU's
**Describe the bug**
When running the hostinfo tests, specifically `HostInfoTests::testHostOS()`, it fails on non-Intel CPU's due to expecting `hostCPU()` returning "Intel". Obviously not everyone has Intel CPU's so it'll fail in those cases, including on my AMD machine.

**To Reproduce**
1. Run the tests (`ctest`) on a non-Intel machine
2. Observe hostinfo failing

**Expected behavior**
The test to succeed no matter what CPU is used

PureTryOut:
https://github.com/calamares/calamares/commit/5066624a8ebaac6c3ab3dc2cfb111e3c7e3cf40d fixed the `testHostOS` test on AMD and arm machines. However, at least the Alpine Linux aarch64 CI builder returns "CPU implementer: 0x50" rather than 0x41 like some other arm CPU's seem to do. Let's make sure it passes there as well, and it could possibly fix it for more distributions building it for arm.

(I should probably add some comment for which kind of device reports 0x50, but I don't know a proper message to put there)

### DIFF
--- a/src/modules/hostinfo/HostInfoJob.cpp
+++ b/src/modules/hostinfo/HostInfoJob.cpp
@@ -117,7 +117,7 @@ static QString
 hostCPUmatchARM( const QString& s )
 {
     // Both Rock64 and Raspberry pi mention 0x41
-    if ( s.contains( ": 0x41" ) )
+    if ( s.contains( ": 0x41" ) || s.contains( ": 0x50" ) )
     {
         return QStringLiteral( "ARM" );
     }


kkofler:
It should actually be enough to just check that the "CPU implementer" field exists to begin with. This is an ARM-specific field, so its mere existence should prove that this is an ARM CPU. And it can probably take several different values, because there are several companies producing ARM CPUs.

PureTryOut:
Not needed since https://github.com/calamares/calamares/commit/5d016baee92aa0c358c58655f4723fb86bb71672

PR #1494, opened 2020-08-21
Network manager internet checking fix

asif-mahmud:
Qt's `QNetworkAccessManager::networkAccessible` is obsolete from version 5.15. I have checked with latest qt 5.15.0. the problem is this method now returns `true` even if the network/internet is not accessible. so using that to determine internet availability fails with newer versions of qt and probably will be removed from later versions of qt. it is still safe to use the `synchronousPing` method and i think it will be safe to use it for foreseeable future.

### DIFF
--- a/src/libcalamares/network/Manager.cpp
+++ b/src/libcalamares/network/Manager.cpp
@@ -144,7 +144,7 @@ Manager::Manager()
 {
 }
 
-Manager::~Manager() {}
+Manager::~Manager() { }
 
 Manager&
 Manager::instance()
@@ -164,7 +164,7 @@ Manager::checkHasInternet()
 {
     d->m_hasInternet = synchronousPing( d->m_hasInternetUrl );
     emit hasInternetChanged( d->m_hasInternet );
-    return hasInternet;
+    return d->m_hasInternet;
 }
 
 void
--- a/src/libcalamares/network/Manager.cpp
+++ b/src/libcalamares/network/Manager.cpp
@@ -163,7 +163,7 @@ bool
 Manager::checkHasInternet()
 {
     d->m_hasInternet = synchronousPing( d->m_hasInternetUrl );
-    emit hasInternetChanged( hasInternet );
+    emit hasInternetChanged( d->m_hasInternet );
     return hasInternet;
 }
 
--- a/src/libcalamares/network/Manager.cpp
+++ b/src/libcalamares/network/Manager.cpp
@@ -162,17 +162,8 @@ Manager::hasInternet()
 bool
 Manager::checkHasInternet()
 {
-    bool hasInternet = d->nam()->networkAccessible() == QNetworkAccessManager::Accessible;
-
-    if ( !hasInternet && ( d->nam()->networkAccessible() == QNetworkAccessManager::UnknownAccessibility ) )
-    {
-        hasInternet = synchronousPing( d->m_hasInternetUrl );
-    }
-    if ( hasInternet != d->m_hasInternet )
-    {
-        d->m_hasInternet = hasInternet;
-        emit hasInternetChanged( hasInternet );
-    }
+    d->m_hasInternet = synchronousPing( d->m_hasInternetUrl );
+    emit hasInternetChanged( hasInternet );
     return hasInternet;
 }
 


kkofler:
Looking at:

https://code.qt.io/cgit/qt/qtbase.git/tree/src/network/access/qnetworkaccessmanager.cpp?h=5.15#n1113

it does not look like this method always returns `true`. What is true is:

* The method works only as well as the underlying NetworkManager connectivity check does. NetworkManager can be set up to do something very similar to `synchronousPing` with a distro-specified website (see, e.g.: https://src.fedoraproject.org/rpms/NetworkManager/blob/master/f/20-connectivity-fedora.conf ). When you don't have such a configuration from your distribution, then NetworkManager can only check that a network connection was established, not that it is actually connected to the Internet. (In particular, you may have to log in to a captive portal to get Internet access.)
* The method is indeed gone in Qt 6, because Qt itself no longer uses it, see https://code.qt.io/cgit/qt/qtbase.git/commit/src/network/access/qnetworkaccessmanager.cpp?h=5.15&id=a96fc6e65cba14d80665ef5f8932f004f7f4a1e3 and https://codereview.qt-project.org/c/qt/qtbase/+/292813 . The replacement will probably be to use https://api.kde.org/frameworks/networkmanager-qt/html/index.html (see `checkConnectivity` and `connectivity`) or the NetworkManager C API directly. This is quite unfortunate, Qt removing APIs useful to the user just because they are no longer used internally.

The problem with just using `synchronousPing` is that we no longer make use of the information that NetworkManager has. If NetworkManager **knows** that we are not connected, it is quite pointless to attempt a ping.

asif-mahmud:
the reason behind saying it always returns true is this, i ran it in 2 virtual machines one with internet access other without. in both cases the following line -
```
bool hasInternet = d->nam()->networkAccessible() == QNetworkAccessManager::Accessible;
```
sets `hasInternet` to `true` because `d->nam()->networkAccessible()` returned `QNetworkAccessManager::Accessible`. qt version 5.15.0. i checked the links which you have provided. it is really surprising that they are removing such functionality. i tried to find an alternative, but no luck. will update if i have any better idea. but for now, `synchronousPing` seemed working, and it was only called once when setting the globalStorage value `hasInternet`. other than the fact the manager wont have the availability of qt api to acknowledge or verify network accessibility, for a module developer i think the globalstorage or the manager's api is just fine. but your point still stays valid.

kkofler:
How is the virtual machine "without network access" set up?

Virtual machines are pretty special: they tend to always have Ethernet emulation configured by default, and that emulation is set up to make the VM think it is connected to something. But if the host does not have Internet connectivity, then of course the guest will be able to connect at most to the host. Still, NetworkManager will think that it is connected, unless you have the connectivity check within NetworkManager configured. (It is off by default in upstream NetworkManager because it means "phoning home" to a distro-provided website, and each distro sets up a different website for it, if it wants this feature to begin with.)

The `synchronousPing` check will of course notice that the website you configure for it is not actually reachable. But if we trust the NetworkManager (or QNetworkAccessManager) "accessible" status, we are not running that check.

asif-mahmud:
In virtualbox i can remove any network adapter. Same can be done with virt manager.

adriaandegroot:
nit-picky bits: @asif-mahmud Can you squash this and force-push so the "fix" and similar commits are gone? In general we try to commit with *[module]* in the commit message (here, module is *libcalamares* I think) as a commit-message-styling thing.

I have not looked into the Qt sources  -- I'll leave that to you to verify -- but I'd like to use the method when / if it's available, and if we can use the talk-to-network-manager code from Qt in Calamares (to avoid doing synchronous ping ourselves) then that'd be fine as well. (But we need to keep in mind that Calamares officially supportsback to Qt 5.9, too)

asif-mahmud:
Sure. I will do that. As for support from Qt 5.9, i think its fine cause the whole purpose of this method is check if calamares can access internet. The `synchronousPing` method works for all the versions. I will clean up the commits. May be a new PR with the changes only.

asif-mahmud:
Closing this PR. A merged/changes only PR will be created.

PR #1501, opened 2020-08-26
[localeq] rewrite Offline.qml

demmm:
once completed, this can be a fully functional (offline) locale selection option
worldmap.png no longer needed/in use
working is the stackview of the region & zones models
Timezone text bar shows correct timezone
currentIndex see comments on lines 65 & 139, not working
update of timezone text bar can't be tested if working as long no index is connected (see lines 93 & 168)
Still, already committing, since it does more then old Offline.qml, which had no function for timezone


demmm:
Qmlscene view of Offline.qml with this commit:
Screenshot_20200826_182216

PR #1502, opened 2020-08-27
Internet accessibility checking done by synchronous ping method

asif-mahmud:
Checking whether or not calamares can access the internet is done by only the `synchronousPing` method. Additionally the `QNetworkAccessManager::networkAccessible` flag is set to reflect the finding.

### DIFF
--- a/src/libcalamares/network/Manager.cpp
+++ b/src/libcalamares/network/Manager.cpp
@@ -149,18 +149,23 @@ Manager::hasInternet()
 bool
 Manager::checkHasInternet()
 {
-    bool hasInternet = d->nam()->networkAccessible() == QNetworkAccessManager::Accessible;
 
-    if ( !hasInternet && ( d->nam()->networkAccessible() == QNetworkAccessManager::UnknownAccessibility ) )
+    d->m_hasInternet = synchronousPing( d->m_hasInternetUrl );
+
+// For earlier Qt versions (< 5.15.0), set the accessibility flag to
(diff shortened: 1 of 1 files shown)


kkofler:
I don't think it is necessary to call `setNetworkAccessible`.

adriaandegroot:
> I don't think it is necessary to call `setNetworkAccessible`.

Calamares code **should** be using the Calamares Network service (e.g. this code) to check for network accessibility ; unless QNAM is doing some sharing under the hood, the  QNAM that the Calamares Network service has is not accessible from elsewhere, and so nothing will be asking that QNAM directly for status. Anything using its own QNAM will have its own problems :shrug: (I just checked, different QNAM objects maintain their own *networkAccessible* value).

That said, I don't think the call is worth worrying about (if anything, it'd make sense to gripe about emitting the changed signal even if the accessibility **hasn't** changed, but that minor problem exists in the original code as well).

adriaandegroot:
Well, leaving out the call if `setNetworkAccessible` would get rid of the `#if` and the comment and turn the whole thing into "just synchronous ping" which I think is what Asif was aiming for from the start

asif-mahmud:
Actually i thought it would make sure any other method using QNAM instance of Network directly wont get incorrect or unexpected value. But if you think its unnecessary then i can of course remove that part. My intension was to make sure every part of the code not just the modules to know whether internet is accessible or not in a reliable way.

PR #1513, opened 2020-09-16
[partition] Correctly handle percentage-define partitions

tintou:
* Use the minSize when the target storage is smaller than the sum of sizes
 * Percentage-defined partitions should be computed after setting hard-defined ones

This fixes issues when 0 byte partitions were created when the disk is too small.
Also fixes an issue with percent-defined partitions being forced to be defined at the end of the disk.

### DIFF
--- a/src/modules/partition/core/PartitionLayout.cpp
+++ b/src/modules/partition/core/PartitionLayout.cpp
@@ -164,22 +164,36 @@ PartitionLayout::execute( Device* dev,
                           const PartitionRole& role )
 {
     QList< Partition* > partList;
+    // Map each partition entry to its requested size (0 when calculated later)
+    QMap< const PartitionLayout::PartitionEntry *, qint64 > partSizeMap;
     qint64 minSize, maxSize, end;
     qint64 totalSize = lastSector - firstSector + 1;
     qint64 availableSize = totalSize;
 
-    // TODO: Refine partition sizes to make sure there is room for every partition
-    // Use a default (200-500M ?) minimum size for partition without minSize
-
-    foreach ( const PartitionLayout::PartitionEntry& part, m_partLayout )
+    // Let's check if we have enough space for each partSize
+    for( const PartitionLayout::PartitionEntry& part : m_partLayout )
     {
-        Partition* currentPartition = nullptr;
-
         qint64 size = -1;
         // Calculate partition size
+
         if ( part.partSize.isValid() )
         {
-            size = part.partSize.toSectors( totalSize, dev->logicalSize() );
+            // We need to ignore the percent-defined
+            if ( part.partSize.unit() != CalamaresUtils::Partition::SizeUnit::Percent)
+            {
+                size = part.partSize.toSectors( totalSize, dev->logicalSize() );
+            }
+            else
+            {
+                if ( part.partMinSize.isValid() )
+                {
+                    size = part.partMinSize.toSectors( totalSize, dev->logicalSize() );
+                }
+                else
+                {
+                    size = 0;
+                }
+            }
         }
         else
         {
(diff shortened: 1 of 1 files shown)


src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -164,60 +164,118 @@ PartitionLayout::execute( Device* dev,
                           const PartitionRole& role )
 {
     QList< Partition* > partList;
+    QMap< const PartitionLayout::PartitionEntry *, qint64 > partSizeMap;
```
adriaandegroot, line 168: This could use a comment explaining that it is the .. um .. calculated minimum size for each partition, or 0 if the partition is defined only as percentages?

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -164,60 +164,118 @@ PartitionLayout::execute( Device* dev,
...
     qint64 minSize, maxSize, end;
     qint64 totalSize = lastSector - firstSector + 1;
     qint64 availableSize = totalSize;
+    qint64 minPartitionSize = 0;
```
adriaandegroot: This seems unused?

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -164,60 +164,118 @@ PartitionLayout::execute( Device* dev,
...
 
-    // TODO: Refine partition sizes to make sure there is room for every partition
-    // Use a default (200-500M ?) minimum size for partition without minSize
-
+    // Let's check if we have enough space for each partSize
     foreach ( const PartitionLayout::PartitionEntry& part, m_partLayout )
```
adriaandegroot: Though this isn't your code, Calamares now uses range-style for, so `for( const PartitionLayout::PartitionsEntry& part : m_partLayout )`. That convention has changed since this code was first written, and now I see it's not updated consistently everywhere. The same applies to your other foreach()es, could you re-do them in range-style?

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -164,60 +164,118 @@ PartitionLayout::execute( Device* dev,
...
-        {
-            minSize = part.partMinSize.toSectors( totalSize, dev->logicalSize() );
-        }
-        else
-        {
-            minSize = 0;
-        }
+        partSizeMap.insert (&part, size);
+        availableSize -= size;
+    }
```
adriaandegroot, line 206: This entire for-loop feels like it could be factored out into a static inline free function, but that's probably more code-churn than you want to get into.

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -164,60 +164,118 @@ PartitionLayout::execute( Device* dev,
...
+            }
+            else
+            {
+                size = 0;
+            }
+
+            partSizeMap.insert (&part, size);
+            availableSize -= size;
         }
+    }
```
adriaandegroot, line 239: So this loop is the same as the previous, only preferring min-size over size (and the previous was preferring size over min-size). That strikes me as a candidate for refactoring with some kind of lambda to get preferred- and secondary-sizes. Again, that might be more code-churn than you want to do.

adriaandegroot (approved):

adriaandegroot:
Hi @tintou, welcome to the club.

Overall this looks good. My review comments above are mostly "this could be even better". If you would be willing to polish it up a little more, that'd be great. If not (or you don't have time right now, or whatever), please say so, and I'll land it as-is.

You will be listed in the contributors-to-this-relese (whenever this lands) as "Corentin Noël" in the release notes, unless you want something else.

src/modules/partition/core/PartitionLayout.cpp
tintou: Indeed, added a small comment to explain it a little bit

src/modules/partition/core/PartitionLayout.cpp
tintou: Yeah let's have incremental steps :slightly_smiling_face:

src/modules/partition/core/PartitionLayout.cpp
tintou: changed

tintou:
> Hi @tintou, welcome to the club.
> 
> Overall this looks good. My review comments above are mostly "this could be even better". If you would be willing to polish it up a little more, that'd be great. If not (or you don't have time right now, or whatever), please say so, and I'll land it as-is.
> 
> You will be listed in the contributors-to-this-relese (whenever this lands) as "Corentin Noël" in the release notes, unless you want something else.

Thanks!
I've pushed the changes that were "straightforward" but don't really want to go into a deep-refactoring of that part yet :slightly_smiling_face: 

You can put "Corentin Noël", that works for me

adriaandegroot:
I can see by the travis notifications that the answer was "yes", too :)

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -164,60 +164,118 @@ PartitionLayout::execute( Device* dev,
...
     qint64 minSize, maxSize, end;
     qint64 totalSize = lastSector - firstSector + 1;
     qint64 availableSize = totalSize;
 
-    // TODO: Refine partition sizes to make sure there is room for every partition
-    // Use a default (200-500M ?) minimum size for partition without minSize
-
-    foreach ( const PartitionLayout::PartitionEntry& part, m_partLayout )
+    // Let's check if we have enough space for each partSize
+    for( const PartitionLayout::PartitionEntry& part : m_partLayout )
```
kkofler, line 174: Shouldn't this say `qAsConst( m_partLayout )`? (The number one pitfall when using C++11 ranged `for` with Qt containers.)

PR #1518, opened 2020-09-30
Inheritance In CPP

adityasunny1189:
OOPS Concept


kkofler:
I do not think that this is the place to which you wanted to submit this file.

PR #1519, opened 2020-10-01
[partition] Fix regression in partition size assignment

tintou:
In some cases, the partition size was set to zero as the maxSize isn't always defined.


adriaandegroot:
Can we come up with some (unit)test cases so that we can be more confident that the partition size assignment is good?

adriaandegroot:
KDE neon reported a problem where "erase" would try to create a 0B partition and FS; this PR fixed that problem.

PR #1522, opened 2020-10-04
[preservefiles] Fix global storage JSON dump

### LINKED ISSUE #1521: preservefile job's global config JSON dump succeeds, but results in a job error
**Describe the bug**
The preservefiles module's special *from* key value `config` results in a job error on successful global config dump:

```
Not all of the configured files could be preserved.
```

Also, the description of special *from* key values in sample config file [preservefiles.conf](https://github.com/calamares/calamares/blob/v3.2.30/src/modules/preservefiles/preservefiles.conf#L34) is inaccurate:

```yaml
# Special values for the key *from* are:
# - *log*, for the complete log file (up to the moment the preservefiles
#   module is run),
# - *config*, for the Calamares configuration file
# - *globals*, for a JSON dump of the contents of global storage
```

Value `config` dumps global storage as JSON, while value `globals` is undefined and results in a noop.

**To Reproduce**
1. Use the following in `/etc/calamares/settings.conf`:
```yaml
dont-chroot: true
sequence:
- show:
  - welcome
  - summary
- exec:
  - preservefiles
- show:
  - finished
```
2. Use the following in `/etc/calamares/modules/preservefiles.conf`:
```yaml
files:
  - from: log
    dest: /root/install.log
  - from: config
    dest: /root/install.json
```
3. Click through to install.
4. See that the install fails with an error from the preservefiles job.
5. Run `cat /root/install.json` in a terminal to confirm that the global config dump worked.

**Expected behavior**
The error check is simply inverted -- a successful config dump produces an error, which it should not.

**Additional context**
The error check at [PreserveFiles.cpp#L140](https://github.com/calamares/calamares/blob/v3.2.30/src/modules/preservefiles/PreserveFiles.cpp#L140) needs to be inverted.

kadler15:
Fixes #1521 

Also includes a bit of cleanup to the relevant section of the module config file.

### DIFF
--- a/src/libcalamares/GlobalStorage.h
+++ b/src/libcalamares/GlobalStorage.h
@@ -22,7 +22,7 @@ namespace Calamares
 
 /** @brief Storage for data that passes between Calamares modules.
  *
- * The Global Storage is global to the Calamars JobQueue and
+ * The Global Storage is global to the Calamares JobQueue and
  * everything that depends on that: all of its modules use the
  * same instance of the JobQueue, and so of the Global Storage.
  *
--- a/src/modules/preservefiles/PreserveFiles.cpp
+++ b/src/modules/preservefiles/PreserveFiles.cpp
@@ -137,9 +137,9 @@ PreserveFiles::exec()
         }
         if ( it.type == ItemType::Config )
         {
-            if ( Calamares::JobQueue::instance()->globalStorage()->saveJson( dest ) )
+            if ( !Calamares::JobQueue::instance()->globalStorage()->saveJson( dest ) )
             {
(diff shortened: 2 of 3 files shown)


adriaandegroot:
That .. wow, the disconnect between what I wrote in the documentation as to what I *intended* and what was actually implemented is pretty big. Thank you for tracking that down. Contributors are listed in the CHANGES file; since I intend to put out a release **tonight** I'll list you as *kadler15* because I don't have anything else to go on :) If you want to be identified in some other way in the credits / CHANGES, just let me know (mail, comment in this PR, whatever).

PR #1524, opened 2020-10-06
[partition] Message user if no option available

gportay:
It happens that no option is available after a disk is selected.

This appears during development while calamares crashes and does not unmount the partitions.

This commit reuses the message label that is hidden since 48d0de2.


adriaandegroot:
Good find! Since these are new strings, I'm going to merge right after releasing Calamares 3.2.31 (which is tonight), and then apply code-style tools to it and then figure out how to review-bot all the things.

PR #1539, opened 2020-10-15
Add missing space

yurchor:
Many thanks in advance for reviewing.


adriaandegroot:
Hi Yuri,

Good catch.  Reading the source message tells me that the message itself needs improving , so I'll fix that after this merge. I would never have spotted that without your PR.

yurchor:
Thanks for your work. :)

PR #1544, opened 2020-10-25
improved btrfs support

ghost:
although it is hardcoded but it is better than current implementation in calamares.

### DIFF
--- a/src/libcalamares/PythonJob.cpp
+++ b/src/libcalamares/PythonJob.cpp
@@ -284,7 +284,7 @@ PythonJob::exec()
             return JobResult::error( message, description );
         }
     }
-    catch ( bp::error_already_set )
+    catch ( bp::error_already_set& )
     {
         QString msg;
         if ( PyErr_Occurred() )
--- a/src/libcalamares/utils/BoostPython.h
+++ b/src/libcalamares/utils/BoostPython.h
@@ -52,6 +52,7 @@
 #endif
 
 #undef slots
+#include <boost/iterator.hpp>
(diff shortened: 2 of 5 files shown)


adriaandegroot:
Yeah, all those hardcoded blocks and code repetition show that this needs a proper configuration and code, but it's good to have a collection of "this is what we would want it to look like" since that can steer the design of that configuration.

I'll leave the PR open and close it once something similar has landed.

adriaandegroot:
This bears directly also on #1375 (and has the same kind of issues -- expanding hardcoded lists, rather than introducing configuration and a bit of software architecture)

Chrysostomus:
Tried using this on an installation but, either I misconfigured it or there is something wrong with the patch. It only configured the root subvolume and nothing else. Testing further...

ghost:
https://gitlab.com/garuda-linux/tools/iso-profiles/-/tree/master/shared/garuda/live-overlay/usr/lib/calamares/modules



Here all the modified modules that i use in garuda linux

adriaandegroot:
@librewish thanks for the PR. As you can see I've merged @Chrysostomus 's version because that is configurable from the config file(s) rather than being hard-coded. I hope that's an improvement that is compatible with your work at Garuda Linux -- could you try to follow the config for the mount- and fstab modules as it is now, to replace your hard-coded setup?

If so, then you can simplify to a config-file rather than patches, and we're all better (and if not, file issues and Matti can fix them).

adriaandegroot:
Closing as "implemented in other way"

PR #1553, opened 2020-10-27
Initial additional layout support for non-ASCII layouts 

LordTermor:


adriaandegroot:
Merged, thank you! I have a *fix-keyboardq* branch that uses all the new Config bits -- it is actually more about fixing the *keyboard* module, and I can tell that the right files are written with improved layouts already -- which will land shortly and bring in your work for the next release.

Looking at the code afterwards for style and improvements, I was wondering about this one:
```
 static AdditionalLayoutInfo getAdditionalLayoutInfo( const QString& layout );
```

That looks like it could be a constructor of *AdditionalLayoutInfo* (although then it would also need a default constructor, I think) and then the documentation for what that represents can move to the *AdditionalLayoutInfo* header.

Issue #1517, opened 2020-09-29
Pick timezone from /etc/timezone


jo-so:
The value of */etc/timezone* might have been set by the live environment. So, it would be useful to pick the initial setting from */etc/timezone*, if it's set.

adriaandegroot:
- What OS , what distro?
- What kind of value is set there?

Making me guess these things makes it much harder to do anything about an issue.

adriaandegroot:
.. since I have a Manjaro VM open just right now, for the record:

- A regular recent Manjaro install has an `/etc/timezone` file
- The file is a regular file
- The content of the file is the name of the selected timezone, e.g. `Europe/Amsterdam` followed by a newline
- In a Manjaro KDE Plasma installation, neither the *Time & Date* nor the *Date and Time* configuration modules change the content of `/etc/timezone`, while they **do** change the standard `/etc/localtime` link

I'm going to mark this issue up as a hacktoberfest one, because it's straightforward to implement (even if I doubt that it is useful; `/etc/timezone` is sufficiently non-standard that `QTimeZone::systemTimeZoneId()` is not going to consult it either)

kkofler:
My Fedora installation does not have an `/etc/timezone` file at all.

adriaandegroot:
Hacktoberfest 2020 will start soon, and this is a good kind of issue to do for it.

## Procedural

- [ ] Check that your Linux distro has a `/etc/timezone`. (Fedora and FreeBSD do not; Manjaro does; I have an older Netrunner Debian that does too). If this is interesting for a live CD, check that the live CD carries such a file or creates one.
- [ ] Make sure you can compile Calamares locally (see https://github.com/calamares/calamares/blob/calamares/CONTRIBUTING.md, which will point you onwards to the wiki)
- [ ] Make sure the locale module is enabled in your Calamares build (by default it is compiled, and in settings.conf: check the CMake output to see that it will be compiled, and then double-check that locale is listed in the show section of the sequence in settings.conf)
- [ ] Module sources are in https://github.com/calamares/calamares/tree/calamares/src/modules/locale ; make a branch, add your copyright statement to one of the .cpp files and add your name to top-level CHANGES, then start a PR with WIP in the title

## Functional

- [ ] Calamares picks up the default timezone (to be pre-selected in the *locale* module) just once, in `src/modules/locale/Config.cpp` in the function *getStartingTimezone()*, where the *region* and *zone* and *useSystemTimezone* are read from the configuration.
[...]

PR #1555, opened 2020-10-27
[WIP] Pick timezone from /etc/timezone #1517

tm9k1:


tm9k1:
I'm thinking
```
bool success = false;
if ( hasInternetConnection )
    success = getTzFromGeoIP(); 
if ( !success && hasEtcTimezone ) 
    success =getTzFromFile(); 
if ( !success ) 
    getTzFromFixedVals();
if ( !success ) 
    success = getTzFromQtDefault(); 
```

adriaandegroot:
The plan I had -- which you don't necessarily have to follow! -- is more along the lines of:
- in `locale.conf`, introduce a key *defaultTimezoneSource*, put it near the top of the file for documentation purposes, since it's going to determine which of the following keys are going to be used; this key can have four values:
  - *fixed*, to use the region/zone set explicitly in the file
  - *system*, to use whatever the system's TZ is (from Qt)
  - *etc-timezone*, to use the contents of that file
  - *geoip*, to do GeoIP lookup

The key *usesystemTimezone* should be marked deprecated; if no source is explicitly set, but *usesystemtimezone* is, then behave as if *system* was specified. There's probably more cases to handle for figuring out which source to use. I like your idea of fallbacks: there's no sense in **failing** to set a TZ if a specific source is mentioned, and it doesn't work but there is other deta available.

adriaandegroot:
Unfortunately, it turns out that *useSystemTimezone* already does all the things we would need to support this, on the systems that usefully support the file. So I'm going to close this as "not needed". Sorry about that, @taskmaster9001

tm9k1:
no worries, @adriaandegroot . I'll try to find something else!

PR #1557, opened 2020-10-30
WIP: Create or reuse home partition

gportay:
Hello Folks,

This PR intends to create a specific home partition or to reuse an existing one, either using `os-probe` and `/etc/fstab` (that is already done) or using the GPT type `933AC7E1-2EB4-4F13-B844-0E14E2AEF915` (new).

The first commits are some cleanup and I will propose them in separates PRs.

The very last is the most important and it is still a WIP and works already. It needs to be split into several commits though (#1558, #1559, #1560, #1561 and #1562).

Regards,
Gaël

### DIFF
--- a/src/libcalamares/partition/PartitionQuery.cpp
+++ b/src/libcalamares/partition/PartitionQuery.cpp
@@ -16,6 +16,7 @@
 
 #include <kpmcore/core/device.h>
 #include <kpmcore/core/partition.h>
+#include <kpmcore/core/partitiontable.h>
 
 namespace CalamaresUtils
 {
@@ -25,6 +26,19 @@ namespace Partition
 // Types from KPMCore
 using ::Device;
 using ::Partition;
+
+const PartitionTable*
+getPartitionTable( const Partition* partition )
+{
+    const PartitionNode* root = partition;
+    while ( root && !root->isRoot() )
+    {
+        root = root->parent();
+    }
+
+    return dynamic_cast< const PartitionTable* >( root );
+}
+
 
 bool
 isPartitionFreeSpace( const Partition* partition )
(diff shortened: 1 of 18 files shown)


abucodonosor:
Can we take care of the forced /home creation when BRFS is used too?

gportay:
For now, I can discover the home partitions (osprober and GPT type) and I can create a home partition if no existing home partition is selected or reuse the one from a combox.

> Can we take care of the forced /home creation when BRFS is used too?

Could you describe more precisely the case so I understand correcly the use case. I know nothing about btrfs for now, I saw that we can create a subvolume in the `/` btrfs device to contains the home.

I would apreciate you tell me more :).

gportay:
Then I will apply the changes to swap so that calamares can [reuse](https://github.com/calamares/calamares/blob/db08d2db8ba31f183d5a538fd838e8b318b18842/src/modules/partition/partition.conf#L49) existing swap.

abucodonosor:
If BTRFS is used we force the creation of a @home subvolume on /@ if a /home mount point doesn't exist which is always true for our default right now- Also that conflicts with the partition layout when used.

See as example https://github.com/calamares/calamares/issues/1222

gportay:
Okay, so there is already a problem as reusing `/home` feature (using `/etc/fstab`+`osprober`) exists already and there may be a conflicts with the `mount` module in case of btrfs.

Normally, the property `reuseHome` should exist in the globalstorage, so the `mount` module should check for this before creating a new btrfs subvolume.

Or maybe, the mountpoint `/home` exists and in that case, there is no problem at all.

I have not yet tested this but normally it is what I expect to have in the end.

Also, I saw btrfs improvement in #1544.

abucodonosor:
Ok, thank you :)

dalto8:
Since this hasn't had a commit in well over 2 years I am going to close it.

If there is ever a desire to finish it please feel free to submit a new PR.

PR #1560, opened 2020-10-31
[partition,mount] Make mountPoint and filesystem optionals

gportay:
Hello,

This PR intends to make both `partitionLayout` attributes `mountPoint` and `filesystem` optional.

In the case of creating a read-only system with `dm-verity`, the block device that contains the hash is not a file-system and cannot be mounted though.

This makes the `mountPoint` irrelevant (as well as `filesystem`).

Note: The "filesystem" type `DM_verityhash` is not supported (yet?) by kpmcore.

```
04:52:18 [0]: unknown file system type  "DM_verity_hash"  on  "/dev/sda3"
``` 

Note: In addition to that PR, the module `mount` fails silently if some partitions failed to be mounted. I have added a warning message to log the failure in a first stage, in order to not break the actual behavior. This behaviour was introduced a while ago with commit 9708669ac7be575beb91dfc0619e4595886bd370. We may want to fail instead in the future.

Regards,
Gaël

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -39,6 +39,9 @@ def mount_partition(root_mount_point, partition, partitions):
     # Create mount point with `+` rather than `os.path.join()` because
     # `partition["mountPoint"]` starts with a '/'.
     raw_mount_point = partition["mountPoint"]
+    if not raw_mount_point:
+        return
+
     mount_point = root_mount_point + raw_mount_point
 
     # Ensure that the created directory has the correct SELinux context on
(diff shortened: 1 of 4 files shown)


gportay:
I have tested it with/without a `btrfs` rootfs and with/without a `/home` partition.

I have not tested it with luks. How can I test it?

gportay:
I have rebased the PR on top for branch `calamares` and have added a new commit to ignore the partitions with the filesystem type `unformatted` in the overview windows.

PR #1561, opened 2020-10-31
[partition] Allow a configurable swap Partition name

gportay:
Hello,

This PR allows setting the GPT name for the Swap partition.

Regards,
Gaël


adriaandegroot:
My only hesitation here is that there doesn't seem to be a reason to push the swap name through Global Storage -- except that existing code does so, indiscriminately. If there's a *later* module that needs the swap name, then GS makes sense, but otherwise we should be looking towards *consolidating* the Partitioning configuration to config / core objects in that plugin, rather than spamming GS with them.

adriaandegroot:
(that's more general "code hygiene in the partitioning module" than an issue with your PR, though)

PR #1563, opened 2020-11-02
[users] Full name transliteration support for user creation

LordTermor:
Use the International Components for Unicode if available to automatically convert unicode full name to unix user name.
screenshot

### DIFF
--- a/src/modules/users/Config.cpp
+++ b/src/modules/users/Config.cpp
@@ -307,12 +307,15 @@ transliterate( const QString& input )
         return input;
     }
 
-    auto transliterable = icu::UnicodeString( input.utf16() );
-
+    icu::UnicodeString transliterable( input.utf16() );
     transliterator->transliterate( transliterable );
-
     return QString::fromUtf16( transliterable.getTerminatedBuffer() );
-
+}
+#else
+static QString
+transliterate( const QString& input )
+{
+    return input;
 }
 #endif
 
@@ -374,13 +377,8 @@ Config::setFullName( const QString& name )
 
         // Build login and hostname, if needed
         static QRegExp rx( "[^a-zA-Z0-9 ]", Qt::CaseInsensitive );
-#ifdef HAVE_ICU
-        QString cleanName = transliterate(name);
-        cleanName.replace("'", "");
-#else
-        QString cleanName = name;
-#endif
-        cleanName = CalamaresUtils::removeDiacritics( cleanName ).replace( rx, " " ).toLower().simplified();
+
+        QString cleanName = CalamaresUtils::removeDiacritics( transliterate( name ) ).replace( QRegExp( "[-']" ), "").replace( rx, " " ).toLower().simplified();
 
 
         QStringList cleanParts = cleanName.split( ' ' );
--- a/src/modules/users/Config.cpp
+++ b/src/modules/users/Config.cpp
@@ -23,6 +23,11 @@
 #include <QFile>
 #include <QRegExp>
 
+#ifdef HAVE_ICU
+#include <unicode/translit.h>
+#include <unicode/unistr.h>
+#endif
+
 static const QRegExp USERNAME_RX( "^[a-z_][a-z0-9_-]*[$]?$" );
 static constexpr const int USERNAME_MAX_LENGTH = 31;
 
@@ -279,6 +284,30 @@ guessProductName()
     }
     return dmiProduct;
 }
+#ifdef HAVE_ICU
+static QString
+transliterate( const QString& input )
+{
+    static UErrorCode ue = UErrorCode::U_ZERO_ERROR;
+    static auto transliterator = std::unique_ptr< icu::Transliterator >(
+                icu::Transliterator::createInstance( "Any-Latin; Latin-ASCII", UTRANS_FORWARD, ue )
+                );
+
+    if(ue!=0){
+        cWarning() << "Can't create transliterator";
+
+        //it'll be checked later for non-ASCII characters
+        return input;
+    }
+
+    auto transliterable = icu::UnicodeString( input.toUtf8().data() );
+
+    transliterator->transliterate( transliterable );
+
+    return QString::fromUtf16( transliterable.getTerminatedBuffer() );
+
+}
+#endif
 
 static QString
 makeLoginNameSuggestion( const QStringList& parts )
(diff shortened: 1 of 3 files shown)


src/modules/users/Config.cpp
```diff
@@ -279,6 +284,30 @@ guessProductName()
...
+                );
+
+    if(ue!=0){
+        cWarning() << "Can't create transliterator";
+
+        //it'll be checked later for non-ASCII characters
+        return input;
+    }
+
+    auto transliterable = icu::UnicodeString( input.toUtf8().data() );
```
kkofler: Please construct the `icu::UnicodeString` directly from UTF-16 data rather than roundtripping through UTF-8. Using `input.utf16()` instead of `input.toUtf8().data()` should do the trick.

LordTermor: Thanks

src/modules/users/Config.cpp
```diff
@@ -337,8 +366,16 @@ Config::setFullName( const QString& name )
         emit fullNameChanged( name );
 
         // Build login and hostname, if needed
+        QString cleanName = CalamaresUtils::removeDiacritics( name ).toLower().simplified();
+
+#ifdef HAVE_ICU
+        cleanName = transliterate(cleanName);
+#else
```
kkofler: Shouldn't the `#else` part always be run (i.e., come after `#endif` rather than in an `#else`), even after `transliterate`? Removing illegal characters is always a good idea.

And I would also do the other cleanups (`removeDiacritics`, `toLower`, `simplified`) after the transliteration, not before. (What if the transliterator spits out some diacritics, uppercase letters, or whitespace?)

adriaandegroot:
- Perhaps comment in the `usersq/CMakeLists.txt` that ICU and the define is needed for `users/Config.cpp`, since it otherwise looks like a weirdly-misplaced duplication (this is partly because the q and non-q modules live apart; longer-term the plan is to merge them)
 - Needs one more run with the code-style tool
 - Add a note to top-level CHANGES about what this does, for the release notes

Other than that, looks really good and useful (and short, and relatively simple)

src/modules/users/Config.cpp
adriaandegroot: It definitely should: transliterations of *Ali* will "correctly" be *`li* (er, I can't even type the little curly for a latinate representation of an 'ain) and that will need another round of stripping and post-processing.

LordTermor: Done f20d0a7593344fac924827e10c2ac16a947f9707

src/modules/users/Config.cpp
```diff
@@ -23,6 +23,17 @@
 #include <QFile>
 #include <QRegExp>
 
+#ifdef HAVE_ICU
+#include <unicode/translit.h>
+#include <unicode/unistr.h>
+
+//Did my best choosing compound ID
+static const char* TRANSLITERATOR_ID = "Russian-Latin/BGN;"
```
adriaandegroot: `static const char TRANSLITERATOR_ID[] =` (for literals that's idiomatic rather than having the array implicit and declaring the pointer)

src/modules/users/Config.cpp
```diff
@@ -23,6 +23,17 @@
 #include <QFile>
 #include <QRegExp>
 
+#ifdef HAVE_ICU
+#include <unicode/translit.h>
+#include <unicode/unistr.h>
+
+//Did my best choosing compound ID
```
adriaandegroot: The comment should probably do two things
 - explain briefly what this does (define a sequence of transliterations that are applied)
 - explain that  this is an initial attempt at doing the transliteration

src/modules/users/Config.cpp
```diff
@@ -279,6 +290,30 @@ guessProductName()
...
+                );
+
+    if( ue != UErrorCode::U_ZERO_ERROR ){
+        cWarning() << "Can't create transliterator";
+
+        //it'll be checked later for non-ASCII characters
+        return input;
+    }
+
+    auto transliterable = icu::UnicodeString( input.utf16() );
```
adriaandegroot: This (and subsequent) is getting a few too many newlines

src/modules/users/Config.cpp
```diff
@@ -279,6 +290,30 @@ guessProductName()
...
+
+    transliterator->transliterate( transliterable );
+
+    return QString::fromUtf16( transliterable.getTerminatedBuffer() );
+
+}
+#endif
```
adriaandegroot, line 320: How about adding a
```
#else
static QString transliterate( const QString& input ) { return input; }
```
here so that the API is available (internally, since it's static) always? Then subsequent code is simpler.

src/modules/users/Config.cpp
```diff
@@ -337,8 +372,16 @@ Config::setFullName( const QString& name )
         emit fullNameChanged( name );
 
         // Build login and hostname, if needed
-        QRegExp rx( "[^a-zA-Z0-9 ]", Qt::CaseInsensitive );
-        QString cleanName = CalamaresUtils::removeDiacritics( name ).toLower().replace( rx, " " ).simplified();
+        static QRegExp rx( "[^a-zA-Z0-9 ]", Qt::CaseInsensitive );
+#ifdef HAVE_ICU
+        QString cleanName = transliterate(name);
+        cleanName.replace("'", "");
```
adriaandegroot: It's a bit odd to handle `'` differently (removed in transliteration, while it becomes a space without)

src/modules/users/Config.cpp
```diff
@@ -337,8 +372,16 @@ Config::setFullName( const QString& name )
...
+#else
+        QString cleanName = name;
+#endif
+        cleanName = CalamaresUtils::removeDiacritics( cleanName ).replace( rx, " " ).toLower().simplified();
```
adriaandegroot: If you introduce a dummy `transliterate()` then the flow becomes clearer and more consistent:
```
QString cleanName = transliterate(name).replace("'","");
cleanName = // the diacritics bits
```
or maybe in three lines, so you can see the three steps:
 - transliterate
 - remove punctuation (possibly introduced by transliteration)
 - remove diacritics and split on non-alphanumerics

src/modules/users/Config.cpp
kkofler: But if you happen to use the same literal more than once, the pointer version can be merged by the compiler (or the linker), the array version you suggest cannot.

On the other hand, the pointer version introduces an extra indirection and an extra relocation compared to the array version.

Issue #1311, opened 2020-01-29
Add an option to not encrypt /boot


Bleuzen:
Hi, I would really like to see an option to enable encryption in the installer, but not encrypt /boot.

This has multiple advantages:
- **decryption is much, much faster, because it can use hardware acceleration**
decryption with encrypted /boot (in grub) can take about 20 - 30 seconds on my machine and I'm not alone
- **you have multiple tries to enter your password**
this is not possible in grub currently. If you put in the wrong password, you have to reboot, which is very frustrating, because checking the password takes so long every time
- **the correct keyboard layout is loaded**
if decrypting in grub, the local key map is not loaded and so you can't use special chars in your password. Leaving /boot unencrypted and do the os do the decryption fixes this.
- **luks-2 is currently not supported in grub**
I really like to use luks-2, because, you know, it is 2020 and luks-2 is out for some time already and has some advantages I like to use. With encrypted /boot, this would currently not be possible.

Maybe there are more points, but that are the one that got me to leave /boot unencryped.

Also, all the other distros I tried so far do NOT encrypt /boot too. (examples would be Pop OS / Fedora / ...).

In my opinion, it is not worth it to encrypt boot. You are giving up too much.
If someone is really paranoid, they should store their /boot on an external usb device that they watch all day an night anyway.

So I think a checkbox in Calamares about encrypting /boot or not would be really nice.

philmmanjaro:
Upstream is already working to add [luks-2 support](http://git.savannah.gnu.org/cgit/grub.git/log/), however I was not able to build **grub-git** yet on my end.

Bleuzen:
@philmmanjaro Yes, it is fine that grub will finally support luks-2 soon.
But still this fixes only one of my points against encrypting /boot.
So I still would like to see an option to leave /boot unencrypted.

kkofler:
Leaving `/boot` unencrypted leaves a huge gaping hole in that there is no authentication of the initramfs, so anybody can trojan your initramfs and steal your encrypted data as soon as you unlock it.

The initramfs may also contain sensitive data that you don't want attackers to see, e.g., decryption keys. (In particular, the current Calamares code always puts a LUKS keyfile into the initramfs as soon as you enable LUKS encryption, which is why leaving `/boot` unencrypted in manual partitioning currently defeats the whole purpose of using LUKS, though of course implementing your change request would necessarily change that.)

apt-ghetto:
[...]

Issue #1073, opened 2019-01-17
Booting fails with encrypted / and separate /boot


eirism:
**Describe the bug**
After installing Manjaro xfce 18.0.2 with an encrypted / partition and either encrypted and unencrypted /boot, it fails to boot.
In the case of unencrypted /boot the initramfs cannot find the decrypted root partition, and in the case of encrypted /boot the /boot partition is mounted as /new_root.

Tested in Virtualbox with 64 bit OS and BIOS.

**To Reproduce**
1. Install Manjaro with one partition for /boot (either encrypted or unencrypted) and one encrypted partition for /.
2. Boot the installation
3. See that the boot fails after GRUB.

**Expected behavior**
The installation should boot correctly.

**Additional context**
It seems that #1022 broke this, as it works Kannolo distro which was released before that change. Reverting that change in Manjaro also fixes booting.
The correct fix to #1022 might be to not set `GRUB_ENABLE_CRYPTODISK=y` if /boot is unencrypted. From what I can understand that option makes GRUB able to decrypt LUKS partitions, so removing it will make GRUB unable to try to decrypt /.

philmmanjaro:
@eirism: please post the working `grub.cfg` and broken one. Also check `/etc/default/grub`.

eirism:
Files are from install with unencrypted /boot and encrypted /.

Working, reverted changes done by d775cee before install:
GRUB asks for password for /. "Fancy"/(Manjaro themed) GRUB menu.
[unencrypted-boot-blkid.txt](https://github.com/calamares/calamares/files/2769942/unencrypted-boot-blkid.txt)
[unencrypted-boot-grub.cfg.txt](https://github.com/calamares/calamares/files/2769943/unencrypted-boot-grub.cfg.txt)
[unencrypted-boot-grub-default.txt](https://github.com/calamares/calamares/files/2769944/unencrypted-boot-grub-default.txt)

Not working:
GRUB does not ask for password. Plain/default GRUB menu.
[unencrypted-noboot-blkid.txt](https://github.com/calamares/calamares/files/2769947/unencrypted-noboot-blkid.txt)
[unencrypted-noboot-grub.cfg.txt](https://github.com/calamares/calamares/files/2769948/unencrypted-noboot-grub.cfg.txt)
[unencrypted-noboot-grub-default.txt](https://github.com/calamares/calamares/files/2769949/unencrypted-noboot-grub-default.txt)

philmmanjaro:
[...]

PR #1565, opened 2020-11-03
Use graphical decryption prompt with plymouth

Chrysostomus:


kkofler:
That is not going to be of any use in practice though, because Calamares uses full disk encryption including `/boot` and hence sets up the system to already prompt for the passphrase in GRUB. Everything else is then unlocked with a keyfile stored in the encrypted `/boot` partition.

Any setup where you would get a decryption prompt from Plymouth is **not** supported by Calamares, see #1311.

Chrysostomus:
Aaah, that's why it wasn't implement yet. Thank you for clarification! 😊

hphilm:
We will have the extra hook now, but it won't be used in that case anyway. Only grub will prompt for the password

Chrysostomus:
But if someone uses manual partitioning, they can still benefit.

kkofler:
No, because Calamares will **not** set up a working and secure system if you manually set up `/boot` to be unencrypted and other partitions encrypted. See #1073. The **only** currently supported setup is the one I described.

Chrysostomus:
So, it is a missing feature. Or, considering that the keyfile is always inserted, a bug actually.

abucodonosor:
It's not a missing feature nor a bug, using unencrypted /boot is simply unsupported.

hphilm:
@abucodonosor: I vaguely remember there was an issue or request to support unencrypted `/boot` for some reason. I understand that we currently only support full encryption, but would encryption only of `/home` not be for some use-cases enough?

hphilm:
Seems even Ubuntu switched to **full-encryption** by now: https://mutschler.eu/linux/install-guides/ubuntu-btrfs/

abucodonosor:
@philmmanjaro 

Sure we can switch/offer to manually unlocking, without a key stored anywhere.

But what people wanted calamares to do is to uses a key for automatic unlocking but offer unencrypted /boot
bc of 'fast boot' issues or other weird reasons. That scenario is completely insecure bc your key is stored in the initramfs,
which is stored in /boot which is then unencrypted.

hphilm:
right, that makes no sense in that case ...

Chrysostomus:
Obviously you wouldn't store the key in initramfs if /boot is unencrypted. From what I've seen people want 
- use keyfile in initramfs if /boot is encrypted
- don't use keyfile if only /root is encrypted

I think that's the pretty reasonable expectation of a user who mounts their /boot to esp or another unencrypted partition.

abucodonosor:
@Chrysostomus 

You say that which is correct, but search what people really expect with unencrypted /boot :).
Discussed in a lot of issues, IRC, etc.

> I think that's the pretty reasonable expectation of a user who mounts their /boot to esp or another unencrypted partition.

Unfortunately, that is not the case. Very few people really care or understand the impact of having /boot unencrypted and a key stored there. Also, they don't like to type passwords for each crypted partition. 

Like I've said I personally don't have a problem offering LUKS without a key stored anywhere.

Chrysostomus:
Okay. I can write a PR for supporting a unencrypted /boot without any keyfiles.

abucodonosor:
@Chrysostomus 

sure go ahead, but keep in mind there are some scenarios you need to take care of.

For auto partitioning / replace there needs to be a GUI option too, for manually partitioning there could be a check
for /boot probably. Also, all initramfs implementations we have need adjusting as well as grub* modules etc.

Chrysostomus:
I'm not doing anything to the autopartitioning, I think the defaults are fine as they are. Only manual partitioning needs support for this as far as I'm concerned. I¨m thinking of a simple check if boot is on unencrypted partition. If yes, don't write keyfile, don't insert it into initramfs and don't enable the password query in grub. And don¨t point to the keyfile in crypttab

Bleuzen:
@Chrysostomus well, only because /boot is not encrypted doesn't mean a keyfile can't be used.
Maybe just the place should change. Calamares could put the keyfile on the root partition (if unencrypted /boot but encrypted root is detected).
This way, if a user has multiple enrypted partitions, for example separate / (root) and /home, they don't have to enter the password twice.
They will only unlock the root partition on system startup, where a keyfile is, which is used to decrypt /home.
In this case it would still be useful to have the keyfile. But not in initramfs, only the place has to change.

PR #1566, opened 2020-11-03
Don't use a keyfile for encrypted partitions if /boot in unecrypted

Chrysostomus:
Needed changes for supporting unencrypted /boot partition. It still needs
1) Handling for dracut
2) Skipping writing and adding the keyfile in the first place
I can't do those parts because I don't know C.

But this pull request should still be useful, and avoid
1) Automatic decryption at boot time without password
2) unbootable system due to the misconfigured grub.

### DIFF
--- a/src/modules/grubcfg/main.py
+++ b/src/modules/grubcfg/main.py
@@ -90,9 +90,7 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     swap_outer_uuid = ""
     swap_outer_mappername = None
     no_save_default = False
-    unencrypted_separate_boot = False
-    if any(partition["mountPoint"] == "/boot" and "luksMapperName" not in partition for partition in partitions):
-        unencrypted_separate_boot = True
+    unencrypted_separate_boot = any(p["mountPoint"] == "/boot" and "luksMapperName" not in p for p in partitions):
 
     for partition in partitions:
         if partition["mountPoint"] in ("/", "/boot") and partition["fs"] in ("btrfs", "f2fs"):
--- a/src/modules/grubcfg/main.py
+++ b/src/modules/grubcfg/main.py
@@ -90,7 +90,7 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     swap_outer_uuid = ""
     swap_outer_mappername = None
     no_save_default = False
-    unencrypted_separate_boot = any(p["mountPoint"] == "/boot" and "luksMapperName" not in p for p in partitions):
+    unencrypted_separate_boot = any(p["mountPoint"] == "/boot" and "luksMapperName" not in p for p in partitions)
 
     for partition in partitions:
         if partition["mountPoint"] in ("/", "/boot") and partition["fs"] in ("btrfs", "f2fs"):
--- a/src/modules/grubcfg/main.py
+++ b/src/modules/grubcfg/main.py
@@ -90,6 +90,12 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     swap_outer_uuid = ""
     swap_outer_mappername = None
     no_save_default = False
+    unencrypted_separate_boot = False
+
+    for partition in partitions:
+        if (partition["mountPoint"] == "/boot"
+                and "luksMapperName" not in partition):
+            unencrypted_separate_boot = True
 
     for partition in partitions:
         if partition["mountPoint"] in ("/", "/boot") and partition["fs"] in ("btrfs", "f2fs"):
@@ -239,7 +245,7 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     if not have_distributor_line:
         lines.append(distributor_line)
 
-    if cryptdevice_params:
+    if cryptdevice_params and not unencrypted_separate_boot:
         lines.append("GRUB_ENABLE_CRYPTODISK=y")
 
     with open(default_grub, 'w') as grub_file:
--- a/src/modules/grubcfg/main.py
+++ b/src/modules/grubcfg/main.py
@@ -91,11 +91,8 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     swap_outer_mappername = None
     no_save_default = False
     unencrypted_separate_boot = False
-
-    for partition in partitions:
-        if (partition["mountPoint"] == "/boot"
-                and "luksMapperName" not in partition):
-            unencrypted_separate_boot = True
+    if any(partition["mountPoint"] == "/boot" and "luksMapperName" not in partition for partition in partitions):
+        unencrypted_separate_boot = True
 
     for partition in partitions:
         if partition["mountPoint"] in ("/", "/boot") and partition["fs"] in ("btrfs", "f2fs"):
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -141,7 +141,7 @@ partitions()
 bool
 hasUnencryptedSeparateBoot()
 {
-    const QVariantList partitions = partitions();
+    const QVariantList partitions = ::partitions();
     for ( const QVariant& partition : partitions )
     {
         QVariantMap partitionMap = partition.toMap();
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -201,7 +201,7 @@ LuksBootKeyFileJob::exec()
     // /boot partition is not encrypted, keyfile must not be used
     if ( hasUnencryptedSeparateBoot() )
     {
-        cDebug() << Logger::SubEntry << "/boot partition is not encryptepted, skipping keyfile creation.";
+        cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
         return Calamares::JobResult::ok();
     }
 
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -137,7 +137,8 @@ LuksBootKeyFileJob::partitions()
     return globalStorage->value( QStringLiteral( "partitions" ) ).toList();
 }
 
-static bool
+// static
+bool
 LuksBootKeyFileJob::hasUnencryptedSeparateBoot()
 {
     const QVariantList partitions = LuksBootKeyFileJob::partitions();
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -129,6 +129,30 @@ setupLuks( const LuksDevice& d )
     return true;
 }
 
+// static
+QVariantList
+LuksBootKeyFileJob::partitions()
+{
+    Calamares::GlobalStorage* globalStorage = Calamares::JobQueue::instance()->globalStorage();
+    return globalStorage->value( QStringLiteral( "partitions" ) ).toList();
+}
+
+static bool
+LuksBootKeyFileJob::hasUnencryptedSeparateBoot()
+{
+    const QVariantList partitions = LuksBootKeyFileJob::partitions();
+    for ( const QVariant& partition : partitions )
+    {
+        QVariantMap partitionMap = partition.toMap();
+        QString mountPoint = partitionMap.value( QStringLiteral( "mountPoint" ) ).toString();
+        if ( mountPoint == QStringLiteral( "/boot" ) )
+        {
+            return !partitionMap.contains( QStringLiteral( "luksMapperName" ) );
+        }
+    }
+    return false;
+}
+
 Calamares::JobResult
 LuksBootKeyFileJob::exec()
 {
@@ -171,6 +195,13 @@ LuksBootKeyFileJob::exec()
     {
         // Then there was no root partition
         cDebug() << Logger::SubEntry << "No root partition.";
+        return Calamares::JobResult::ok();
+    }
+
+    // /boot partition is not encrypted, keyfile must not be used
+    if ( hasUnencryptedSeparateBoot() )
+    {
+        cDebug() << Logger::SubEntry << "/boot partition is not encryptepted, skipping keyfile creation.";
         return Calamares::JobResult::ok();
     }
 
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -131,7 +131,7 @@ setupLuks( const LuksDevice& d )
 
 // static
 QVariantList
-LuksBootKeyFileJob::partitions()
+partitions()
 {
     Calamares::GlobalStorage* globalStorage = Calamares::JobQueue::instance()->globalStorage();
     return globalStorage->value( QStringLiteral( "partitions" ) ).toList();
@@ -139,9 +139,9 @@ LuksBootKeyFileJob::partitions()
 
 // static
 bool
-LuksBootKeyFileJob::hasUnencryptedSeparateBoot()
+hasUnencryptedSeparateBoot()
 {
-    const QVariantList partitions = LuksBootKeyFileJob::partitions();
+    const QVariantList partitions = partitions();
     for ( const QVariant& partition : partitions )
     {
         QVariantMap partitionMap = partition.toMap();
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.h
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.h
@@ -30,6 +30,9 @@ public:
     QString prettyName() const override;
 
     Calamares::JobResult exec() override;
+private:
+    static QVariantList partitions();
+    static bool hasUnencryptedSeparateBoot();
 };
 
 CALAMARES_PLUGIN_FACTORY_DECLARATION( LuksBootKeyFileJobFactory )
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.h
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.h
@@ -30,9 +30,6 @@ public:
     QString prettyName() const override;
 
     Calamares::JobResult exec() override;
-private:
-    static QVariantList partitions();
-    static bool hasUnencryptedSeparateBoot();
 };
 
 CALAMARES_PLUGIN_FACTORY_DECLARATION( LuksBootKeyFileJobFactory )
--- a/src/modules/openrcdmcryptcfg/main.py
+++ b/src/modules/openrcdmcryptcfg/main.py
@@ -21,10 +21,6 @@ _ = gettext.translation("calamares-python",
                         fallback=True).gettext
 
 
-for partition in partitions:
-        if (partition["mountPoint"] == "/boot"
-                and "luksMapperName" not in partition):
-            unencrypted_separate_boot = True
 
 def pretty_name():
     return _("Configuring OpenRC dmcrypt service.")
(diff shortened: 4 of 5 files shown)


kkofler:
It is not enough to just remove the `key=/crypto_keyfile.bin` line from the crypttab, you also have to change the code that writes the file to begin with and that puts the key into the keyslot. Otherwise, all an attacker has to do is to add that line and voilà, the partition is decrypted without a password.

kkofler:
(This happens in the `luksbootkeyfile` module.)

kkofler:
And yes, the module is in C++. This change cannot be done without C++ knowledge.

kkofler:
As for the `dracutlukscfg` module, that one already has the required logic:
https://github.com/calamares/calamares/blob/1cd9b93a224e3e4bf08a75ed1c86469fb26694ab/src/modules/dracutlukscfg/DracutLuksCfgJob.cpp#L152

So, not only is no change needed there, but you can actually copy the static `partitions` and `hasUnencryptedSeparateBoot` methods to `luksbootkeyfile` and it will do a large part of what is needed for you. It is not like C++ is impossible to learn. :-)

Chrysostomus:
I know it's not enough. However, mkinitcpio module also already includes the same logic as dracut. So the keyfile will not be present in the initcpio, but stored within the encrypted root partition. This makes me think that the attack vector is smaller than you make it seem, because the potential exploit requires the attacker to already have unrestricted access to the encrypted / partition to take advantage of the keyfile. 

But yes, not creating the keyfile is a necessity. Since dracut already has the required logic, I'll give it a shot. Thanks for the pointer, I really appreciate it! :-)

src/modules/grubcfg/main.py
```diff
@@ -90,6 +90,12 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     swap_outer_uuid = ""
     swap_outer_mappername = None
     no_save_default = False
+    unencrypted_separate_boot = False
+
+    for partition in partitions:
+        if (partition["mountPoint"] == "/boot"
+                and "luksMapperName" not in partition):
+            unencrypted_separate_boot = True
```
adriaandegroot: Somehow this code just seems icky: what you're expressing is "is any partition an unencrypted boot partition". Python has `any()` for that (and C++ has `std::any_of()`) and the predicate is fairly simple to write .. but nothing I can come up with is much shorter than what you have here :(

src/modules/openrcdmcryptcfg/main.py
```diff
@@ -20,6 +20,12 @@
                         languages=libcalamares.utils.gettext_languages(),
                         fallback=True).gettext
 
+unencrypted_separate_boot = False
```
adriaandegroot: This looks really out-of-place: it's indented wrong, and is outside of any function. I would expect it to live inside the `write_dmcrypt_conf()` function.

adriaandegroot (changes requested):

adriaandegroot:
You can find the bits that wrangle the keyfile with `grep -r crypto_keyfile.bin src/` . The initramfs encrypt hook looks a bit sketchy to me, actually.

The code generating the keyfile is in *src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp*, function `generateTargetKeyfile()` which can use at **least** the following cleanups:
 - there are helper functions for generating random files (see the *machineid* module) rather than running dd in the target system
 - I gather from Kevin's comments that the path for the keyfile may need to be changed

If the path changes, then `setupLuks()` also needs fixing. But that can just be an extra parameter to both functions, and the path calculated in `exec()` based on unencrypted-separate-boot.

src/modules/grubcfg/main.py
Chrysostomus: Yes, it is ugly. Also duplicated code from the initramfs module. I can see if I can come up with something better.

kkofler: Personally, I consider this standard iterative pattern no harder to read than the functional version. My C++ version in `dracutlukscfg` is almost the same, except that I actually break the loop after an unencrypted separate `/boot` is found, which is missing in the above Python code.

kkofler:
> I gather from Kevin's comments that the path for the keyfile may need to be changed

Well, actually, the keyfile is written to the target's `/`, which should also be encrypted, and then the initramfs generator (one of the supported ones) is configured to include the file in the initramfs, which is written to `/boot`. So @Chrysostomus is right that the keyfile should actually not end up on an unencrypted `/boot` after this, but that we should still not create a keyfile to begin with when `/boot` is unencrypted.

src/modules/grubcfg/main.py
Chrysostomus: I updated to use a generator expression instead to look more pretty

src/modules/openrcdmcryptcfg/main.py
Chrysostomus: Should be fixed now

Chrysostomus:
The requested changes were made. I'll try to look at the keyfile generation tomorrow.

Chrysostomus:
It seems unfortunately that C++ is still a bridge too far for me. I could write a workaround in python to remove and delete the keyfile, but I don't believe that would be a good approach. I'll see if I can find someone who can help with this. 

I still suggest merging the python changes, as they are beneficial even without the keyfile fix.

src/modules/openrcdmcryptcfg/main.py
```diff
@@ -46,7 +49,9 @@ def write_dmcrypt_conf(partitions, root_mount_point, dmcrypt_conf_path):
             with open(os.path.join(root_mount_point, dmcrypt_conf_path), 'a+') as dmcrypt_file:
                 dmcrypt_file.write("\ntarget=" + crypto_target)
                 dmcrypt_file.write("\nsource=" + crypto_source)
-                dmcrypt_file.write("\nkey=/crypto_keyfile.bin")
+                # Don't use keyfile if boot is unecrypted, keys must not be stored on unencrypted partitions
```
apt-ghetto: typo: unecrypted => unencrypted

Chrysostomus: Fixed

Chrysostomus:
Okay, I tried to avoid writing the keyfile. Not sure if this even compiles, I'll try to test it a bit... But it's really a poke in the dark.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -174,6 +198,13 @@ LuksBootKeyFileJob::exec()
         return Calamares::JobResult::ok();
     }
 
+    // /boot partition is not encrypted, keyfile must not be used
+    if ( hasUnencryptedSeparateBoot() )
+    {
+        cDebug() << Logger::SubEntry << "/boot partition is not encryptepted, skipping keyfile creation.";
```
apt-ghetto: typo: encryptepted

Chrysostomus: Thanks, fixed

Bleuzen:
To be honest, I don't really like this approach. It works, as long as there is only one encrypted partition, but can get annoying if there are multiple encrypted partitions (for example separate / and /home, which is pretty common).

I think the keyfile should not be removed / not generated, but only the place should be changed.
What about not storing the keyfile in initramfs, but on the root partition? (only if root is encrypted and /boot is not)

This way, if users have separate / and /home, they don't have to enter the password twice.
After first password prompt, the root partition will be unlocked, where the keyfile is, which is used to automatically unlock the /home partition.

Chrysostomus:
> To be honest, I don't really like this approach. It works, as long as there is only one encrypted partition, but can get annoying if there are multiple encrypted partitions (for example separate / and /home, which is pretty common).
> 
> I think the keyfile should not be removed / not generated, but only the place should be changed.
> What about not storing the keyfile in initramfs, but on the root partition? (only if root is encrypted and /boot is not)
> 
> This way, if users have separate / and /home, they don't have to enter the password twice.
> After first password prompt, the root partition will be unlocked, where the keyfile is, which is used to automatically unlock the /home partition.

The keyfile is already stored in / and not copied to initramfs if /boot is unencrypted. But if you generate the keyfile, there probably should be some logic to account for the scenario where only /home is encrypted. Adding keyfile after the fact is not that hard.

abucodonosor:
> > To be honest, I don't really like this approach. It works, as long as there is only one encrypted partition, but can get annoying if there are multiple encrypted partitions (for example separate / and /home, which is pretty common).
> > I think the keyfile should not be removed / not generated, but only the place should be changed.
> > What about not storing the keyfile in initramfs, but on the root partition? (only if root is encrypted and /boot is not)
> > This way, if users have separate / and /home, they don't have to enter the password twice.
> > After first password prompt, the root partition will be unlocked, where the keyfile is, which is used to automatically unlock the /home partition.
> 
> The keyfile is already stored in / and not copied to initramfs if /boot is unencrypted. But if you generate the keyfile, there probably should be some logic to account for the scenario where only /home is encrypted. Adding keyfile after the fact is not that hard.

Strongly disagree with all this. Either full like now with a key stored into initramfs or without any keyfile stored, any other scenario can be changed after installation by the user, eg: storing the key on a USB or even adding one into initramfs, not our problem at this point if people want to compromise their installation.

Chrysostomus:
I would agree, using keyfiles without fde is not a very good idea.

Chrysostomus:
Does not seem to compile:

```
LuksBootKeyFileJob.cpp:141:48: error: cannot declare member function ‘bool LuksBootKeyFileJob::hasUnencryptedSeparateBoot()’ to have static linkage [-fpermissive]
  141 | LuksBootKeyFileJob::hasUnencryptedSeparateBoot()
      |                                                ^
```
Not sure what exactly is the issue. I pretty much copied the dracut config modules and just changed the job name to match. Removing the static keyword from the line does not seem to change anything either. It would be really useful if someone who actually knew what they were doing would take a look this 😅

kkofler:
You have to also add the function (method) to the class definition, which is in the corresponding header file.

kkofler:
See: https://www.w3schools.com/cpp/cpp_class_methods.asp
The "Outside Example" is the one that matters here. Putting everything inside the class definition is what programming languages like Java always do, but in C++, it implicitly makes the function inline and it also puts it in the header file, slowing down the compilation, so it is only done for trivial methods (such as a getter that only returns a member variable unprocessed). The normal way in C++ is the "outside" way.

Chrysostomus:
Thank you!

Chrysostomus:
EDIT: found the issue, extra static keyword in the function definition is what kept it from compiling.

Chrysostomus:
And now it compiles. So it should be ready for a review, I think.

src/modules/grubcfg/main.py
```diff
@@ -90,6 +90,9 @@ def modify_grub_default(partitions, root_mount_point, distributor):
     swap_outer_uuid = ""
     swap_outer_mappername = None
     no_save_default = False
+    unencrypted_separate_boot = False
+    if any(partition["mountPoint"] == "/boot" and "luksMapperName" not in partition for partition in partitions):
+        unencrypted_separate_boot = True
```
adriaandegroot: Since `any()` returns a boolean, you can write this shorter as
```
unencrypted_separate_boot = any(partition["mountPoint"] == "/boot" and "luksMapperName" not in partition for partition in partitions)
```
(maybe rename *partition* to *p* to make the line shorter, which is ok since *p* is only in the scope of this list comprehension)

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -129,6 +129,31 @@ setupLuks( const LuksDevice& d )
     return true;
 }
 
+// static
+QVariantList
+LuksBootKeyFileJob::partitions()
+{
+    Calamares::GlobalStorage* globalStorage = Calamares::JobQueue::instance()->globalStorage();
+    return globalStorage->value( QStringLiteral( "partitions" ) ).toList();
+}
```
adriaandegroot, line 138: Congratulations on writing your first C++! Now I see where your earlier confusion about `static` came from. This function does not use any of the class members; it doesn't do anything that is special to the class. So you can remove the `LuksBootKeyFileJob::` there: that makes it a "free function" not associated with any class. Then you remove the declaration of the function from the header, too.

The thing with free functions is that they're "visible" on the outside, and sometimes they are not interesting, useful or whatever for the outside. In order to hide them, C++ has a variety of things you can use. In Calamares, we write `static` in front of free functions that should be local to the C++ file they are defined in.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -129,6 +129,31 @@ setupLuks( const LuksDevice& d )
...
+        QVariantMap partitionMap = partition.toMap();
+        QString mountPoint = partitionMap.value( QStringLiteral( "mountPoint" ) ).toString();
+        if ( mountPoint == QStringLiteral( "/boot" ) )
+        {
+            return !partitionMap.contains( QStringLiteral( "luksMapperName" ) );
+        }
+    }
+    return false;
+}
+
```
adriaandegroot, line 156: The same applies to this function: it doesn't use the class members, **except** for the `partitions()` function defined above. If you make that function above a free function, then this one can become a free function too.

The advantage here is that you can refactor parts of the program so that it reads like a story, while hiding the implementation details and names from the outside world.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.h
```diff
@@ -30,6 +30,9 @@ class PLUGINDLLEXPORT LuksBootKeyFileJob : public Calamares::CppJob
     QString prettyName() const override;
 
     Calamares::JobResult exec() override;
+private:
+    static QVariantList partitions();
+    static bool hasUnencryptedSeparateBoot();
```
adriaandegroot: These declarations would go away if you make them free functions.

src/modules/openrcdmcryptcfg/main.py
```diff
@@ -21,13 +21,17 @@
...
 def pretty_name():
     return _("Configuring OpenRC dmcrypt service.")
 
 
 def write_dmcrypt_conf(partitions, root_mount_point, dmcrypt_conf_path):
     crypto_target = ""
     crypto_source = ""
+    unencrypted_separate_boot = False
+    if any(partition["mountPoint"] == "/boot" and "luksMapperName" not in partition for partition in partitions):
+        unencrypted_separate_boot = True
```
adriaandegroot: You can use a simpler call to `any()` here.

adriaandegroot:
Let's take a step back to see what this PR does, now.

- If `/boot` exists but is itself not encrypted, no key file is written and no encryption is configured in grub or elsewhere.

What scenario's can lead to that? If those scenario's include encrypted partitions, where / how is the key material stored for them?

Chrysostomus:
The encryption is still configured in initramfs, just without keyfile. Initramfs provides the password dialog for decrypting the root partition, with gui if Plymouth is installed. This has been previously requested by multiple distros.

Compared with grub, initramfs decryption has the following advantages:

- it uses hardware acceleration, so decryption can be up to 10 times faster (on my previous hardware 4 seconds vs 56 seconds) 
- possibility to use graphical password dialog
- use your actual keymap to input password! No more the awkwardness of trying to input cyrillic password with us keymap

The main scenario where this would happen is that user mounts their esp partition to /boot. This is relatively common in single boot systems and required by some bootloaders like refind and systemd-boot. 

The only scenario where this pull request hampers the user experience is if user chooses the compilation of unencrypted /boot, encrypted / and separate encrypted /home partition. In that case, the user would have to input their password twice on boot, until they manually setup the keyfile to unlock /home. But I would argue that it is more of a fringe use case.

src/modules/grubcfg/main.py
Chrysostomus: That's a great idea, it makes it look quite a bit prettier!

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
Chrysostomus: Great! Not knowing C++, I did not know that you can write free functions, but that is definitely preferable here.

Chrysostomus:
I have made the requested changes. Thanks!

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
kkofler: In my defense (since I wrote the code from `dracutlukscfg` on which you based this, using private static member functions), I would like to point out that several C++ developers consider file-scope functions (free functions declared as `static`) deprecated, and that using private static member functions is the object-oriented approach. On the other hand, the static member functions do pollute the header file with private functions that do not need to be there, so I am not opposed to using free functions instead. We can change `dracutlukscfg` as well. Or we could even make the functions truly global rather than file-scope, to avoid the code duplication (but then they would have to be somewhere in `libcalamares`, I guess).

Chrysostomus:
The code does not actually compile after the suggested changes. I suspect I made a typo somewhere. The call for function partitions fails:
```

/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp: In function ‘bool hasUnencryptedSeparateBoot()’:
/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp:144:48: error: no match for call to  (const QVariantList {aka const QList<QVariant>}) ()’
  144 |     const QVariantList partitions = partitions();
      |                              
```
I'll see if I can fix it. If not, I'll maybe revert the commit, as it worked before it.

Chrysostomus:
Fixed, now it builds

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
Chrysostomus: Sharing them globally and running only once would probably the cleanest way. But it is a very small check, so it doesn't probably matter that much.

Chrysostomus:
Testing the installation with this patch in kvm, legacy boot and automatic encryption (which this patch should not affect), I get some errors:

```
00:26:29 [6]: static CalamaresUtils::ProcessResult CalamaresUtils::System::runCommand(CalamaresUtils::System::RunLocation, const QStringList&, const QString&, const QString&, std::chrono::seconds)
    Running "chroot" ("/tmp/calamares-root-gjodf8db", "dd", "bs=512", "count=4", "if=/dev/urandom", "of=/crypto_keyfile.bin") 
    ..  Finished. Exit code: 1 
    ..  Target cmd: ("dd", "bs=512", "count=4", "if=/dev/urandom", "of=/crypto_keyfile.bin") output:
 dd: failed to open '/dev/urandom': No such file or directory
00:26:29 [2]: bool generateTargetKeyfile()
    WARNING: Could not create LUKS keyfile: "dd: failed to open '/dev/urandom': No such file or directory" (exit code 1 ) 

```
It does not seem to be related to this PR though. Exploring further.

Chrysostomus:
Okay, the C++ part seems to work as intended when I tested it, but there is a syntax error in the python. Fixing it tomorrow.

Chrysostomus:
Fixed, I had forgotten remove the colons when it the generators stopped being if clauses.

kkofler:
That's the mean thing about Python. In C++, this kind of errors gets caught immediately by the compiler.

Though this is a syntax error and so would at least be caught by the Python bytecompilation, if that were run somewhere during Calamares builds or CI (hint hint… ;-) ). There are also classes of errors that would be caught at compile time in C++ but even the Python bytecompilation process won't catch, such as type mismatches (due to weak typing) or misspelled method names (due to "duck typing").

Chrysostomus:
In theory, those issues could we addressed if we converted python modules to cython. Types would be consistent and syntax errors would show at compile time, while the modules would still be easy to write to people who only know python. 

Although, the nice thing about python modules is that I can modify them at runtime and test if the actual functionality is right without needing to recompile the whole package.

Chrysostomus:
I tested the patches version in kvm. Normal automatic partitioning is unaffected as it should, and creating unencrypted /boot in manual partitioning works as intended. Plymouth-encrypt is used, no keyfile was created and initcpio handled the decryption as expected.

Chrysostomus:
@adriaandegroot are further changes needed before merging this pull request? It has now been tested to work and not break default setup.

If user chooses unencrypted /boot, they still get the warning dialog about it after manual partitioning. Now that it actually produces a reasonable setup, that could be removed, but that is not necessary if we want to advocate the usage of full disk encryption as before. I think either option is fine.

adriaandegroot:
Code-wise this looks fine. Thank you for taking all the stylistic recommendations up.

adriaandegroot:
Note that now we need to turn to documentation, to make sure that this feature is clear when used (although as you point out, you can only reach this scenario through manual partitioning, where we let people shoot themselves in the foot already)

PR #1571, opened 2020-11-09
[unpackfs] Skip overlay extended attributes

gportay:
Hello,

This PR intends to fix mount issues if using an overlay filesystem, after rebooting the target.

The module `unpackfs` preserves the extended attributes at `rsync` and the overlay filesystem stores extended attributes by inodes.

The overlay filesystem keeps traces of the lower directory by encoding and storing its `UUID` to the attribute `trusted.overlay.origin`.

If the index feature is on, that attribute is compared to the `UUID` of the __lower directory__ at every subsequent mount and causes `mount` to fail with **ESTATE** if it does not match.

This filters the namespace `trusted.overlay.*` by using the `rsync` option `'--filter=-x trusted.overlay.*'` to make sure the overlays extended attributes are not preserved.

Fixes:

```
# mount -t overlay -o lowerdir=...,upperdir,...,workdir= overlay /mnt/etc
mount: /var/mnt/etc: mount(2) system call failed: Stale file handle.
```

```
# dmesg
(...)
overlayfs: "xino" feature enabled using 32 upper inode bits.
overlayfs: failed to verify origin (/etc, ino=524292, err=-116)
overlayfs: failed to verify upper root origin
```

### DIFF
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -181,7 +181,7 @@ def file_copy(source, entry, progress_cb):
     num_files_total_local = 0
     num_files_copied = 0  # Gets updated through rsync output
 
-    args = ['rsync', '-aHAXr']
+    args = ['rsync', '-aHAXr', '--filter=-x trusted.overlay.*']
     args.extend(global_excludes())
     if entry.excludeFile:
         args.extend(["--exclude-from=" + entry.excludeFile])


adriaandegroot:
Are there ever cases where it **does** make sense to keep the *trusted.overlay* xattrs? (Probably not, during installation: the way things are mounted during installation does not necessarily reflect how things are mounted in production after reboot)

gportay:
Yes, I guess there are cases that "make sense" to keep those _xattrs_. Anyway, I assume these attributes are rebuilt after the next mount. [EDIT: they are]

For example, in the case of the `lowerdir` is a filesystem that is _dd' ed_ from the host to the target (by the module `rawfs` ). In this case, the _UUID_ is left unchanged. In this situation, the `upper` filesystem can have its _xattrs_ preserved. However, the _UUID_ filesystem is not anymore unique :shrug:, but this is inherent to the module `rawfs`.

That is why I do think the module `unpackfs` should be more modular and accepts a new attribute that allows us to pass extra `rsync` options.

```
unpack:
    -   source: ../CHANGES
        sourcefs: file
        destination: "/tmp/changes.txt"
        weight: 1  # Single file
        rsyncOptions: "--filter -x trusted.overlay.*"
```

This solution is very flexible, one can set any option he needs and calamares is independent from the will of the user and the version of `rsync` that is installed (new options can be introduced in the future).

But this solution does not feat well with the existing attributes `exclude` and `excludeFile` that expand the `rsync` command-line with `--exclude` and `--exclude-from`.

Therefore, I was thinking of adding the new boolean attribute `xattrs` (set `-X` if `True`) and `xattrsExclude` to filter out for xattr namespaces.

```
unpack:
    -   source: ../CHANGES
        sourcefs: file
        destination: "/tmp/changes.txt"
        weight: 1  # Single file
        xattrs: True # -X
        xattrsExclude: trusted.overlay.* # "--filter -x trusted.overlay.*"
```

But `rsync` accepts filtering out and filtering in, thus better to add these two attributes `xattrFilterOut` and `xattrFilterIn` but I have not tested these two `--filter` expression together so far and thus I do not know if we can mix them.

```
unpack:
    -   source: ../CHANGES
        sourcefs: file
        destination: "/tmp/changes.txt"
        xattrsFilterOut: trusted.overlay.* # "--filter -x trusted.overlay.*"
        xattrsFilterIn: trusted.overlay.* # "--filter -x! user.*"
```
That is why I came with that very first hardcoding solution and posted an RFC to start the conversation.

adriaandegroot:
I get the feeling that this is headed towards "support an arbitrary command-line for rsync". Which might, to some extent, make sense for specialist uses (like yours). If we were to define a few *substitutions* (e.g, `${excludeExtraMount}` and other parts of the current command-line, then someone with specialist needs could use those -- or ignore them -- and we could define default arguments for rsync as `-aHAXr ${excludeExtraMounts} ${excludeFrom} ${exclude} --progress ${source} ${dest}` and do the substitution systematically.

gportay:
> I get the feeling that this is headed towards "support an arbitrary command-line for rsync".

This is the more flexible we can do.

gportay:
Also, I hit a new issue with the `rsync` options yesterday.

The docker daemon fell back to the `devicemapper` backend that has the consequence of creating a 100G data-sparse file at path `/var/lib/docker/devicemapper/devicemapper/data`.

Of course, the target file-system was too narrow to store it causes `rsync` to fail with the error below:

```
            18:03:36 [6]: static CalamaresUtils::ProcessResult CalamaresUtils::System::runCommand(CalamaresUtils::System::RunLocation, const QStringList&, const QString&, const QString&, std::chrono::seconds)
                Running "env" ("sync")
                ..  Finished. Exit code: 0
                ..  Target cmd: ("sync") output:

            rsync: [receiver] write failed on "/tmp/calamares-root-81qie5d1/var/lib/docker/devicemapper/devicemapper/data": No space left on device (28)
            rsync error: error in file IO (code 11) at receiver.c(378) [receiver=v3.2.3]
```

I have fixed the source of that issue, I have created that [change](https://github.com/gportay/calamares/commit/4aca5a0e30b390d45451d36634e93136ce3ddfbc) as well that adds the options `-S`. Do you want me to make a PR for that?

I guess that this option should be a part of the hard-coded options in the module `unpackfs`, but I have not a strong opinion about that.

Or should we think about a more modular way for users to set `rsync` options as a new attribute in `module.conf` `rsyncOption: -S --filter -x trusted.overlay.*` that is really easy to implement or something with more abstraction like options `sparse: True`, `xattrs: True`, `xattrsFilterOut: trusted.overlay.*` and `xattrsFilterIn: user.*` that is not complicated to implement but is fewer versatile; every option must be implemented.

kkofler:
Using `-S` sounds reasonable, but I wonder what the performance impact is.

adriaandegroot:
Merging it, let's think about the more general substitutions case shortly, maybe before taggin 3.2.36

PR #1572, opened 2020-11-10
[partition] output more things in overiew

gportay:
Hello,

This RFC intends to output more things in the overview, such as the file-system features or the GPT entries (name, UUID, attributes, type).

It adds the new helper `CalamaresUtils::getPartitionTable()`.

It adds the filesystem `features` to the global storage.

It introduces a mapping between the GPT types and a more human-friendly string (it may go to kpmcore?).

What do you think?

Regards,
Gaël

### DIFF
--- a/src/libcalamares/partition/PartitionQuery.cpp
+++ b/src/libcalamares/partition/PartitionQuery.cpp
@@ -16,6 +16,7 @@
 
 #include <kpmcore/core/device.h>
 #include <kpmcore/core/partition.h>
+#include <kpmcore/core/partitiontable.h>
 
 namespace CalamaresUtils
 {
@@ -25,6 +26,19 @@ namespace Partition
 // Types from KPMCore
 using ::Device;
 using ::Partition;
+
+const PartitionTable*
+getPartitionTable( const Partition* partition )
+{
+    const PartitionNode* root = partition;
+    while ( root && !root->isRoot() )
+    {
+        root = root->parent();
+    }
+
+    return dynamic_cast< const PartitionTable* >( root );
+}
+
 
 bool
 isPartitionFreeSpace( const Partition* partition )
--- a/src/modules/partition/jobs/CreatePartitionJob.cpp
+++ b/src/modules/partition/jobs/CreatePartitionJob.cpp
@@ -12,6 +12,7 @@
 #include "CreatePartitionJob.h"
 
 #include "partition/FileSystem.h"
+#include "partition/PartitionQuery.h"
 #include "utils/Logger.h"
 #include "utils/Units.h"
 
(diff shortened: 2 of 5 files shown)


src/modules/partition/jobs/CreatePartitionJob.cpp
```diff
@@ -32,9 +33,134 @@ CreatePartitionJob::CreatePartitionJob( Device* device, Partition* partition )
...
+    { "933ac7e1-2eb4-4f13-b844-0e14e2aef915", "Home Partition" },
+    { "3b8f8425-20e0-4f3b-907f-1a25a76f98e8", "Server Data Partition" },
+    { "4d21b016-b534-45c2-a9fb-5c16e091fd2d", "Variable Data Partition" },
+    { "7ec6f557-3bc5-4aca-b293-16ef5df639d1", "Temporary Data Partition" },
+    { "0657fd6d-a4ab-43c4-84e5-0933c84b4f4f", "Swap" },
+    { "c12a7328-f81f-11d2-ba4b-00a0c93ec93b", "EFI System Partition" },
+    { "bc13c2ff-59e6-4262-a352-b275fd6f7172", "Extended Boot Loader Partition" },
+    { "0fc63daf-8483-4772-8e79-3d69d8477de4", "Other Data Partitions" },
+    { "ebd0a0a2-b9e5-4433-87c0-68b6b72699c7", "Microsoft basic data" },
+};
```
adriaandegroot, line 66: This table is going to be problematic for translations, but that is something we can tackle later .

src/libcalamares/partition/PartitionQuery.cpp
```diff
@@ -26,6 +27,19 @@ namespace Partition
 using ::Device;
 using ::Partition;
 
+const PartitionTable*
+getPartitionTable( const Partition* partition )
+{
```
adriaandegroot, line 32: This is a useful factoring-out, thanks.

src/modules/partition/jobs/CreatePartitionJob.cpp
adriaandegroot: Also, just in terms of visible strings: do we need the more obscure ones (all those Verity types, where do they come from?)

src/modules/partition/jobs/CreatePartitionJob.cpp
```diff
@@ -32,9 +33,134 @@ CreatePartitionJob::CreatePartitionJob( Device* device, Partition* partition )
...
+        if ( attributes & ( 1UL << bit ) )
+        {
+            list += QString( "Bit" ) + QString::number( bit );
+        }
+    }
+
+    QString string = list.join( QStringLiteral( " and " ) );
+    if ( string.size() > 64 )   
+    {
+        return QString( "0x" ) + QString::number( attributes, 0x10 );
```
adriaandegroot: This seems a bit dramatic: perhaps keep the human-readable bits and hex the rest, rather than **also** dumping the human-readable part overboard?

src/modules/partition/jobs/CreatePartitionJob.cpp
```diff
@@ -32,9 +33,134 @@ CreatePartitionJob::CreatePartitionJob( Device* device, Partition* partition )
...
+
+    if ( attributes & ( 1UL << 1 ) )
+    {
+        list += QString( "NoBlockIOProtocol" );
+    }
+
+    if ( attributes & ( 1UL << 2 ) )
+    {
+        list += QString( "LegacyBIOSBootable" );
+    }
```
adriaandegroot: Human-readable part goes to here -- should these flag names be translated?

src/modules/partition/jobs/CreatePartitionJob.cpp
```diff
@@ -32,9 +33,134 @@ CreatePartitionJob::CreatePartitionJob( Device* device, Partition* partition )
...
+        list += attributes;
+    }
+
+    return list.join( QStringLiteral( ", " ) );
+}
+
 QString
 CreatePartitionJob::prettyName() const
 {
+    if ( CalamaresUtils::Partition::getPartitionTable( m_partition )->type() == PartitionTable::TableType::gpt )
```
adriaandegroot: The function can return nullptr (it's a screwed up system then, but still).

src/modules/partition/jobs/CreatePartitionJob.cpp
```diff
@@ -32,9 +33,134 @@ CreatePartitionJob::CreatePartitionJob( Device* device, Partition* partition )
...
+    {
+        QString entries = prettyGptEntries( m_partition );
+        if ( entries.size() )
+        {
+             entries = QString( " with entries " ) + entries;
+        }
+
+        return tr( "Create new %1MiB partition on %3 (%2)%4." )
```
adriaandegroot: This really needs some translator hints (and maybe a space after `)`?)

adriaandegroot: .. ah, I see that %4 is the bit of the sentence that describes the entries, and if it is not empty, it starts with a space. This is an unpleasant thing to do with translations, so it probably needs some more code to handle the zero / one / many cases of entries.

src/modules/partition/jobs/CreatePartitionJob.cpp
```diff
@@ -46,6 +172,21 @@ CreatePartitionJob::prettyName() const
 QString
 CreatePartitionJob::prettyDescription() const
 {
+    if ( CalamaresUtils::Partition::getPartitionTable( m_partition )->type() == PartitionTable::TableType::gpt )
+    {
+        QString entries = prettyGptEntries( m_partition );
+        if ( entries.size() )
```
adriaandegroot: More idiomatic would be `!entries.isEmpty()`, since it's a string. The same comments about translatability apply here.

adriaandegroot:
Who is this extra information **for**? That's a bit fuzzy for me, since to my mind the user is either a power user doing manual partitioning -- so they know what they just picked, and/or can look in the log file -- or using the automatic layouts from Calamares, and unlikely to be bothered by GPT flags.

src/modules/partition/jobs/CreatePartitionJob.cpp
gportay: I do need them, the output is less nice without them. Anyway, I think this part may go to kpmcore.

src/libcalamares/partition/PartitionQuery.cpp
gportay: Feel free to cherry-pick the commit to master.

src/modules/partition/jobs/CreatePartitionJob.cpp
gportay: Yep, this is another solution. I have not make my mind yet.

Normally no one use these bits, at least for now and it concerns only the users of `partitionLayout`.

There are only 2 bits (+ 1 for ESP) used by [systemd](https://www.freedesktop.org/software/systemd/man/systemd-gpt-auto-generator.html) (see Table 2. Partition Attributes):

Name | Value | Applicable to | Explanation
-- | -- | -- |  --
GPT_FLAG_READ_ONLY | 0x1000000000000000 | /, /home/, /srv/, /var/, /var/tmp/, Extended Boot Loader Partition | Partition is mounted read-only
GPT_FLAG_NO_AUTO | 0x8000000000000000 | /, /home/, /srv/, /var/, /var/tmp/, Extended Boot Loader Partition | Partition is not mounted automatically
GPT_FLAG_NO_BLOCK_IO_PROTOCOL | 0x0000000000000002 | EFI System Partition (ESP) | Partition is not mounted automatically

4 bits used by [Microsoft](https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_entries_(LBA_2%E2%80%9333)) to define the type-specific attributes for basic data partition, aka `EBD0A0A2-B9E5-4433-87C0-68B6B72699C7` (See Partition entries (LBA 2–33), Basic data partition attributes):

Bit | Content
-- | --
60 | Read-only
61 | Shadow copy (of another partition)
62 | Hidden
63 | No drive letter (i.e. do not automount)

And there are 9 others bits used by [Google](https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_entries_(LBA_2%E2%80%9333)) to define the type-specific attributes for Chrome OS kernel (See Partition entries (LBA 2–33), Chrome OS kernel partition attributes):

Bit | Content
-- | --
56 | Successful boot flag
55–52 | Tries remaining
51–48 | Priority (15: highest, 1: lowest, 0: not bootable)

src/modules/partition/jobs/CreatePartitionJob.cpp
gportay: As far as I know, `sfdisk` does not have a translation for the flags.

src/modules/partition/jobs/CreatePartitionJob.cpp
gportay: Good catch

src/modules/partition/jobs/CreatePartitionJob.cpp
gportay: Oh, I was not aware of that.

gportay:
> Who is this extra information **for**? That's a bit fuzzy for me, since to my mind the user is either a power user doing manual partitioning -- so they know what they just picked, and/or can look in the log file -- or using the automatic layouts from Calamares, and unlikely to be bothered by GPT flags.

It is for the users of `partitionLayout` with custom partitions that uses a GPT table.

TBH, the most useful entries are the label/name and the type.

My idea is that in the GPT world, the type (Home, Boot, Root...) is more important than the name of the file-system.

I wanted to have your opinion, this is why that PR is an RFC.

src/modules/partition/jobs/CreatePartitionJob.cpp
gportay: Make sense.

src/modules/partition/jobs/CreatePartitionJob.cpp
gportay: Now I understand why there are such if/else in the code somewhere to deal with cases; that makes the code no pleasant to me lol.

kkofler: Placeholders (or worse, concatenation in hardcoded order – the placeholders can at least be reordered) are generally a possible source of trouble for translations, they should only be used where they make linguistic sense.

I have seen worse abuses in other projects (usually in translator complaints), such as `tr("item") + (n>1 ? "s" : "")`. (Hint: That cannot possibly work, and no, `tr("s")` does not fix it either. Just see what happens if you translate `item` to `entry` in some `en_*`, and that's not even taking into account all the strange things other languages do with plurals.) Not everything that works in English, or even in European languages in general, is safe for translation.

src/modules/partition/jobs/CreatePartitionJob.cpp
gportay: They comes from systemd, chromeos, linux.

gportay:
I have simplified a bit, by removing some of the entries that are not really important at least for now (attributes, uuid). I add the prefix Linux for the partition type that belongs to the Linux world; BSD has its own type but not listed here.

src/modules/partition/jobs/CreatePartitionJob.cpp
adriaandegroot: We could do this:
```
  // String we already have
  QString overall = tr( "Create new %2MiB partition on %4 (%3) with file system %1." );
  if (!entries.isEmpty()) {
    overall += QStringLiteral(" ") + tr("Special partition entries:") + QStringLiteral(" ") + entries;
  }
  return overall;
```

That makes each call to tr() a single sentence, and avoids translating sentence fragments.

src/modules/partition/jobs/CreatePartitionJob.cpp
```diff
@@ -32,9 +33,92 @@ CreatePartitionJob::CreatePartitionJob( Device* device, Partition* partition )
...
+{
+    return gptTypePrettyStrings.value( type.toLower(), type );
+}
+
+static QString
+prettyGptEntries( const Partition* partition )
+{
+    if ( !partition )
+    {
+        return QString("");
```
adriaandegroot: Just `return QString()` is enough here.

src/modules/partition/jobs/CreatePartitionJob.cpp
```diff
@@ -32,9 +33,92 @@ CreatePartitionJob::CreatePartitionJob( Device* device, Partition* partition )
...
+    if ( table && table->type() == PartitionTable::TableType::gpt )
+    {
+        QString entries = prettyGptEntries( m_partition );
+        if ( !entries.isEmpty() )
+        {
+            return tr( "Create new %1MiB partition on %3 (%2) with entries %4." )
+                       .arg( CalamaresUtils::BytesToMiB( m_partition->capacity() ) )
+                       .arg( m_device->name() )
+                       .arg( m_device->deviceNode() )
+                       .arg( prettyGptEntries( m_partition ) );
```
adriaandegroot: You already have *entries*, no need to call `prettyGpyEntries()` again.

adriaandegroot (approved):

adriaandegroot:
Minor nits notwithstanding, this now looks pretty good: I understand what the descriptions are for and where they come from, and the translations are now less problematic. If you want to take the *[RFC]* off this PR, we can merge it (and then I'll briefly chase, for instance, the table of partition types).

gportay:
After a month and a half, I find the time to address your review. Sorry for the delay.

PR #1578, opened 2020-11-19
[unpackfs] Create empty mountpoint

gportay:
Hello,

This PR aims to not omit the creation of the directories that are mountpoints.

It uses the rsync option `-x` or `--one-file-system`.

Note: The other solution is to exclude the pattern `--exclude mountpoint/*`.

What do you think?

Regards,
Gaël

### DIFF
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -136,23 +136,6 @@ class UnpackEntry:
 ON_POSIX = 'posix' in sys.builtin_module_names
 
 
-def global_excludes():
-    """
-    List excludes for rsync.
-    """
-    lst = []
-    extra_mounts = globalstorage.value("extraMounts")
-    if extra_mounts is None:
-        extra_mounts = []
-
-    for extra_mount in extra_mounts:
-        mount_point = extra_mount["mountPoint"]
-
-        if mount_point:
-            lst.extend(['--exclude', mount_point + '/'])
-
-    return lst
-
 def file_copy(source, entry, progress_cb):
     """
     Extract given image using rsync.
@@ -181,8 +164,7 @@ def file_copy(source, entry, progress_cb):
     num_files_total_local = 0
     num_files_copied = 0  # Gets updated through rsync output
 
-    args = ['rsync', '-aHAXr']
-    args.extend(global_excludes())
+    args = ['rsync', '-aHAXxr']
     if entry.excludeFile:
         args.extend(["--exclude-from=" + entry.excludeFile])
     if entry.exclude:


kkofler:
IIRC (and I was one of those who worked on the `rsync` flags in *unpackfs*), the `-x` option does not do what we want: it checks whether the directory is on the same file system on the **source** of the sync, whereas what matters for us is whether it is on the same file system on the **destination**.

In addition, *unpackfs* runs **after** *mount*, so we cannot create the directories for the mountpoints there, because something is already mounted on top of them.

It is the job of the *mount* job to create those directories and to set useful permissions and xattrs, and it is supposed to already do that (it even runs `chcon` to set the correct SELinux context where applicable). How is that not working for you?

In short, I think this pull request should be closed without merging, and I would like to see the actual issue that made you submit it filed as a bug instead.

adriaandegroot:
This change seems strange to me, too: *extraMounts* comes from the *mount* module, where it is an explicit configuration value: for instance, it causes `/dev` to be mounted. The setup is such that we try to prevent writing things into those specially-extra-mounts.

From the [manpage](https://linux.die.net/man/1/rsync), `-x` is as Kevin says: it affects the *source* recursion, which is generally on one single loopback-mounted squashfs anyway. The current code prevents writing to the mount-points in the target.

gportay:
The thing that puzzles me, is the none creation of the mount-point. I do think it should be created by `unpackfs`.

The other solution would consist of this change.

``` diff

        if mount_point:
-            lst.extend(['--exclude', mount_point + '/'])
+            lst.extend(['--exclude', mount_point + '/*'])

    return lst
```

`rsync` will create the directory but not "populate" it.

For example, we can imagine that we want to create a read-only filesystem, we probably do the following:

``` yaml
- partition # create the partition (usual)
- unpackfs  # rsync the files and directories (usual)
- tunefs    # custom module that turns read-only on (tune2fs -O read-only /dev/sdxY)
- mount     # that can or cannot create the mount-point (I really dunno).
            # if it can, calamares succeed (fine), but the system will no run because the /proc, /sys, /dev
```

Note: this is not my use case, it is weirder :p

kkofler:
You are running the modules in the wrong order. `mount` must run **before** `unpackfs`, because `mount` is what mounts the partition that `unpackfs` unpacks to to begin with. And hence, the directory for the extra mount point must also be created within `mount` (and it already does that!), because, after `mount`, the extra mount point is mounted and hides the directory underneath.

gportay:
> You are running the modules in the wrong order. `mount` must run **before** `unpackfs`, because `mount` is what mounts the partition that `unpackfs` unpacks to to begin with. And hence, the directory for the extra mount point must also be created within `mount` (and it already does that!), because, after `mount`, the extra mount point is mounted and hides the directory underneath.

Ouch, you are right, my fake example is totally wrong...

Back to the issue, what I am trying to solve is the non-creation of the mount-point by the module unpackfs. For sure the module should not iterate over the mount-point but it should be present at least in the file-system (unless rsync resets the attributes `chcon`... set by the module `mount`).

For example, let's say that I want to create two rootfs with calamares, the main one on `/` and a second one on `/tmp/rescue/` which is turned read-only for the purpose of the example. Yes, I know this is very unconventional and calamares is not designed for creating a second file-system.

The sequence would be:

``` yaml
sequence:
- exec
  - partition
  - mount # which mounts the secondary rootfs on /tmp/rescue
  - unpackfs # which unpacks the host rootfs to /tmp/rescue with --exclude /proc --exclude /sys --exclude /dev
  - shellprocess # which turns read-only the file-system on /tmp/rescue
```

In the end, things are fine for the main rootfs, but the second rootfs cannot be boot because the directory `/proc`, `/sys`, `/dev`... are missing. However, this can be easily fixed/worked-around by adding the additional mount-points `/tmp/rescue/proc`, `/tmp/rescue/sys`, `/tmp/rescue/dev`... to the configuration file of the module `mount`.

Simply, I wanted to raise this point, before closing the discussion and the PR. If you think this is not appropriate (mainly because rsync would discard the mount-point attributes), please close it without merging and forget about all of this. Or maybe, if this makes a little bit of sense, the change below may be considered:

``` diff
        if mount_point:
-            lst.extend(['--exclude', mount_point + '/'])
+            lst.extend(['--exclude', mount_point + '/*'])

    return last
```

Sorry for this mess ;)

kkofler:
So now we are getting to the core of the issue. So you want Calamares to create a partition (`/tmp/rescue`) with mountpoint placeholders without actually having it mount anything to them. Indeed, that is not how things are designed (because those mounts are normally needed within the chroot). Now you want to create that rescue root without actually chrooting into it. (I presume that the rescue root contains something set up similarly to the live ISO itself, so that it does not need host-specific setup.)

The problem with your proposed change is that it does the wrong thing for the common case, where `/proc` etc. are actually mounted and so trying to create the directory or setting any properties on it is not going to work. We really do want to exclude the entire directory, not just its contents.

kkofler:
Well, actually, the real bug is entirely different: if we are `rsync`ing to `/tmp/rescue`, then we should only `--exclude` mountpoints starting with `/tmp/rescue`, and remove the `/tmp/rescue` prefix, i.e., if we have `extraMountPoints` `/foo`, `/bar`, and `/tmp/rescue/baz`, then we should `--exclude=/baz/`, not `--exclude=/foo/ --exclude=/bar/ --exclude=/tmp/rescue/baz/`.

gportay:
> Well, actually, the real bug is entirely different: if we are `rsync`ing to `/tmp/rescue`, then we should only `--exclude` mountpoints starting with `/tmp/rescue`, and remove the `/tmp/rescue` prefix, i.e., if we have `extraMountPoints` `/foo`, `/bar`, and `/tmp/rescue/baz`, then we should `--exclude=/baz/`, not `--exclude=/foo/ --exclude=/bar/ --exclude=/tmp/rescue/baz/`.

Exactly, all file-systems are unpacked with the same rsync command, but the command should be specific by mount-points; which is the true source of the bug.

gportay:
> Exactly, all file-systems are unpacked with the same rsync command, but the command should be specific by mount-points; which is the true source of the bug.

I mean only if we consider that the mount-point is another rootfs like in my use case. That is why I said that calamares is not designed for creating a second file-system.

Anyway, we can close the PR I guess.

kkofler:
The pull request indeed needs to be rewritten from scratch, but there is definitely a bug: `unpackfs` accepts multiple file systems to unpack, and it takes a `destination` parameter for each. The exclusions need to be relative to that destination.

Issue #1588, opened 2020-12-04
Links on welcome page of installer pointing to wrong pages


deprov447:
**Describe the bug**
The links "Generic Support" and "Known Issues" on the first page of installer are pointing to wrong pages.

**To Reproduce**
1. Open Calamares
2. Click on 'Generic support' or "Known Issues"
3. See the links that open up in browser

**Expected behavior**
"Generic support" should point to [Calamares Wiki](https://github.com/calamares/calamares/wiki) and "Known Issues" to [issue page](https://github.com/calamares/calamares/issues) on Github.

adriaandegroot:
Fixed by merge

PR #1589, opened 2020-12-04
[Fix] Welcome Page links

deprov447:
"Generic support" and "Known Issues" links fixed + "releaseNotesURL" changed appropriately.

Referencing issue #1588


adriaandegroot:
They **are** just examples, but they should do something useful and correct(-ish). I don't think we should drop the https though, in linking to the news.

PR #1595, opened 2020-12-06
[initcpiocfg] code tidy up

codesardine:


src/modules/initcpiocfg/main.py
```diff
@@ -21,20 +21,19 @@
...
 #   along with Calamares. If not, see <http://www.gnu.org/licenses/>.
 
-import libcalamares
 from libcalamares.utils import debug, target_env_call
 import os
 from collections import OrderedDict
 
 import gettext
-_ = gettext.translation("calamares-python",
+translate = gettext.translation("calamares-python",
```
kkofler: Will the string extractor still recognize the translatable strings with this rename? As far as I know, the name `_` is special in `gettext`, it is needed so that the string extraction tools find the translatable strings.

src/modules/initcpiocfg/main.py
```diff
@@ -180,10 +189,11 @@ def modify_mkinitcpio_conf(partitions, root_mount_point):
...
         hooks.extend(["filesystems"])
 
-    if btrfs == "yes" and cpu['proc0']['vendor_id'].lower() != "genuineintel":
+    cpu_intel = "genuineintel"
+    if btrfs and cpu["proc0"]["vendor_id"].lower() != cpu_intel:
         modules.append("crc32c")
-    elif (btrfs == "yes"
-          and cpu['proc0']['vendor_id'].lower() == "genuineintel"):
+    elif (btrfs
+          and cpu["proc0"]["vendor_id"].lower() == cpu_intel):
```
adriaandegroot: We no longer stick really-closely to PEP8, so you can join those two lines.

src/modules/initcpiocfg/main.py
```diff
@@ -180,10 +189,11 @@ def modify_mkinitcpio_conf(partitions, root_mount_point):
     else:
         hooks.extend(["filesystems"])
 
-    if btrfs == "yes" and cpu['proc0']['vendor_id'].lower() != "genuineintel":
+    cpu_intel = "genuineintel"
```
adriaandegroot: The following code becomes shorter if you do `cpu_is_intel = (cpu["proc0"]["vendor_id"].lower() == "genuineintel")` here.

src/modules/initcpiocfg/main.py
```diff
@@ -21,20 +21,19 @@
...
 
-import libcalamares
 from libcalamares.utils import debug, target_env_call
 import os
 from collections import OrderedDict
 
 import gettext
-_ = gettext.translation("calamares-python",
+translate = gettext.translation("calamares-python",
                         localedir=libcalamares.utils.gettext_path(),
```
adriaandegroot: How is `libcalamares.utils` still in-scope as a name here?

src/modules/initcpiocfg/main.py
```diff
@@ -21,20 +21,19 @@
...
                         languages=libcalamares.utils.gettext_languages(),
                         fallback=True).gettext
 
 
 def pretty_name():
-    return _("Configuring mkinitcpio.")
+    return translate("Configuring mkinitcpio.")
 
 
 def cpuinfo():
```
adriaandegroot: This whole function is called once, and then the information it digs up is used once, in exactly one way (to find the vendor_id of proc0). I think replacing it with a `is_intel_cpu()` that returns bool (and then patching up the one place the information is used) is worthwhile.

src/modules/initcpiocfg/main.py
adriaandegroot: It's not:

```
14:39:57 [6]: virtual Calamares::JobResult Calamares::PythonJob::exec()
    Job file "/home/adridg/src/calamares/build/src/modules/initcpiocfg/main.py" 
Python Error:
 <class 'NameError'> 
 name 'libcalamares' is not defined 
 File "/home/adridg/src/calamares/build/src/modules/initcpiocfg/main.py", line 21, in <module>
    localedir=libcalamares.utils.gettext_path(), 
 
```

src/modules/initcpiocfg/main.py
adriaandegroot: It does not, by default. `_` and `_n` are special (er, maybe `_n` is a Calamares thing). In any case, the translations are read by `xgettext --keyword=_n:1,2 -L python` (full details, see the script `ci/txpush.sh`, which you can run with `sh ci/txpush.sh --no-tx` to just update translations locally).

You'll see that this change drops the corresponding strings from the translation catalogue.

kkofler: I think the default keywords are just `_` and `_N` (note: capital `N`!), where the latter stands for `I18N_NOOP`, i.e., mark this string for translation, but do not actually translate it here (because it is later passed to the translation function as a variable).

kkofler: Or actually, I think the NOOP is `N_`.

codesardine: You right  this needs to be reverted.

src/modules/initcpiocfg/main.py
codesardine: Sorry that was a typo.

src/modules/initcpiocfg/main.py
```diff
@@ -78,31 +77,33 @@ def write_mkinitcpio_lines(hooks, modules, files, root_mount_point):
 
     for i in range(len(mklins)):
         if mklins[i].startswith("HOOKS"):
-            joined_hooks = ' '.join(hooks)
-            mklins[i] = "HOOKS=\"{!s}\"".format(joined_hooks)
+            joined_hooks = " ".join(hooks)
+            mklins[i] = "HOOKS='{!s}'".format(joined_hooks)
```
adriaandegroot: Does `mkinitcpio.conf` understand single-quotes?

src/modules/initcpiocfg/main.py
```diff
@@ -173,20 +181,28 @@ def modify_mkinitcpio_conf(partitions, root_mount_point):
     else:
         hooks.extend(["filesystems"])
 
-    if btrfs == "yes" and cpu['proc0']['vendor_id'].lower() != "genuineintel":
+    def is_cpu_intel():
+        cpu_intel = "genuineintel"
+        if cpu["proc0"]["vendor_id"].lower() == cpu_intel:
+            return True
+        else:
+            return False
```
adriaandegroot: `return cpu["proc0"]["vendor_id"].lower() == "genuineintel"`

Or just make a variable out of it -- this doesn't have to be a function.

src/modules/initcpiocfg/main.py
```diff
@@ -173,20 +181,28 @@ def modify_mkinitcpio_conf(partitions, root_mount_point):
...
+
+    if btrfs and not is_cpu_intel():
         modules.append("crc32c")
-    elif (btrfs == "yes"
-          and cpu['proc0']['vendor_id'].lower() == "genuineintel"):
+
+    elif btrfs and is_cpu_intel():
         modules.append("crc32c-intel")
     else:
+        # What we doing here this one never runs?
```
adriaandegroot: Case `not btrfs` ?

src/modules/initcpiocfg/main.py
abucodonosor: I think this vendor check is not needed, it loads on AMD too.

```
crazy@crazy-nitroan51543:~/kernel/linux$ sudo modprobe btrfs
crazy@crazy-nitroan51543:~/kernel/linux$ dmesg | tail -n1 
[185152.927724] Btrfs loaded, crc32c=crc32c-intel
crazy@crazy-nitroan51543:~/kernel/linux$ uname -a
Linux crazy-nitroan51543 5.9.12-fw1 #1 SMP PREEMPT Wed Dec 2 17:32:29 CET 2020 x86_64 AMD Ryzen 7 3750H with Radeon Vega Mobile Gfx AuthenticAMD GNU/Linux
```

adriaandegroot:
Using f-strings makes this a Calamares 3.3 PR

adriaandegroot:
Merged into *update-python-requirements* branch, which is pre-Calamares-3.3 work on the Python bits.

PR #1597, opened 2020-12-07
[WIP] Btrfs swapfile handling

### LINKED ISSUE #1538: Swapfile on btrfs should be on different subvolume like @swap instead of @
The reason is when there is swapfile on btrfs in /swapfile
It will create problem when you create snapshot of @

To avoid this 

Create swapfile in 
/swap/swapfile

Instead of /swapfile

And also create a @swap subvolume
By default like how @home is created

Chrysostomus:
Closes #1538. If swapfile is chosen with btrfs root, creates subvolume @swap, mounts it to /swap and creates swapfile in /swap/swapfile. 

This will need some refactoring when subvolume layout is made configurable, but at least this will prevent filesystem corruption that might result from snapshotting a swapfile.

Note: not tested yet, opening pull request for purposes of discussion.

### DIFF
--- a/src/modules/fstab/main.py
+++ b/src/modules/fstab/main.py
@@ -386,7 +386,12 @@ def run():
         swap_choice = swap_choice.get( "swap", None )
         if swap_choice and swap_choice == "file":
             # There's no formatted partition for it, so we'll sneak in an entry
-            partitions.append( dict(fs="swap", mountPoint=None, claimed=True, device="/swapfile", uuid=None) )
+            root_partitions = [ p["fs"].lower() for p in partitions if p["mountPoint"] == "/" ]
+            root_btrfs = (root_partitions[0] == "btrfs") if root_partitions else False
+            if root_btrfs:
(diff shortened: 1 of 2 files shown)


Chrysostomus:
I finally found the time to test this. The patch is faulty, it does not create swap at all and no subvolumes besides @ for /.

Chrysostomus:
Okay, now it works, safe to merge.

image

Chrysostomus:
@abucodonosor @adriaandegroot, I think this is ready for a review. If any other changes are needed please let me know.

adriaandegroot (approved):
Overall, LGTM. I would probably pull the determination of swapfile-path (either /swapfile or /swap/swapfile) out into a function so it is easier to stay consistent, but this seems good enough.

Chrysostomus:
@abucodonosor?

abucodonosor (commented):
It looks good to me too.

abucodonosor (approved):

PR #1617, opened 2021-01-15
[WIP] Upload Installation log

deprov447:
Addressing #1182


bill-auger:
could you mention exactly which bits of #1182 you intend to address - the standard 'finished' view already has a button and handling for the paste functionality - the "copy to clipboard" feature is novel; but i dont remember that being requested in the #1182 discussion

other than that, the only thing i can comment on, is commit cc7a6bf - that adds a copyright line to paste.h; but nothing else was added to that file

deprov447:
Okay as I understand I have to remove these commits. @adriaandegroot @bill-auger Do we need that sendURLtoClipboard feature?
Guess we can discuss this with others in issue #1182

PR #1619, opened 2021-01-23
Implementing LogUpload functionality from branding

deprov447:
This PR adds functionality of picking up `uploadServer` signals from `branding.desc` rather than hardcoding them (in ViewManager).

Addressing #1182

### DIFF
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -220,17 +220,14 @@ slideshowAPI: 2
 
 
 # These options are to customize online uploading of logs to pastebins:
-#  - style  :  Defines the kind of pastebin service to be used. ie it can
-#              provide the functionality of controlling privacy of 
-#              paste (in future).Currently only "fiche" servers
-#              are supported. Takes string as input
+#  - type   :  Defines the kind of pastebin service to be used.Currently
+#              it accepts two values:
+#              - none    :    disables the pastebin functionality
+#              - fiche   :    use fiche pastebin server
 #  - url    :  Defines the address of pastebin service to be used.
 #              Takes string as input
 #  - port   :  Defines the port number to be used to send logs. Takes 
 #              integer as input
-#  - enable :  Defines if the functionality is to be used or not. Takes 
-#              bool as input
-logUpload.style   :    "fiche"
-logUpload.url     :    "termbin.com"
-logUpload.port    :    9999
-logUpload.enable  :    true
+uploadServer.type    :    "fiche"
+uploadServer.url     :    "termbin.com"
+uploadServer.port    :    9999
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -219,3 +219,16 @@ slideshow:               "show.qml"
 slideshowAPI: 2
 
 
+# These options are to customize online uploading of logs to pastebins:
+#  - style  :  Defines the kind of pastebin service to be used. Takes
+#              string as input
+#  - url    :  Defines the address of pastebin service to be used.
+#              Takes string as input
+#  - port   :  Defines the port number to be used to send logs. Takes 
+#              integer as input
+#  - enable :  Defines if the functionality is to be used or not. Takes 
+#              bool as input
+logUpload.style   :    "fische"
+logUpload.url     :    "termbin.com"
+logUpload.port    :    9999
+logUpload.enable  :    true
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -228,6 +228,7 @@ slideshowAPI: 2
 #              Takes string as input
 #  - port   :  Defines the port number to be used to send logs. Takes 
 #              integer as input
-uploadServer.type    :    "fiche"
-uploadServer.url     :    "termbin.com"
-uploadServer.port    :    9999
+uploadServer :
+    type :    "fiche"
+    url :      "termbin.com"
+    port :     9999
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -220,15 +220,17 @@ slideshowAPI: 2
 
 
 # These options are to customize online uploading of logs to pastebins:
-#  - style  :  Defines the kind of pastebin service to be used. Takes
-#              string as input
+#  - style  :  Defines the kind of pastebin service to be used. ie it can
+#              provide the functionality of controlling privacy of 
+#              paste (in future).Currently only "fiche" servers
+#              are supported. Takes string as input
 #  - url    :  Defines the address of pastebin service to be used.
 #              Takes string as input
 #  - port   :  Defines the port number to be used to send logs. Takes 
 #              integer as input
 #  - enable :  Defines if the functionality is to be used or not. Takes 
 #              bool as input
-logUpload.style   :    "fische"
+logUpload.style   :    "fiche"
 logUpload.url     :    "termbin.com"
 logUpload.port    :    9999
 logUpload.enable  :    true
--- a/src/libcalamaresui/Branding.cpp
+++ b/src/libcalamaresui/Branding.cpp
@@ -86,6 +86,7 @@ const QStringList Branding::s_styleEntryStrings =
     "sidebarTextSelect",
     "sidebarTextHighlight"
 };
+
 // clang-format on
 // *INDENT-ON*
 
@@ -510,6 +511,11 @@ Branding::initSimpleSettings( const YAML::Node& doc )
     {
         m_windowHeight = WindowDimension( CalamaresUtils::windowPreferredHeight, WindowDimensionUnit::Pixies );
     }
+
+    m_logUploadEnable = doc[ "logUpload.enable" ].as< bool >( false );
+    m_logUploadURL = getString( doc, "logUpload.url") ;
+    m_logUploadPort = doc[ "logUpload.port" ].as< int >();
+    m_logUploadStyle= getString(doc, "logUpload.style" );
 }
 
 void
--- a/src/libcalamaresui/Branding.cpp
+++ b/src/libcalamaresui/Branding.cpp
@@ -87,6 +87,13 @@ const QStringList Branding::s_styleEntryStrings =
     "sidebarTextSelect",
     "sidebarTextHighlight"
 };
+
+const QStringList Branding::s_uploadServerStrings =
+{
+    "type",
+    "url",
+    "port"
+};
 // clang-format on
 // *INDENT-ON*
 
@@ -219,6 +226,12 @@ Branding::Branding( const QString& brandingFilePath, QObject* parent )
                 return imageFi.absoluteFilePath();
             } );
             loadStrings( m_style, doc, "style", []( const QString& s ) -> QString { return s; } );
+
+            const QVariantMap temp = CalamaresUtils::yamlMapToVariant( doc[ "uploadServer" ] );
+            for ( auto it = temp.constBegin(); it != temp.constEnd(); ++it )
+            {
+                m_uploadServer.insert( it.key(), it.value().toString() );
+            }
         }
         catch ( YAML::Exception& e )
         {
@@ -279,6 +292,11 @@ Branding::imagePath( Branding::ImageEntry imageEntry ) const
     return m_images.value( s_imageEntryStrings.value( imageEntry ) );
 }
 
+QString
+Branding::uploadServer( Branding::UploadServerEntry uploadServerEntry ) const
+{
+    return m_uploadServer.value( s_uploadServerStrings.value( uploadServerEntry ) );
+}
 
 QPixmap
 Branding::image( Branding::ImageEntry imageEntry, const QSize& size ) const
@@ -511,10 +529,6 @@ Branding::initSimpleSettings( const YAML::Node& doc )
     {
         m_windowHeight = WindowDimension( CalamaresUtils::windowPreferredHeight, WindowDimensionUnit::Pixies );
     }
-
-    m_uploadServerURL  = getString( doc, "uploadServer.url" );
-    m_uploadServerPort = doc[ "uploadServer.port" ].as< int >();
-    m_uploadServerType = getString( doc, "uploadServer.type" );
 }
 
 void
--- a/src/libcalamaresui/Branding.cpp
+++ b/src/libcalamaresui/Branding.cpp
@@ -87,7 +87,6 @@ const QStringList Branding::s_styleEntryStrings =
     "sidebarTextSelect",
     "sidebarTextHighlight"
 };
-
 // clang-format on
 // *INDENT-ON*
 
@@ -513,10 +512,9 @@ Branding::initSimpleSettings( const YAML::Node& doc )
         m_windowHeight = WindowDimension( CalamaresUtils::windowPreferredHeight, WindowDimensionUnit::Pixies );
     }
 
-    m_logUploadEnable = doc[ "logUpload.enable" ].as< bool >( false );
-    m_logUploadURL = getString( doc, "logUpload.url" );
-    m_logUploadPort = doc[ "logUpload.port" ].as< int >();
-    m_logUploadStyle = getString( doc, "logUpload.style" );
+    m_uploadServerURL  = getString( doc, "uploadServer.url" );
+    m_uploadServerPort = doc[ "uploadServer.port" ].as< int >();
+    m_uploadServerType = getString( doc, "uploadServer.type" );
 }
 
 void
--- a/src/libcalamaresui/Branding.cpp
+++ b/src/libcalamaresui/Branding.cpp
@@ -4,6 +4,7 @@
  *   SPDX-FileCopyrightText: 2017-2019 Adriaan de Groot <groot@kde.org>
  *   SPDX-FileCopyrightText: 2018 Raul Rodrigo Segura (raurodse)
  *   SPDX-FileCopyrightText: 2019 Camilo Higuita <milo.h@aol.com>
+ *   SPDX-FileCopyrightText: 2021 Anubhav Choudhary <ac.10edu@gmail.com>
  *   SPDX-License-Identifier: GPL-3.0-or-later
  *
  *   Calamares is Free Software: see the License-Identifier above.
@@ -513,9 +514,9 @@ Branding::initSimpleSettings( const YAML::Node& doc )
     }
 
     m_logUploadEnable = doc[ "logUpload.enable" ].as< bool >( false );
-    m_logUploadURL = getString( doc, "logUpload.url") ;
+    m_logUploadURL = getString( doc, "logUpload.url" );
     m_logUploadPort = doc[ "logUpload.port" ].as< int >();
-    m_logUploadStyle= getString(doc, "logUpload.style" );
+    m_logUploadStyle = getString( doc, "logUpload.style" );
 }
 
 void
--- a/src/libcalamaresui/Branding.h
+++ b/src/libcalamaresui/Branding.h
@@ -218,10 +218,9 @@ public:
 
 
     //Paste functionality related
-    bool logUploadEnable() { return m_logUploadEnable; };
-    QString logUploadURL() { return m_logUploadURL; };
-    int logUploadPort() { return m_logUploadPort; };
-    QString logUploadStyle() { return m_logUploadStyle; };
+    QString uploadServerType() { return m_uploadServerType; };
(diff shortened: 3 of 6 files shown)


src/branding/default/branding.desc
```diff
@@ -219,3 +219,18 @@ slideshow:               "show.qml"
 slideshowAPI: 2
 
 
+# These options are to customize online uploading of logs to pastebins:
+#  - style  :  Defines the kind of pastebin service to be used. ie it can
```
deprov447: style key is not hooked up to anywhere in backend.

src/branding/default/branding.desc
```diff
@@ -219,3 +219,18 @@ slideshow:               "show.qml"
...
+#              provide the functionality of controlling privacy of 
+#              paste (in future).Currently only "fiche" servers
+#              are supported. Takes string as input
+#  - url    :  Defines the address of pastebin service to be used.
+#              Takes string as input
+#  - port   :  Defines the port number to be used to send logs. Takes 
+#              integer as input
+#  - enable :  Defines if the functionality is to be used or not. Takes 
+#              bool as input
```
deprov447: Tried creating YAML list here but was not able to call them properly in `branding.cpp`. So instead declared independent keys. Can be changed to following format in future : 
`logUpload:
        - style
        - url
        - port
        - enable`

src/branding/default/branding.desc
bill-auger: that the 'enable' key is used but the 'style' key is not, indicates a smell - i would eliminate the 'enable' key - that semantics could be deduced by the value of 'style' (eg: 'none' or empty-string) - the only utility i can imagine for an 'enable' key, would be if there existed a mechanism to enable/disable the feature at runtime

i would also suggest renaming those logUpload* keys to: uploadServer*, and the 'style' key name to: 'type' instead - "style" is an aesthetics term

src/libcalamaresui/ViewManager.cpp
```diff
@@ -530,13 +532,9 @@ ViewManager::updateCancelEnabled( bool enabled )
...
 void
-ViewManager::updateBackAndNextVisibility( bool visible)
-{
-    UPDATE_BUTTON_PROPERTY( backAndNextVisible, visible )
-}
+ViewManager::updateBackAndNextVisibility( bool visible ) { UPDATE_BUTTON_PROPERTY( backAndNextVisible, visible ) }
 
-QVariant
-ViewManager::data( const QModelIndex& index, int role ) const
+QVariant ViewManager::data( const QModelIndex& index, int role ) const
```
bill-auger: these style changes to ViewManager::updateBackAndNextVisibility() and ViewManager::data() introduce inconsistent code style in this file

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -15,14 +15,17 @@
...
+#include <QApplication>
 
 namespace CalamaresUtils
 {
 
 QString
 sendLogToPastebin( QObject* parent, const QString& ficheHost, quint16 fichePort )
 {
-    QString pasteUrlFmt = parent->tr( "Install log posted to:\n%1" );
+    QString pasteUrlFmt = parent->tr( "Install log posted to\n\n%1\n\nLink copied to clipboard" );
```
bill-auger, line 39: this is a new or modified translation string - a new translation entry should be added, or the existing one modified - should that be done in the source tree, or is it an admin task via the translations service?

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -78,6 +81,14 @@ sendLogToPastebin( QObject* parent, const QString& ficheHost, quint16 fichePort
     QRegularExpression pasteUrlRegex( "^http[s]?://" + ficheHost );
     QString pasteUrlMsg = QString( pasteUrlFmt ).arg( pasteUrlStr );
 
+    QClipboard* clipboard = QApplication::clipboard();
```
bill-auger: this patch segment should be subject to the guarding clause below, which validates the data and URL - i suggest this change:

```
    if ( nBytesRead >= 8 && pasteUrl.isValid() && pasteUrlRegex.match( pasteUrlStr ).hasMatch() )
    {
        QClipboard* clipboard = QApplication::clipboard();
        clipboard->setText(pasteUrlStr, QClipboard::Clipboard);

        if (clipboard->supportsSelection())
        {
             clipboard->setText(pasteUrlStr, QClipboard::Selection);
        }
    }
    else
    {
        cError() << "No data from paste server";
        return QString();
    }
```

src/libcalamaresui/ViewManager.cpp
```diff
@@ -140,7 +141,7 @@ ViewManager::insertViewStep( int before, ViewStep* step )
 void
 ViewManager::onInstallationFailed( const QString& message, const QString& details )
 {
-    bool shouldOfferWebPaste = false;  // TODO: config var
+    bool shouldOfferWebPaste = Calamares::Branding::instance()->logUploadEnable();
```
bill-auger: the simplest thing to do is something like this:

```
QString uploadServerType = Calamares::Branding::instance()->uploadServerType();
bool shouldOfferLogUpload = CalamaresUtils::UploadServers.contains( uploadServerType );
```

where: UploadServers is const QStringList of uploadServerTypes (currently only [ 'fiche' ]); which does not contain the special dummy uploadServerType: 'none'

that eliminates the 'isEnabled' config boolean, the class member, and getter; but still the uploadServerType is unused - FWIW, this is actually suggesting a new CalamaresUtils::UploadServer class

```
CalamaresUtils::UploadServer* uploadServer = CalamaresUtils::UploadServer->instance();
bool shouldOfferLogUpload = uploadServer != nullptr;
```

that is effectively replacing the UploadServers list with a singleton getter, which returns an instance of an UploadServers subclass (eg: FicheUploadServer), instantiated lazily, and selected by the branding config, or nullptr if no UploadServers subclass corresponds to the branding selector

src/libcalamaresui/ViewManager.cpp
```diff
@@ -183,8 +184,9 @@ ViewManager::onInstallationFailed( const QString& message, const QString& detail
     connect( msgBox, &QMessageBox::buttonClicked, [msgBox]( QAbstractButton* button ) {
         if ( msgBox->buttonRole( button ) == QMessageBox::ButtonRole::YesRole )
         {
-            // TODO: host and port should be configurable
-            QString pasteUrlMsg = CalamaresUtils::sendLogToPastebin( msgBox, QStringLiteral( "termbin.com" ), 9999 );
+            QString pasteURLHost = Calamares::Branding::instance()->logUploadURL();
+            int pasteURLPort = Calamares::Branding::instance()->logUploadPort();
+            QString pasteUrlMsg = CalamaresUtils::sendLogToPastebin( msgBox, pasteURLHost, pasteURLPort );
```
bill-auger: naively, what is needed here is some sort of switch, eg:

```
QString pasteUrlMsg = (uploadServerType == 'fiche') ?
                      CalamaresUtils::sendLogToPastebin( msgBox, pasteURLHost, pasteURLPort ) :
                      QString();
```

passing in the hostname and port, to me is something of a smell - the current paste function is intrinsically specialized for a fiche server; so it could grab it's data directly from the branding class - besides, this logic really does not belong in a view class - (just an optional suggestion)

that would be sufficient to satisfy the feature request; but clearly not as DRY as it could be if, per my suggestion above, the paste function were an abstract sendLog() method of a subclass of an abstract UploadServer class:

```
QString pasteUrlMsg = uploadServer.sendLog( msgBox );
```
i would be leaning toward that approach; because it nicely avoids the need to modify the ViewManager class in the future

src/libcalamaresui/utils/Paste.cpp
deprov447: I don't have an idea regarding how translation is handled. But would be good if this piece is brought into focus of translators

src/libcalamaresui/ViewManager.cpp
deprov447: I am unable to comprehend what you said after `special dummy uploadServerType: 'none'`
It would be helpful if you can clarify a bit. Sorry for my weak english / tech language.

bill-auger:
i was only suggesting that a new class be added, to contain the
existing paste function - it would make the next patch very
simple; but consider that to be an optional suggestion, a
possible idea for the future - you can ignore it now

i gave two separate (analogous) suggestions, for calling the
function, in the latter part of this ViewManager file - like
these two suggestions, the first (naive) one is the simplest to
implement - the second one involves the idea of the new class -
you can ignore that one too

src/branding/default/branding.desc
```diff
@@ -219,3 +219,15 @@ slideshow:               "show.qml"
...
+# These options are to customize online uploading of logs to pastebins:
+#  - type   :  Defines the kind of pastebin service to be used.Currently
+#              it accepts two values:
+#              - none    :    disables the pastebin functionality
+#              - fiche   :    use fiche pastebin server
+#  - url    :  Defines the address of pastebin service to be used.
+#              Takes string as input
+#  - port   :  Defines the port number to be used to send logs. Takes 
+#              integer as input
+uploadServer.type    :    "fiche"
```
adriaandegroot: Typically you would structure this as
```
uploadServer:
    type: fiche
    url: https://termbin.com
    port: 9999
```
That also makes it slightly easier to write the schema for this. A downside is that the code that *loads* the settings is more complicated, since it first needs to get the *uploadServer*  object, and then get the subkeys from inside it.

src/libcalamaresui/Branding.cpp
```diff
@@ -510,6 +511,10 @@ Branding::initSimpleSettings( const YAML::Node& doc )
     {
         m_windowHeight = WindowDimension( CalamaresUtils::windowPreferredHeight, WindowDimensionUnit::Pixies );
     }
+
+    m_uploadServerURL  = getString( doc, "uploadServer.url" );
```
adriaandegroot: .. for the different approach to keys which I mentioned, you would use `getSubMap()` (um .. *is* this even a QVariantMap?) first and then `getString(mysubmap, "url")`.

After checking: no, this is all in a YAML doc. OK, then it's `YAML::Node submap = get(doc, "uploadServer");` followed by others.

src/libcalamaresui/Branding.cpp
```diff
@@ -510,6 +511,10 @@ Branding::initSimpleSettings( const YAML::Node& doc )
...
+    m_uploadServerPort = doc[ "uploadServer.port" ].as< int >();
```
adriaandegroot: This will throw if the value isn't an int, right? Does that get caught somewhere?

src/libcalamaresui/Branding.h
```diff
@@ -214,6 +216,12 @@ class UIDLLEXPORT Branding : public QObject
      */
     void setGlobals( GlobalStorage* globalStorage ) const;
 
+
+    //Paste functionality related
+    QString uploadServerType() { return m_uploadServerType; };
```
adriaandegroot: If this is an enumeration (and it is: none, or fiche, and we expect other paste-servers to be used, eventually) then we should have an enumeration for it and return that rather than a string. I'll describe that in a separate comment.

src/libcalamaresui/Branding.h
```diff
@@ -214,6 +216,12 @@ class UIDLLEXPORT Branding : public QObject
...
+    QString uploadServerURL() { return m_uploadServerURL; };
```
adriaandegroot: If it's a URL, this should return a QUrl

src/libcalamaresui/Branding.h
```diff
@@ -261,6 +269,10 @@ public slots:
     bool m_welcomeStyleCalamares;
     bool m_welcomeExpandingLogo;
 
+    QString m_uploadServerType;
+    QString m_uploadServerURL;
```
adriaandegroot: Same comments about types: the enumeration type should be an enum and the URL a QUrl.

src/libcalamaresui/ViewManager.cpp
```diff
@@ -140,7 +141,8 @@ ViewManager::insertViewStep( int before, ViewStep* step )
 void
 ViewManager::onInstallationFailed( const QString& message, const QString& details )
 {
-    bool shouldOfferWebPaste = false;  // TODO: config var
+    QString UploadServerType = Calamares::Branding::instance()->uploadServerType();
```
adriaandegroot: Unusual variable name; initial upper-case is used for types

src/libcalamaresui/ViewManager.cpp
```diff
@@ -140,7 +141,8 @@ ViewManager::insertViewStep( int before, ViewStep* step )
...
+    bool shouldOfferWebPaste = CalamaresUtils::UploadServersList.contains( UploadServerType );
```
adriaandegroot: This would become `uploadServerType != UploadServer::None` under the proposed use-an-enum change

src/libcalamaresui/ViewManager.cpp
```diff
@@ -183,8 +185,16 @@ ViewManager::onInstallationFailed( const QString& message, const QString& detail
     connect( msgBox, &QMessageBox::buttonClicked, [msgBox]( QAbstractButton* button ) {
         if ( msgBox->buttonRole( button ) == QMessageBox::ButtonRole::YesRole )
         {
-            // TODO: host and port should be configurable
-            QString pasteUrlMsg = CalamaresUtils::sendLogToPastebin( msgBox, QStringLiteral( "termbin.com" ), 9999 );
+            QString pasteUrlMsg;
+            QString UploadServerType = Calamares::Branding::instance()->uploadServerType();
```
adriaandegroot: Unusual variable name.

src/libcalamaresui/ViewManager.cpp
```diff
@@ -183,8 +185,16 @@ ViewManager::onInstallationFailed( const QString& message, const QString& detail
...
+            if ( UploadServerType == "fiche" )
```
adriaandegroot: This would be `uploadServerType == UploadServer::Fiche`

src/libcalamaresui/ViewManager.cpp
```diff
@@ -183,8 +185,16 @@ ViewManager::onInstallationFailed( const QString& message, const QString& detail
...
+            {
+                pasteUrlMsg = CalamaresUtils::sendLogToPastebin( msgBox );
```
adriaandegroot: We should rename this to something mentioning Fiche, though.

src/libcalamaresui/ViewManager.cpp
```diff
@@ -530,12 +540,12 @@ ViewManager::updateCancelEnabled( bool enabled )
 }
 
 void
-ViewManager::updateBackAndNextVisibility( bool visible)
+ViewManager::updateBackAndNextVisibility( bool visible )
```
adriaandegroot, line 543: Well-spotted! +1 for chasing the style guide.

src/libcalamaresui/ViewManager.cpp
```diff
@@ -530,12 +540,12 @@ ViewManager::updateCancelEnabled( bool enabled )
...
 {
     UPDATE_BUTTON_PROPERTY( backAndNextVisible, visible )
 }
 
-QVariant
+QVariant
```
adriaandegroot: .. and -1 for trailing spaces here. If you run `ci/calamaresstyle` on the files you edit, it will massage things into the "right" shape; that includes things like include-file-orderinging.

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -9,20 +9,35 @@
...
+#include "Branding.h"
 #include "utils/Logger.h"
 
 #include <QFile>
 #include <QRegularExpression>
 #include <QTcpSocket>
 #include <QUrl>
+#include <QClipboard>
+#include <QApplication>
+#include <QStringList>
```
adriaandegroot, line 21: Canonically, includes are ordered alphabetically within a block; the `calamaresstyle` script will do the right thing (using astyle and clang-format).

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -9,20 +9,35 @@
...
 
 namespace CalamaresUtils
 {
 
+QStringList UploadServersList = {
+    "fiche"
+    // In future more serverTypes can be added as Calamares support them
+    // "none" serverType is explicitly not mentioned here
+};
```
adriaandegroot, line 30: Make it an enum. I'll put some documentation in other comments.

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -9,20 +9,35 @@
...
+
 QString
-sendLogToPastebin( QObject* parent, const QString& ficheHost, quint16 fichePort )
+sendLogToPastebin( QObject* parent )
 {
-    QString pasteUrlFmt = parent->tr( "Install log posted to:\n%1" );
+
+    const QString& ficheHost = Calamares::Branding::instance()->uploadServerURL();
+    quint16 fichePort = Calamares::Branding::instance()->uploadServerPort();
```
adriaandegroot: There's some int/uint mismatch here, but come to think of it if we use a QUrl, we could put the port number right there in the URL instead (eg. http://termbin.com:9999).

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -9,20 +9,35 @@
...
+
+    QString pasteUrlFmt = parent->tr( "Install log posted to\n\n%1\n\nLink copied to clipboard" );
```
adriaandegroot, line 39: You can use rich-text formatting here, so it might be good to have something like <br/> and maybe even a clickable link. You would need to check if the layout / formatting comes out right, though.

adriaandegroot:
Regarding enums / the *type* configuration, there are lots of configuration items in Calamares where the configuration file contains a string that must be one string out of an enumeration (e.g. in the branding file, *windowExpanding*). There is a mechanism for "conveniently" building up an enum, and back-and-forth translations between enum and strings; this is used in the config-parsing code which gets a string, converts it to an enum, and then the code itself uses the enum value.

"Named Enums" can be found in `libcalamares/utils/NamedEnum.h` which doesn't have enough documentation to be really useful :disappointed: but you can see an example in `Branding.cpp` in the function `flavorAndSide()`. A `NamedEnumTable<E>` is the mapping between enum values and strings; it supports aliases when mapping **to** enums, and maps to a canonical string value when going back. The table really is just a list of strings and values. Many classes have a `static const NamedEnumTable<E>& namesOfE()` function that returns a reference to the table (not, though, the branding example I mentioned).

While it does not necessarily reduce the amount of overall code, it replaces string comparisons with strongly-typed enum comparisons, and that improves code quality overall. There is one source of truth about spelling, for instance.

I'll try to expand the NamedEnum documentation so it becomes easier to use (and if it also fits in the developers guide, fit it in there).

bill-auger:
On Tue, 09 Feb 2021 05:23:02 -0800 Adriaan wrote:
> +    m_uploadServerURL  = getString( doc, "uploadServer.url" );
> +    m_uploadServerPort = doc[ "uploadServer.port" ].as< int >();
> 
> This will throw if the value isn't an int, right? Does that get caught somewhere?

the docs for that method are not great; but (at a shallow
look) it does not appear that it would throw for that reason

there are two overloads, one which takes no args, and will
throw if the node is not 'm_is_valid' - the other takes a
fallback to avoid the exception

i did not follow-through to discover what exactly qualifies as:
'm_is_valid'; but i assume that is merely the presence of the
node itself (with any value, perhaps even the empty string) -
ie: an existing node with non-numeric input probably will return
'0' (or the fallback param)

https://yaml-cpp.docsforge.com/master/api/YAML/Node/as/

src/libcalamaresui/Branding.cpp
```diff
@@ -218,6 +226,12 @@ Branding::Branding( const QString& brandingFilePath, QObject* parent )
                 return imageFi.absoluteFilePath();
             } );
             loadStrings( m_style, doc, "style", []( const QString& s ) -> QString { return s; } );
+
+            const QVariantMap temp = CalamaresUtils::yamlMapToVariant( doc[ "uploadServer" ] );
+            for ( auto it = temp.constBegin(); it != temp.constEnd(); ++it )
+            {
+                m_uploadServer.insert( it.key(), it.value().toString() );
+            }
```
deprov447, line 234: Have to use loadStrings() function

PR #1620, opened 2021-01-25
ci: move from travisci to github actions

boredland:
- moved from scripts to step definitions in github actions
- moved from the `kdeneon/all:user`-base to `kdeneon/plasma:user` (half the size and lots of newer software)

Overall the speed increased a lot. Current builds on Travis seem to take [14.5m](https://travis-ci.org/github/calamares/calamares
). In GH actions they take 8.5m. My proposal would be, that we always run with coverage. Can't test though, since I dont have the `COVERITY_SCAN_TOKEN` secret. And I'm not sure coverity currently even works, since it ran ["a while ago"](https://scan.coverity.com/projects/5389). If it isn't being used anymore, we could drop that code.

missing: 
- IRC (no idea what kind of notifications you have there, so hard to replicate)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,118 @@
...
+      options: --tmpfs /build:rw,size=512M --user 0:0
+    steps:
+      - 
+        name: prepare
+        env:
+          COVERITY_SCAN_TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install git-core
+          if [ -n "$COVERITY_SCAN_TOKEN" ]; then
```
boredland: only made conditional bc I don't have that secret.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,118 @@
...
+      -
+        name: cmake
+        run: cmake $CMAKE_ARGS $SRCDIR
+      -
+        name: generate coverage report
+        env:
+          COVERITY_SCAN_TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
+        run: |
+          if [ -n "$COVERITY_SCAN_TOKEN" ]; then
+            $BUILDDIR/coveritytool/bin/cov-build --dir cov-int make -j2
```
boredland: if this does what I think this does, than it could make sense to run this instead of the regular "make" https://github.com/calamares/calamares/pull/1620/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR107

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+    -DKDE_INSTALL_USE_QT_SYS_PATHS=ON
+    -DWITH_PYTHONQT=OFF"
+    -DCMAKE_BUILD_TYPE=Debug
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: docker://kdeneon/plasma:user
+      options: --tmpfs /build:rw,size=512M --user 0:0
```
boredland: the tmpfs size was `112M` before. for some reason this wasn't enough anymore

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+env:
+  BUILDDIR: /build
+  SRCDIR: ${{ github.workspace }}
+  DESTDIR: /INSTALL_ROOT
+  CMAKE_ARGS: |
+    -DCMAKE_BUILD_TYPE=Release
+    -DWEBVIEW_FORCE_WEBKIT=1
+    -DKDE_INSTALL_USE_QT_SYS_PATHS=ON
+    -DWITH_PYTHONQT=OFF"
+    -DCMAKE_BUILD_TYPE=Debug
```
adriaandegroot, line 21: It seems reasonable to specify the build type only once (let's pick *Debug*)

.github/workflows/ci.yml
adriaandegroot: If you switched to debug builds, then the build size is going to be much bigger, yes. I tried to keep the tmpfs size down -- which caused CI failures in the past when some change somewhere caused the build size to grow suddenly, like new languages -- partly out of respect for Travis' resources. I don't think that's all that important anymore.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+            libqt5svg5-dev \
+            libqt5webkit5-dev \
+            libyaml-cpp-dev \
+            os-prober \
+            pkg-config \
+            python3-dev \
+            qtbase5-dev \
+            qtdeclarative5-dev \
+            qttools5-dev \
+            qttools5-dev-tools
```
adriaandegroot, line 68: I'll double-check this list of packages against *deploy-cala* (for consistency sake -- nothing for you to do)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+      -
+        name: check environment
+        run: |
+          test -f $SRCDIR/CMakeLists.txt || { echo "! Missing $SRCDIR/CMakeLists.txt" ; exit 1 ; }
+          test -n "$BUILDDIR" || { echo "! \$BUILDDIR not set" ; exit 1 ; }
+          test -n "$SRCDIR" || { echo "! \$SRCDIR not set" ; exit 1 ; }
+          mkdir -p $BUILDDIR $SRCDIR $DESTDIR
```
adriaandegroot: Don't need to mkdir `$SRCDIR` since it was created on checkout and tested in L74 already.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+            pkg-config \
+            python3-dev \
+            qtbase5-dev \
+            qtdeclarative5-dev \
+            qttools5-dev \
+            qttools5-dev-tools
+      -
+        name: check environment
+        run: |
+          test -f $SRCDIR/CMakeLists.txt || { echo "! Missing $SRCDIR/CMakeLists.txt" ; exit 1 ; }
```
adriaandegroot: This should got **after** the `test -n` for `$SRCDIR` (or if it's acceptable for `$SRCDIR` to be empty and sources checked out in `/`, massage the tests that way)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+        name: build results
+        run: ls -la $( find "$BUILDDIR" -type f -name '*.so' )
+      - 
+        name: install
+        id: install
+        run:  make install VERBOSE=1
+      -
+        name: debug install
+        if: ${{ failure() && steps.install.outcome == 'failure' }}
+        run: ls -la $( find "$DESTDIR" -type f -name '*.so' )
```
adriaandegroot: Since `$DESTDIR` isn't used anywhere else, I wonder if this is going to provide any useful information. The whole find-on-failure is a bit of a relic of running out of disk space on Travis (see earlier comment on the size of tmpfs) rather than anything specific.

adriaandegroot (commented):
Overall makes sense, and thank you for understanding the GH actions syntax so that I don't have to.

.github/workflows/ci.yml
boredland: copied that from the Dockerfile that was in here before. If sth. isn't needed, that'd be great I guess.

.github/workflows/ci.yml
boredland: sorry for the copy&pasta :ramen:

.github/workflows/ci.yml
boredland: guess it doesn't

PR #1622, opened 2021-01-26
Create and mount btrfs subvolumes in generalized manner

Chrysostomus:
The aim is to have distro configurable btrfs subvolumes instead of hardcoded @ and @home.

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -112,7 +112,7 @@ def mount_partition(root_mount_point, partition, partitions):
 
         # Mount the subvolumes
         for s in btrfs_subvolumes:
-            mount_option = f"subvol={s[subvolume]}"
+            mount_option = f"subvol={s['subvolume']}"
             if libcalamares.utils.mount(device,
                                     mount_point,
                                     fstype,
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -104,13 +104,13 @@ def mount_partition(root_mount_point, partition, partitions):
 
         # Mount the subvolumes
         for s in btrfs_subvolumes:
-            mount_option = f"subvol={s['subvolume']}"
+            mount_option = "subvol={}".format(s['subvolume'])
             subvolume_mountpoint = mount_point[:-1] + s['mountPoint']
             if libcalamares.utils.mount(device,
                                     subvolume_mountpoint,
                                     fstype,
                                     ",".join([mount_option, partition.get("options", "")])) != 0:
-                libcalamares.utils.warning(f"Cannot mount {device}")
+                libcalamares.utils.warning("Cannot mount {}".format(device))
 
 
 def run():
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -89,7 +89,7 @@ def mount_partition(root_mount_point, partition, partitions):
             if "mountPoint" not in p or not p["mountPoint"]:
                 continue
             if p["mountPoint"] in subvolume_mountpoints and p["mountPoint"] != '/':
-                btrfs_subvolumes = [d for d in btrfs_subvolumes if d['mountPoint'] != p["mountpoint"]]
+                btrfs_subvolumes = [d for d in btrfs_subvolumes if d['mountPoint'] != p["mountPoint"]]
         # Check if we need a subvolume for swap file
         swap_choice = global_storage.value( "partitionChoices" )
         if swap_choice:
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -89,10 +89,8 @@ def mount_partition(root_mount_point, partition, partitions):
             if "mountPoint" not in p or not p["mountPoint"]:
                 continue
             if p["mountpoint"] in subvolume_mountpoints and p["mountpoint"] != '/':
-                # mountpoint is already defined, remove subvolume from the list
                 btrfs_subvolumes = [d for d in btrfs_subvolumes if d['mountPoint'] != p["mountpoint"]]
         # Check if we need a subvolume for swap file
-        needs_swap_subvolume = False
         swap_choice = global_storage.value( "partitionChoices" )
         if swap_choice:
             swap_choice = swap_choice.get( "swap", None )
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -81,26 +81,29 @@ def mount_partition(root_mount_point, partition, partitions):
     # If a swapfile is used, we also create a subvolume "@swap".
     # Finally we remount all of the above on the correct paths.
     if fstype == "btrfs" and partition["mountPoint"] == '/':
-        has_home_mount_point = False
+        # Root has been mounted to btrfs volume -> create subvolumes from configuration
+        btrfs_subvolumes = libcalamares.job.configuration.get("btrfsSubvolumes") or []
+        subvolume_mountpoints = [d['mountPoint'] for d in btrfs_subvolumes]
+        # Check if listed mountpoints besides / are already present and don't create subvolume for those
         for p in partitions:
             if "mountPoint" not in p or not p["mountPoint"]:
                 continue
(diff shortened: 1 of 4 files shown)


Chrysostomus:
To do:

- Write fstab correctly in generalized manner. I don't have that much experience in writing files in python so tips are welcome
- Test if this actually works (once all other parts are in place)
- Clean the code

Chrysostomus:
@abucodonosor @adriaandegroot 
The pull request has progressed to the point where it would benefit from review. The code is not yet tested and I'm not sure if I'm using global storage correctly, but at least the main logic should now be clear, so you can tell if this approach is desireable and if I should continue working on this, or trying something completely different.

Chrysostomus:
Got to the testing phase. Fixed a few general bugs. It seems that the swapfile handling is broken at the moment at least.

src/modules/mount/main.py
```diff
@@ -81,29 +81,27 @@ def mount_partition(root_mount_point, partition, partitions):
     # If a swapfile is used, we also create a subvolume "@swap".
     # Finally we remount all of the above on the correct paths.
     if fstype == "btrfs" and partition["mountPoint"] == '/':
-        has_home_mount_point = False
+        # Root has been mounted to btrfs volume -> create subvolumes from configuration
+        btrfs_subvolumes = libcalamares.job.configuration.get("btrfsSubvolumes") or []
```
Chrysostomus, line 77: This returns only [] for some reason.

src/modules/mount/mount.conf
```diff
@@ -19,6 +19,10 @@
 # The device is not mounted if the mountPoint is unset or if the fs is
 # set to unformatted.
 #
+---
+# Btrfs subvolumes to create if root filesystem is on btrfs volume.
+# If mountpoint is mounted already to another partition, it is ignored.
+# Separate subvolume for swapfile is handled separately and automatically.
```
adriaandegroot: This comment should probably be with the configuration key it describes, not here.

src/modules/mount/mount.schema.yaml
```diff
@@ -29,3 +29,12 @@ properties:
...
             required: [ device, mountPoint ]
+    btrfsSubvolumes:
+        type: array
+        items:
+            type: object
+            additionalProperties: false
+            properties:
+                mountPoint: { type: string }
+                subvolume: { type: string }
+            required: [ subvolume, mountPoint ]
```
adriaandegroot, line 40: Thanks for immediately updating the schema. This looks fine.

src/modules/mount/main.py
```diff
@@ -112,25 +110,14 @@ def mount_partition(root_mount_point, partition, partitions):
...
-                libcalamares.utils.warning("Cannot mount {}".format(device))
-        
-        if needs_swap_subvolume:
-            if libcalamares.utils.mount(device,
-                                        root_mount_point + "/swap",
-                                        fstype,
-                                        ",".join(["subvol=@swap", partition.get("options", "")])) != 0:
-                libcalamares.utils.warning("Cannot mount {}".format(device))
+                                    ",".join([mount_option, partition.get("options", "")])) != 0:
+                libcalamares.utils.warning(f"Cannot mount {device}")
```
adriaandegroot: Using f-strings makes this a Calamares 3.3 PR, because f-strings are a newer Python feature than the Python supported by Calamares.

src/modules/mount/main.py
```diff
@@ -81,29 +81,27 @@ def mount_partition(root_mount_point, partition, partitions):
...
             if swap_choice and swap_choice == "file":
-                needs_swap_subvolume = True
-
-        subprocess.check_call(['btrfs', 'subvolume', 'create',
-                               root_mount_point + '/@'])
-
-        if not has_home_mount_point:
+                btrfs_subvolumes.append({'mountPoint': '/swap', 'subvolume': '/@swap'})
+        # Store created list in global storage so it can be used in the fstab module
+        libcalamares.globalstorage.insert("btrfsSubvolumes", btrfs_subvolumes)
```
adriaandegroot, line 92: Maybe log here what is being stored, so you're sure from the log that it was set?

src/modules/mount/main.py
```diff
@@ -81,29 +81,27 @@ def mount_partition(root_mount_point, partition, partitions):
...
+        # Check if listed mountpoints besides / are already present and don't create subvolume for those
         for p in partitions:
             if "mountPoint" not in p or not p["mountPoint"]:
                 continue
-            if p["mountPoint"] == "/home":
-                has_home_mount_point = True
-                break
-        needs_swap_subvolume = False
+            if p["mountPoint"] in subvolume_mountpoints and p["mountPoint"] != '/':
+                btrfs_subvolumes = [d for d in btrfs_subvolumes if d['mountPoint'] != p["mountpoint"]]
```
adriaandegroot: Looks like a typo to me: one line above, you use (case sensitive!) "mountPoint" for both, here the second one is all-lower-case.

adriaandegroot (changes requested):

src/modules/mount/main.py
Chrysostomus: Good idea,I was testing it with added print statements in local branch.

src/modules/mount/main.py
Chrysostomus: Good to know. When is calamares 3.3 release planned? I can adjust it to another style if 3.3 is far away, but it kinda feels silly to change it to less efficient and older form.

src/modules/mount/mount.conf
Chrysostomus: Done

src/modules/mount/mount.schema.yaml
Chrysostomus: No problem, I was not sure if it was needed so I added it just in case.

src/modules/mount/main.py
Chrysostomus: There was a typo in mount.conf and keyerror in this file, fixed now and subvolumes are correctly generated. Not mounted though, besides /@.

src/modules/mount/main.py
Chrysostomus: Thanks, fixed.

Chrysostomus:
Okay, tested installation, it works now. I'll replace the f strings to make it compatible with older python versions.

src/modules/mount/main.py
Chrysostomus: F-strings removed and replaced with format() method

Chrysostomus:
Should be okay to merge now, I'll run some more tests to make sure f-string removal broke nothing.

Chrysostomus:
Tested now, works with automatic partitioning. I'll test if ignoring subvolumes with automatic partitioning works as intended.

Chrysostomus:
Okay, works on manual paritioning too. mountpoints mounted to other volumes are ignored and no subvolumes are created for them, just as intended. This is now ready to be merged. I got some grub-install errors, but those don't seem to be related to this PR.

More specifically:

```
Traceback:
File "/usr/lib/calamares/modules/bootloader/main.py", line 476, in run
    prepare_bootloader(fw_type)

  File "/usr/lib/calamares/modules/bootloader/main.py", line 448, in prepare_bootloader
    install_grub(efi_directory, fw_type)

  File "/usr/lib/calamares/modules/bootloader/main.py", line 353, in install_grub
    check_target_env_call([libcalamares.job.configuration["grubInstall"],

  File "<string>", line 6, in <module>
```

Chrysostomus:
Confirmed, the same bug is present on unpatched calamares, so no relation to this pull request.

Conan-Kudo:
@adriaandegroot Any chance this could get looked at soon? I'm interested in using this in OpenMandriva and openSUSE.

Conan-Kudo (approved):
The only issue I can see with it is that there's no obvious way to handle multiple btrfs partitions with their own subvolume layouts, but that case should be fairly rare and it should be possible to extend this to handle that properly in the future.

PR #1623, opened 2021-01-27
Create shellprocess.schema.yaml

### LINKED ISSUE #1530: Add schema file for shellprocess
The *shellprocess* module has a `shellprocess.conf` with plenty of documentation, but there is no `shellprocess.schema.yaml` that describes the same file in a machine-readable way.

The schema files from other modules can serve as an example / inspiration. Once there's a schema file, the configuration file is checked against the schema to ensure they stay in-step during development, and downstreams can use the schema to check their own configuration files.

benne-dee:
Handwrote schema in [json-schema] then auto-converted to yaml to match the 
schema format in other modules. Resolves #1530

Thanks for the lesson in json schema

Have verified the schema validates the config and examples in the .conf file.
Don't have many test cases though, so please confirm it's working as expected.

[json-schema]: https://json-schema.org/understanding-json-schema/

### DIFF
--- a/src/modules/shellprocess/shellprocess.schema.yaml
+++ b/src/modules/shellprocess/shellprocess.schema.yaml
@@ -1,0 +1,54 @@
+$schema: http://json-schema.org/draft-07/schema#
+$id: https://calamares.io/schemas/shellprocess
+definitions:
+  command:
+    $id: '#definitions/command'
+    type: string
+    description: This is one command that is executed.  If a command starts with '-'
+      (a single minus sign), then the return value of the command following the -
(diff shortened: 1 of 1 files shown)


adriaandegroot:
Oh, well done! I like that you have used `$ref` to isolate some of the definition parts; that makes them easier to extract to other modules, too (e.g. the *contextualprocess* module has the same definition for *script*)

benne-dee:
> e.g. the contextualprocess module has the same definition for script

Cool! And thanks :)
I have more for you. Did some writing (schema) for _group_ in _netinstall_ module. 
Will bring it to you in a bit.

Issue #1543, opened 2020-10-23
Schema file for netinstall


adriaandegroot:
The *netinstall* module has a fairly complex configuration: the *groups* key contains the full netinstall YAML with groups, subgroups, and package lists. The documentation is extensive and there are a few examples, but the schema file does not specify the groups at all (beyond "the key exists and the value is an array").

## Procedural

 - [ ] Make sure you can compile Calamares locally (see https://github.com/calamares/calamares/blob/calamares/CONTRIBUTING.md, which will point you onwards to the wiki)
 - [ ] Make sure the *netinstall* module is enabled in your Calamares build (by default it is compiled, and it might be in `settings.conf`: check the CMake output to see that it will be compiled, and then double-check that *netinstall* is listed in the show section of the sequence in `settings.conf`)
 - [ ] Module sources are in https://github.com/calamares/calamares/tree/calamares/src/modules/netinstall ; make a branch, add your copyright statement to the existing schema file and add your name to top-level CHANGES, then start a PR with WIP in the title

## Functional

 - [ ] Brush up on the JSON schema documentation, e.g. https://json-schema.org/learn/getting-started-step-by-step.html
 - [ ] Expand the schema of *groups* to at **least** name the fields of a group (e.g. *name*, *description*, *hidden*, ...) It's OK at this point to skip *subgroups* (which is a recursive type).
 - [ ] Make sure *packages* is also part of the schema.
 - [ ] Write some "simple" `netinstall.conf` files and add them as test files and check that they validate. These don't need to have subgroups at all.
 - [ ] Look into splitting the schema out into a separate schema-file, since the `netinstall.yaml` files follow the same schema as the content of the *groups* key in the `netinstall.conf` file.
 - [ ] Extend the schema to handle the recursive subgroups entries.

## Post-Procedural

Feel free to push to the PR branch regularly; this shows how your progress is and allows early comments. Please stick to the commit-message style used in the rest of Calamares; do write out what-and-why with commits. It's OK to leave fixme's around for later in the branch and we can coordinate force-pushes if need be.

adriaandegroot:
This was resolved by @benne-dee when the schema was merged; it looks complete-enough for me.

PR #1624, opened 2021-01-27
[netinstall] Schema for groups in netinstall.schema.yaml

benne-dee:
Tackles #1543. 
Ticks most of the functional boxes. (~Looking into splitting)  
I am afraid not the procedural ones though. (That could take a while)  
And am not really experienced with the etiquette for copyright-CHANGES.

### DIFF
--- a/src/modules/netinstall/netinstall.schema.yaml
+++ b/src/modules/netinstall/netinstall.schema.yaml
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2020 Adriaan de Groot <groot@kde.org>
 # SPDX-License-Identifier: GPL-3.0-or-later
 ---
-$schema: http://json-schema.org/draft-07/schema#
+$schema: https://json-schema.org/draft-07/schema#
 $id: https://calamares.io/schemas/netinstall
 definitions:
     package:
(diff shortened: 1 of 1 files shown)


benne-dee:
Did a debug build and ran the _netinstalltest_ binary. Current tests seem to all pass.   
It might be beneficial though to now add a test for validating of `netinstall.yaml` against schema.  

About splitting, I am not sure if it might still be required. There's a nice clean groups schema definition in the schema. 
Anyone needing to validate the groups file can (hopefully) easily reference it from there.  
I can **try** to add a separate schema file that does just that (reference groups from the main _netinstall_ schema).

Although note that I had some trouble trying to reference definitions across files. It seems the _calamares.io_ url in _$id_ are not actually responding with the schemas.

benne-dee:
This schema now (after latest commit) validates dedicated groups file as well as the regular .conf file
Separate validator for groups files does not seem necessary. But if required a redundant schema as follows should do the job. 

```
$schema: https://json-schema.org/draft-07/schema#
$id: https://calamares.io/schemas/netinstall-groups

oneOf:
-   # Groups file with top level "groups" key
    type: object
    description: Groups file with top level *groups* key
    additionalProperties: false
    properties:
        groups: { $ref: './netinstall#definitions/groups' }
    required: [ groups ]
-   # Groups file bare
    { $ref: './netinstall#definitions/groups' }

```

PR #1628, opened 2021-01-30
modules/bootloader: Use the correct names for the shim binaries

Conan-Kudo:
Ever since signed shim binaries for multiple architectures became available, the shim binaries installed in Linux distributions have been renamed to include the EFI architecture in the binary names.

This started in Fedora, but is now used in openSUSE and Ubuntu too.

Reference for shim binary names comes from shim spec in Fedora: https://src.fedoraproject.org/rpms/shim/blob/d8c3c8e39235507feb17b4e81851da087ae83c77/f/shim.spec#_23-32

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -38,6 +38,7 @@ and moral support from (alphabetically by first name or nickname):
  - Kevin Kofler
  - Kyle Robertze
  - Lisa Vitolo
+ - Neal Gompa
  - n3rdopolis
  - Philip Müller
  - Ramon Buldó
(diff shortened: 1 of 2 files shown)


kkofler:
Thanks for the fix! But I am a bit sceptical about making this change unconditionally, because it will break support for older distributions, or distributions less up to date with upstream. Typically, we try to support those too in Calamares. @adriaandegroot: What do you think? Can we merge this as is or do we need to handle both the old and new names?

Conan-Kudo:
If it helps, it seems the shim binary name change also was backported all the way back to Ubuntu 16.04 and RHEL/CentOS 7. So the old code wasn't going to work there anyway. ☹️

kkofler:
Considering the above (that even the LTS distros got the renames backported), I think we can merge this pull request as is, but I'm waiting for confirmation from @adriaandegroot.

Issue #1456, opened 2020-07-14
Need an option to skip the Bootloader installation - essential for expert users


informer2016:
If you're trying to do a full disk encryption using a method like [**Artix Linux** _(great Arch with OpenRC or runit/s6 instead of SystemD)_ - **Full Disk Encryption for a new install**](https://forum.artixlinux.org/index.php/topic,1541.0.html) - creating/mounting the partitions encrypted with the custom flags and then pointing a Calamares to them - there's a high chance of running into this installation error:
```
Installation Failed

Boost.Python error in job "bootloader".

Command 'grub-install --target=i386-pc --recheck --force /dev/sda' returned non-zero exit status 1.
Installing for i386-pc platform.
File descriptor 7 (/root/.cache/calamares/session.log) leaked on vgs invocation.
Parent PID 7025: grub-install File descriptor 17 (/dev/mapper/control) leaked on vgs invocation.
Parent PID 7025: grub-install WARNING: Failed to connect to lvmetad. Falling back to device scanning.
File descriptor 7 (/root/.cache/calamares/session.log) leaked on vgs invocation.
Parent PID 7025: grub-install File descriptor 17 (/dev/mapper/control) leaked on vgs invocation.
Parent PID 7025: grub-install WARNING: Failed to connect to lvmetad. Falling back to device scanning.
grub-install: error: attempt to install to encrypted disk without cryptodisk enabled. Set `GRUB_ENABLE_CRYPTODISK=y' in file `/etc/default/grub'.

Traceback:

File "/usr/lib/calamares/modules/bootloader/main.py", line 465, in run
    prepare_bootloader(fw_type)

  File "/usr/lib/calamares/modules/bootloader/main.py", line 437, in prepare_bootloader
    install_grub(efi_directory, fw_type)

  File "/usr/lib/calamares/modules/bootloader/main.py", line 344, in install_grub
    check_target_env_call([libcalamares.job.configuration["grubInstall"],

  File "<string>", line 11, in <module>
```
There should be an option to **not to do** a GRUB installation, just in case a user needs to do it later _(after completing a Calamares install)_ in a custom way, or doesn't want to overwrite the existing bootloader which may be not a GRUB. Currently, the only way to skip a GRUB installation is to manually edit a `/usr/lib/calamares/modules/bootloader/main.py` 
and comment out this line near the end of file at "def run()" function:
```
prepare_bootloader(fw_type)
```
--->
```
# prepare_bootloader(fw_type)
```
[...]

PR #1632, opened 2021-02-02
[WIP] Skipping bootloader fix

deprov447:
Addressing issue #1456

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -465,6 +465,10 @@ def run():
         libcalamares.utils.warning( "Non-EFI system, and no bootloader is set." )
         return None
 
+    if (libcalamares.globalstorage.value("bootLoader") is None and fw_type == "efi"):
+     ┊   libcalamares.utils.warning( "EFI system detected, and no bootloader is set." )
+     ┊   return None
+
     partitions = libcalamares.globalstorage.value("partitions")
     if fw_type == "efi":
         efi_system_partition = libcalamares.globalstorage.value("efiSystemPartition")
--- a/src/modules/partition/core/BootLoaderModel.cpp
+++ b/src/modules/partition/core/BootLoaderModel.cpp
@@ -3,6 +3,7 @@
  *   SPDX-FileCopyrightText: 2014 Aurélien Gâteau <agateau@kde.org>
  *   SPDX-FileCopyrightText: 2015 Teo Mrnjavac <teo@kde.org>
  *   SPDX-FileCopyrightText: 2019 Adriaan de Groot <groot@kde.org>
+ *   SPDX_FileCopyrightText: 2021 Anubhav Choudhary <ac.10edu@gmail.com>
  *   SPDX-License-Identifier: GPL-3.0-or-later
  *
  *   Calamares is Free Software: see the License-Identifier above.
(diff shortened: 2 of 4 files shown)


src/modules/partition/gui/ChoicePage.cpp
```diff
@@ -983,6 +984,51 @@ ChoicePage::updateActionChoicePreview( InstallChoice choice )
         m_previewAfterFrame->show();
         m_previewAfterLabel->show();
 
+
+        {
```
deprov447: These braces can be removed at last. (there for visibility)

src/modules/bootloader/main.py
```diff
@@ -465,6 +465,10 @@ def run():
         libcalamares.utils.warning( "Non-EFI system, and no bootloader is set." )
         return None
 
+    if (libcalamares.globalstorage.value("bootLoader") is None and fw_type == "efi"):
+     ┊   libcalamares.utils.warning( "EFI system detected, and no bootloader is set." )
+     ┊   return None
```
adriaandegroot, line 470: The indentation looks funny here; tabs vs spaces maybe?

src/modules/partition/core/BootLoaderModel.cpp
```diff
@@ -124,9 +125,9 @@ BootLoaderModel::updateInternal()
             appendRow( createBootLoaderItem( partitionText, PartitionInfo::mountPoint( partition ), true ) );
         }
 
-        // Create "don't install bootloader" item
-        appendRow( createBootLoaderItem( tr( "Do not install a boot loader" ), QString(), false ) );
     }
+	//Create "don't install bootloader" item
+	appendRow( createBootLoaderItem( tr( "Do not install a boot loader" ), QString(), false ) );
```
adriaandegroot, line 130: And indentation looks funny here, too (it's not a big deal, the formatting tool will take care of it).

src/modules/partition/gui/ChoicePage.cpp
adriaandegroot: So this adds (duplicates) a whole chunk of code from the Erase/Replace case into Alongside as well. At this point I'd insist on refactoring it so that there's a function that returns a widget with bootloader-selection combo. That can keep the lambda's connected to the signals, and we can paste the widget into the layout for Erase or Alongside as needed.

adriaandegroot:
It's WIP, so I'll comment on general bits here.

One of the problems we have in the innards of the partition module is that things that looks like models aren't, or change in ways that models should not really change: for instance, replacing the entire data structure with a pointer to a **new** model. You can see that in `createBootloaderComboBox()` a little, and also in the lambda that is connected to `deviceReverted()` which needs to check object identities.

Part of an effort to improve this situation would expose a consistent model **and** a selection model; possibly a filtered model as well to show relevant bootloaders for the chosen device. That might lead to some serious Model/View wrangling, though -- I haven't looked at the internals recently.

src/modules/bootloader/main.py
deprov447: Will run the style script once PR is ready for merge 😅.

adriaandegroot:
This has effectively landed in *work-3.3* now.

PR #1645, opened 2021-02-21
kernel-install changes to bootloader module

dalto8:
We made some changes for EndeavourOS to support systemd-boot/kernel-install and we thought we should make them available here.

These changes have the following effects/benefits:
* Produce entries and use install locations compatible with kernel-install as provided by systemd
* Support for multiple kernels to be installed/available
* Support for additional initrd images such ucode files required by Arch

There are some downsides as well
* Some of the customization options were removed to produce something compatible with kernel-install
* There are two pieces of this that are Arch specific that I didn't see an obvious way to generalize
  * The generation of the initramfs requires the kernel version to be passed to it and calls mkinitcpio directly
  * The way that we are getting the kernel version is also specific to the file layout in Arch

I am not really sure this is better or worse than what was there before.  It is just a different approach.


killajoe:
Tested already on EndeavourOS:
https://github.com/endeavouros-team/ISO/releases/download/0-EndeavourOS-development-ISO-releases/endeavouros-devel-SDB-2021.02.26-x86_64.iso
Includes this changes.

src/modules/bootloader/bootloader.schema.yaml
```diff
@@ -7,10 +7,10 @@ additionalProperties: false
 type: object
 properties:
     efiBootLoader: { type: string }
-    kernel:  { type: string }
-    img:  { type: string }
-    fallback:  { type: string }
+    kernelSearchPath:  { type: string }
+    kernelName:  { type: string }
```
adriaandegroot, line 11: Because this changes the schema -- so distro's need to go over their configs extra-carefully -- I'm tagging this as a 3.3 pull-request. For a distro like EndeavourOS that knows what it is doing and **chooses** to pull this in, updating the config is natural, but I don't want to surprise other consumers with it.

src/modules/bootloader/main.py
```diff
@@ -101,9 +82,10 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
     :param uuid:
     :param entry:
     :param entry_name:
+    :param kernel:
```
adriaandegroot: There's not much point in expanding bad documentation with more filler. If you're going to add `:param` lines, please be better than previous authors and document what the parameters really do (presumably both are strings, used for .. something)

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
         kernel_params.append("resume=/dev/mapper/{!s}".format(
             swap_outer_mappername))
 
-    kernel_line = get_kernel_line(kernel_type)
+    kernel_line = entry_name + " " + kernel_version
```
adriaandegroot: Since this is a 3.3 PR **anyway**, you might want to look at Python's f-strings as well. That makes string-formatting and concatenation just a little easier to read (e.g. `f"{entry_name} {kernel_version}"`) -- easier to read once you're used to it, I suppose.

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+        initrd = "initrd"
+        mkinitcpio_option = ""
 
-    conf_path = os.path.join(install_path + efi_dir,
-                             "loader",
-                             "entries",
-                             entry_name + ".conf")
+    # get the machine-id
+    with open(os.path.join(install_path, "etc", "machine-id"), 'r') as machineid_file:
+        machine_id = machineid_file.read().rstrip('\n')
```
adriaandegroot, line 146: What if the file is not there (yet)? Do we assume the *machineid* module is run before this one, or the installer has some other mechanism to generate the machineid? (You can set some kind of dependencies between modules in `module.desc`)

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
 
-    # Copy kernel and initramfs to a subdirectory of /efi partition
-    files_dir = os.path.join(install_path + efi_dir, entry_name)
-    os.mkdir(files_dir)
+    # Copy kernel to a subdirectory of /efi partition
+    machine_dir = os.path.join(install_path + efi_dir, machine_id)
+    try:
+        os.mkdir(machine_dir)
```
adriaandegroot: `os.makedirs(machine_dir, exist_ok=True)`

(Not a blocker; could-be-better)

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+    except FileExistsError: # We can ignore errors caused by the directory existing already
+        pass
 
-    kernel_path = install_path + kernel
+    files_dir = os.path.join(machine_dir, kernel_version)
+    try:
+        os.mkdir(files_dir)
```
adriaandegroot: .. unless you specifically need the name `machine_dir`, you can skip over that one and go straight to

```
files_dir = os.path.join(install_path, efi_dir, machine_id, kernel_version)
os.makedirs(files_dir, exist_ok=True)
```

to make all the intermediate ones in one go, without worrying about `FileExistsError`. (Assumes *efi_dir* is supposed to be its own directory)

(Not a blocker)

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+    kernel_path = os.path.join(install_path, kernel)
     kernel_name = os.path.basename(kernel_path)
-    shutil.copyfile(kernel_path, os.path.join(files_dir, kernel_name))
-
-    img_path = install_path + img
-    img_name = os.path.basename(img_path)
-    shutil.copyfile(img_path, os.path.join(files_dir, img_name))
+    shutil.copyfile(kernel_path, os.path.join(files_dir, "linux"))
 
+    # generate the initramfs - this is Arch specific and should be replaced for other distros
```
adriaandegroot: This will block the PR though: having to edit Python code for other distro's isn't acceptable.

**Why** is this arch-specific? (It looks like a pretty generic call to mkinitcpio to me) Does the initcpiocfg / initcpio module not do enough for you? Is it essential to do the mkinitcpio **here**, **now**, or could it be done later in a dedicated module?

I'm asking that last question especially because there's configuration options elsewhere that might be able to handle the "arch-specificness" making the rest of this PR acceptable as-is.

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+        "linux      {!s}\n".format(os.path.join("/", machine_id, kernel_version, "linux")),
     ]
 
+    try:
+        additional_initrd_files = libcalamares.job.configuration["additionalInitrdFiles"]
+        for initrd_file in additional_initrd_files.split(','):
+            libcalamares.utils.debug("Attempting to handle initrd image " + initrd_file)
+            if os.path.isfile(os.path.join(install_path, initrd_file)):
+                libcalamares.utils.debug("Found image " + initrd_file)
+                shutil.copyfile(os.path.join(install_path, initrd_file), os.path.join(files_dir, os.path.basename(initrd_file)))
```
adriaandegroot: By this point I'd forgotten what `files_dir` was; can we find a better name for that? It is the "EFI path for this bootloader and its files", maybe `efi_dir_for_bootloader_files`? (Not a blocker)

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+                lines.append("initrd     {!s}\n".format(os.path.join("/", machine_id, kernel_version, os.path.basename(initrd_file))))
+    except KeyError: # If the configuration option isn't set, we can just move on
```
adriaandegroot, line 177: Thank you for taking care with KeyError; many of the Python modules are sensitive to having all their configuration keys, and your code makes using this module in the face of "I don't care about additional Initrd" **much** nicer.

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
-
-    img_path = install_path + img
-    img_name = os.path.basename(img_path)
-    shutil.copyfile(img_path, os.path.join(files_dir, img_name))
+    shutil.copyfile(kernel_path, os.path.join(files_dir, "linux"))
 
+    # generate the initramfs - this is Arch specific and should be replaced for other distros
+    try:
+        subprocess.run(["chroot " +
+                        install_path +
```
adriaandegroot: Maybe we should rename `install_path` to `target_system_root_mountpoint` to make the code more self-documenting.

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
-    img_path = install_path + img
-    img_name = os.path.basename(img_path)
-    shutil.copyfile(img_path, os.path.join(files_dir, img_name))
+    shutil.copyfile(kernel_path, os.path.join(files_dir, "linux"))
 
+    # generate the initramfs - this is Arch specific and should be replaced for other distros
+    try:
+        subprocess.run(["chroot " +
+                        install_path + 
+                        " mkinitcpio -k " + kernel_version + " -g " + os.path.join("/", os.path.relpath(files_dir, install_path), initrd)],
```
adriaandegroot: That join-and-relpath looks really weird. Er .. maybe not, since `files_dir` is a subdirectory of `install_path` (those are "where is EFI mounted right now in the target system" and "target system root mount" respectively). How about `assert(files_dir.startswith(install_path))` and using `files_dir[len(install_path):]` ? Needs a comment, anyway, that that second path is relative to the target.

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+        libcalamares.utils.debug("STDERR: " + cpe.stderr.decode())
+        raise
+
+
+    # write the entry
     lines = [
-        '## This is just an example config file.\n',
-        '## Please edit the paths and kernel parameters according\n',
-        '## to your system.\n',
+        '## Generated by Calamares\n',
```
adriaandegroot, line 161: +1

src/modules/bootloader/main.py
```diff
@@ -145,43 +127,84 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+                shutil.copyfile(os.path.join(install_path, initrd_file), os.path.join(files_dir, os.path.basename(initrd_file)))
+                lines.append("initrd     {!s}\n".format(os.path.join("/", machine_id, kernel_version, os.path.basename(initrd_file))))
+    except KeyError: # If the configuration option isn't set, we can just move on
+        libcalamares.utils.debug("Failed to find key additionalInitrdFiles")
+        pass
+    
+    lines.append("initrd     {!s}\n".format(os.path.join("/", machine_id, kernel_version, initrd)))
+    lines.append("options    {!s} rw\n".format(" ".join(kernel_params)))
+
+    conf_path = os.path.join(install_path + efi_dir,
```
adriaandegroot: Seems weird to mix `join()` and `+`, plus this has been calculated before -- how about a `target_efi_top_dir` or so? Those `target_`* variables could be placed together so it is clear how the structure works.

src/modules/bootloader/main.py
```diff
@@ -265,19 +290,30 @@ def install_systemd_boot(efi_directory):
...
+            if file == source_kernel_name:
+                rel_root = os.path.relpath(root, install_path)
+                create_systemd_boot_conf(install_path,
+                                        efi_directory,
+                                        uuid,
+                                        distribution,
+                                        distribution_translated,
+                                        os.path.join(rel_root, file),
+                                        "default",
+                                        os.path.basename(root))
```
adriaandegroot: Seeing these long parameter lists makes me thing that kwargs-style might be a more readable approach here. Maybe it's just because the PR review doesn't offer me the convenience of an IDE to help me understand what does what.

src/modules/bootloader/main.py
```diff
@@ -265,19 +290,30 @@ def install_systemd_boot(efi_directory):
...
-                                 distribution,
-                                 distribution_translated,
-                                 "fallback")
+
+    # find all the installed kernels and generate default and fallback entries for each
+    # This is Arch-specific and may need adjustment for other distros
+    for root, dirs, files in os.walk(os.path.join(install_path, kernel_search_path.lstrip('/'))): 
+        for file in files:  
+            if file == source_kernel_name:
+                rel_root = os.path.relpath(root, install_path)
```
adriaandegroot: The four lines above are used to generate two variables for use in the calls to `create_systemd_boot_conf()`, namely *file* and *rel_root*.  Then there's two variants / flavors, namely "default" and "fallback".

A function returning a list of 3-tuples could factor out some of the Arch-ness: on arch, do this `os.walk()`, elsewhere return a list with just the default flavor (and maybe "fallback").

adriaandegroot (changes requested):
I like it in general, with three points (which you can find in the detailed review comments):
- this is a 3.3-timeframe change because the schema / configuration changes
- blocked by arch-specifics which would need tidying up or making-configurable; some of the arch-specifics might be movable to other modules
- various tidbits and the source could use more overall cleanup (but that's not necessarily on you)

src/modules/bootloader/main.py
dalto8: I suppose it isn't Arch specific.  It is specific to a distro using both `mkinitcpio` and `kernel-install` which is still fairly specific.

It doesn't have to be done here.  It could be done before or after this point.  It needs the kernels to be installed and the machine id to be generated.

As for using the existing initcpio module, the problem would be that it would need to be refactored substantially in order to add the ability to search for all the kernels to know where to write the initrd.  In c++, that would be a decent amount of code.  It would also mean all that logic was duplicated in two places, once in c++ and once in python.

I am not broadly familiar with Calamares outside of this module.  Is there a better way to approach it?

src/modules/bootloader/main.py
dalto8: Yes, that is a much better idea.  I wasn't familiar with os.makedirs.

src/modules/bootloader/main.py
dalto8: That is a great idea.  If we can devise a strategy for resolving the mkinitcpio problem above, there would be nothing left that isn't generic.

src/modules/bootloader/main.py
dalto8: Sorry about that, I tried to stay consistent with what was already there.  I can clean up and document all the params here.

src/modules/bootloader/main.py
dalto8: I should add some better error handling there but that wouldn't really resolve the issue since we need the machine-id.

I can add the module dependency if that would be an acceptable solution.

src/modules/bootloader/main.py
dalto8: Yes, that isn't documented very well.  I sort of assumed it would need to be refactored since it is overly specific.

To be clear, the `relpath` is producing a path suitable for use in the chroot by removing `install_path`.  I think it could be made self documenting by making the relpath call a separate statement with a descriptive variable name.

adriaandegroot:
@dalto8 @killajoe Can you tweak / massage this PR so it can go in? I got the impression it was pretty close and would only need some minor twiddles to be more applicable.

In the meantime, it would be nice if you re-target it to the *work-3.3* branch, because that one does not care about Python-version-bumps and things.

dalto8:
> @dalto8 @killajoe Can you tweak / massage this PR so it can go in? I got the impression it was pretty close and would only need some minor twiddles to be more applicable.

Yes.  I will go through it and resolve all the open issues.

> In the meantime, it would be nice if you re-target it to the _work-3.3_ branch, because that one does not care about Python-version-bumps and things.

I can do that.

killajoe:
yes i am in if i can 😌

dalto8:
I think this is ready for review again.

One problem is that I have some commits here from master.  I am not entirely sure how to fix that from a git perspective.

kkofler:
You should rebase your `non-EOS` branch onto `work-3.3` and force-push it (the `non-EOS` branch) when done.

dalto8:
> You should rebase your `non-EOS` branch onto `work-3.3` and force-push it (the `non-EOS` branch) when done.

I rebased off the work-3.3 and force pushed the changes but github is still showing 21 changed files so I may have missed something.

Any tips on how to clean this up from a git perspective would be appreciated.

kkofler:
You can do an interactive rebase, in which you can skip unwanted commits and retain only the relevant ones.

dalto8:
> You can do an interactive rebase, in which you can skip unwanted commits and retain only the relevant ones.

Thanks for the advice.  That cleaned it up.

I think this is ready for review now.

src/modules/bootloader/bootloader.conf
```diff
@@ -9,19 +9,19 @@
...
-kernel: "/vmlinuz-linux"
-img: "/initramfs-linux.img"
-fallback: "/initramfs-linux-fallback.img"
+kernelSearchPath: "/usr/lib/modules"
+kernelName: "vmlinuz"
 timeout: "10"
 
-# Optionally set the menu entry name and kernel name to use in systemd-boot.
+# additionalInitrdFiles is a comma seperated list of file names
+additionalInitrdFiles: "/boot/amd-ucode,/boot/intel-ucode"
```
adriaandegroot: If this is supposed to be a list? Why not make it a list in YAML then? Notation is with bullets or `[ ]`, and then you get a list in the job-configuration, rather than a string that you need to split and handle separately.

src/modules/bootloader/main.py
```diff
@@ -73,37 +73,18 @@ def get_bootloader_entry_name():
...
-    :param entry:
-    :param entry_name:
-    :param kernel_type:
+    :param installation_root_path: A string containing the absolute path to the root of the installation
+    :param efi_dir: A string containing the path to the efi dir relative to the root of the installation
+    :param uuid: A string containing the UUID of the root volume
+    :param entry: A string containing the name of the entry as it will be displayed on boot
+    :param kernel: A string containing the path to the kernel relative to the root of the installation
+    :param kernel_type: A string which should be set if there is a special version of the entry, for example "fallback"
+    :param kernel_version: The kernel version string
```
adriaandegroot, line 86: Thank you for useful docstrings.

adriaandegroot (commented):
LGTM, with the one this-could-be-a-list as a non-blocking suggestion for improvement.

adriaandegroot (approved):

dalto8:
I made that last change you requested and rebased it again.  Let me know if there is anything that needs to be done before it can be merged.

PR #1647, opened 2021-02-27
[finishedq] adding QML finished module

demmm:
module builds & runs, config connections are not registering
no errors
finishedq.qml is offering a different option though, running commands directly in qml
plasma-framework executer is used for that


demmm:
Image of finishedq (with sed line used to set the `Done` button to invisible):
Screenshot_20210227_223547

PR #1652, opened 2021-03-11
[packages] Drop urpmi support

Conan-Kudo:
This code is essentially untested and unused, as OpenMandriva has been using DNF for three years now.

Reference: https://www.openmandriva.org/en/news/article/switching-to-rpmv4

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -356,26 +356,6 @@ class PMPortage(PackageManager):
         pass
 
 
-class PMUrpmi(PackageManager):
-    backend = "urpmi"
-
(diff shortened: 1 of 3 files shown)


kkofler:
What about Mageia? As far as I can tell, they have both DNF and urpmi available. And are there any other Mandriva derivatives left using urpmi?

Conan-Kudo:
Mageia does not have Calamares and is not interested in it, and there are no other derivatives using URPMI.

PackageKit removed the urpmi backend for similar reasons: https://github.com/hughsie/PackageKit/commit/6a837b5c5b7a5f1aefa48aae1b6a70d8e1436b92

PR #1666, opened 2021-03-29
[logUpload] Limit upload size

deprov447:
Implementing a configuration key `sizeLimit` from `branding.desc` to limit maximum size of logs to be uploaded.

### DIFF
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -220,13 +220,17 @@ slideshowAPI: 2
 
 
 # These options are to customize online uploading of logs to pastebins:
-#  - type : Defines the kind of pastebin service to be used. Currently
-#           it accepts two values:
-#           - none    :    disables the pastebin functionality
-#           - fiche   :    use fiche pastebin server
-#  - url  : Defines the address of pastebin service to be used.
-#           Takes string as input. Important bits are the host and port,
-#           the scheme is not used.
+#  - type      : Defines the kind of pastebin service to be used. Currently
+#                it accepts two values:
+#                - none    :    disables the pastebin functionality
+#                - fiche   :    use fiche pastebin server
+#  - url       : Defines the address of pastebin service to be used.
+#                Takes string as input. Important bits are the host and port,
+#                the scheme is not used.
+#  - sizeLimit : Defines maximum size limit (in KiB) of log file to be pasted. 
+#                Takes integer as input. If <=0, no limit will be forced,
+#                else only last 'n' KiB of log file will be pasted. 
 uploadServer :
     type :    "fiche"
     url :     "http://termbin.com:9999"
+    sizeLimit : 20
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -227,9 +227,11 @@ slideshowAPI: 2
 #  - url       : Defines the address of pastebin service to be used.
 #                Takes string as input. Important bits are the host and port,
 #                the scheme is not used.
-#  - sizeLimit : Defines maximum size limit (in KiB) of log file to be pasted. 
-#                Takes integer as input. If <=0, no limit will be forced,
-#                else only last 'n' KiB of log file will be pasted. 
+#  - sizeLimit : Defines maximum size limit (in KiB) of log file to be pasted.
+#                Takes integer as input. If < 0, no limit will be forced,
+#                else only last (approximately) 'n' KiB of log file will be pasted.
+#                Please note that upload size may be slightly over the limit (due
+#                to last minute logging), so provide a suitable value.
 uploadServer :
     type :    "fiche"
     url :     "http://termbin.com:9999"
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -233,4 +233,4 @@ slideshowAPI: 2
 uploadServer :
     type :    "fiche"
     url :     "http://termbin.com:9999"
-    sizeLimit : 20
+    sizeLimit : -1
--- a/src/libcalamaresui/Branding.h
+++ b/src/libcalamaresui/Branding.h
@@ -223,8 +223,9 @@ public:
 
     /** @brief Upload server configuration
      *
-     * This is both the type (which may be none, in which case the URL
-     * is irrelevant and usually empty) and the URL for the upload.
+     * This object has 3 items : the type (which may be none, in which case the URL
+     * is irrelevant and usually empty), the URL for the upload and the size limit of upload
+     * in bytes (for configuration value < 0, it serves -1, which stands for having no limit).
      */
     using UploadServerInfo = std::tuple< UploadServerType, QUrl, qint64 >;
     UploadServerInfo uploadServer() const { return m_uploadServer; }
--- a/src/libcalamaresui/Branding.h
+++ b/src/libcalamaresui/Branding.h
@@ -226,7 +226,7 @@ public:
      * This is both the type (which may be none, in which case the URL
      * is irrelevant and usually empty) and the URL for the upload.
      */
-    using UploadServerInfo = QPair< UploadServerType, QUrl >;
+    using UploadServerInfo = std::tuple< UploadServerType, QUrl, qint64 >;
     UploadServerInfo uploadServer() const { return m_uploadServer; }
 
     /**
--- a/src/libcalamaresui/ViewManager.cpp
+++ b/src/libcalamaresui/ViewManager.cpp
@@ -144,7 +144,7 @@ void
 ViewManager::onInstallationFailed( const QString& message, const QString& details )
 {
     bool shouldOfferWebPaste
-        = Calamares::Branding::instance()->uploadServer().first != Calamares::Branding::UploadServerType::None;
+        = std::get<2>(Calamares::Branding::instance()->uploadServer()) != Calamares::Branding::UploadServerType::None;
 
     cError() << "Installation failed:" << message;
     cDebug() << Logger::SubEntry << "- message:" << message;
--- a/src/libcalamaresui/ViewManager.cpp
+++ b/src/libcalamaresui/ViewManager.cpp
@@ -143,9 +143,9 @@ ViewManager::insertViewStep( int before, ViewStep* step )
 void
 ViewManager::onInstallationFailed( const QString& message, const QString& details )
 {
-    bool shouldOfferWebPaste
-        = std::get<0>(Calamares::Branding::instance()->uploadServer()) != Calamares::Branding::UploadServerType::None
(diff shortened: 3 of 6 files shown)


src/libcalamaresui/utils/Paste.cpp
```diff
@@ -40,11 +40,15 @@ logFileContents()
         return QByteArray();
     }
     QFileInfo fi( pasteSourceFile );
-    if ( fi.size() > 16_KiB )
+    sizeLimit = ( sizeLimit < 0 ) ? 1024*1024 : sizeLimit * 1024;            //For KiB to bytes
```
adriaandegroot: The documentation says "no limit", while this implementation turns it into 1024KiB.

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -30,7 +30,7 @@ using namespace CalamaresUtils::Units;
  * Returns an empty QByteArray() on any kind of error.
  */
 STATICTEST QByteArray
-logFileContents()
+logFileContents( qint64 sizeLimit )
```
adriaandegroot: Since this is intended to be in KiB, it's probably useful to name it `sizeLimitKiB` or so.

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -40,11 +40,15 @@ logFileContents()
         return QByteArray();
     }
```
adriaandegroot, line 45: If `sizeLimit` is 0, you can also bail out early.

src/libcalamaresui/utils/Paste.cpp
adriaandegroot: Hm .. I was also going to suggest using (from `utils/Units.h`) function `KiBtoBytes()`, but that takes unsigned. You would need to cast `sizeLimit` to unsigned for that (which is ok: in that branch of the ternary, you know it is non-negative)

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -40,11 +40,15 @@ logFileContents()
...
     QFileInfo fi( pasteSourceFile );
-    if ( fi.size() > 16_KiB )
+    sizeLimit = ( sizeLimit < 0 ) ? 1024*1024 : sizeLimit * 1024;            //For KiB to bytes
+    cDebug() << "Log upload size limit was set to " << sizeLimit << " bytes";
+    if ( fi.size() > sizeLimit and sizeLimit > 0 )
     {
-        pasteSourceFile.seek( fi.size() - 16_KiB );
+        // Fixme : this following line is not getting pasted
```
adriaandegroot: Should be `FIXME` (until it's fixed :) )

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -40,11 +40,15 @@ logFileContents()
...
+        cDebug() << "Only last " << sizeLimit << " bytes of log file (" << fi.size() << ") uploaded" ;
```
adriaandegroot: For debugging this (and the FIXME) you should write a static, C-style function that does a `stat()` on the log file (get its name via `Logger::logFile()`) and `fprintf(stderr, ..)` the results.

What I suspect is this: `QFileInfo` does the `stat()` call in its constructor; let's say it finds the file to be exactly 70000 bytes long. Now you log one extra line, so you have about 70060 bytes in the file. Below, we `seek()` to the size from the fileinfo, minus the size limit: but that is `70000 - sizeLimit`, not the **current** size of the file after that debug line! So you never see that line in the uploaded log, even though it might be there in the log at the moment of upload.

Another way of testing this is to seek to `fi.size()-sizeLimit+10`, or to read `sizeLimit+10`, just so see if the bytes are there. Come to think of it, reading `sizeLimit+1_KiB` might even be a good thing **anyway** in case new logging was done from another thread.

adriaandegroot: You don't need extra spaces between items; `std::cout` does not introduce spaces between items, but QDebug does (by default, although there is a NoQuote item you can send to turn that off).

adriaandegroot:
I'm going to backpedal a little on some of my comments: I think the API should send around **bytes** for sizeLimit,  not KiB.

The reason for that can be seen in the unit tests:
```
QByteArray contentsOfLogfileAfterSetup = logFileContents( 16 );
```
16? 16 what? If we were to write 16_KiB it would be immediately clear (well, clearer that this is byte count).

The change that is needed for that is to move the convert-value-to-KiB into the part that reads the **branding** info and returns the `UploadServerInfo`. That is, in `uploadServerFromMap()`, take the size limit from the config file, and convert right there into bytes: -1 for negative values, 0 for 0, * 1024 (use `KiBtoBytes()`) for others. Then any consumers of an `UploadServerInfo` object don't need to worry about KiB-vs-anything else: it's just a byte count.

[[ Somewhat related: the docs for `UploadServerInfo` are still for the pair-valued one, not the 3-tuple. ]]

adriaandegroot:
Overall: looks pretty good, could be better in some spots, I support your wish for FIXME -- I think that you can hunt that down with some debugging experiments -- and a little bit of polish will make it just lovely.

src/branding/default/branding.desc
```diff
@@ -220,13 +220,17 @@ slideshowAPI: 2
...
+#  - type      : Defines the kind of pastebin service to be used. Currently
+#                it accepts two values:
+#                - none    :    disables the pastebin functionality
+#                - fiche   :    use fiche pastebin server
+#  - url       : Defines the address of pastebin service to be used.
+#                Takes string as input. Important bits are the host and port,
+#                the scheme is not used.
+#  - sizeLimit : Defines maximum size limit (in KiB) of log file to be pasted. 
+#                Takes integer as input. If <=0, no limit will be forced,
+#                else only last 'n' KiB of log file will be pasted.
```
adriaandegroot: The comment now says that 0 is "no limit", but the implementation uses 0 as "no data", so there's a mismatch.

adriaandegroot: Probably also worthwhile mentioning that slightly more than the limit may be uploaded, due to last-minute logging.

src/libcalamaresui/Branding.h
```diff
@@ -226,7 +226,7 @@ class UIDLLEXPORT Branding : public QObject
      * This is both the type (which may be none, in which case the URL
      * is irrelevant and usually empty) and the URL for the upload.
      */
```
adriaandegroot, line 229: Still needs documentation for that third element in the tuple (which you have kept as "number of kibibytes")

src/libcalamaresui/ViewManager.cpp
```diff
@@ -144,7 +144,7 @@ void
 ViewManager::onInstallationFailed( const QString& message, const QString& details )
 {
     bool shouldOfferWebPaste
-        = Calamares::Branding::instance()->uploadServer().first != Calamares::Branding::UploadServerType::None;
+        = std::get<0>(Calamares::Branding::instance()->uploadServer()) != Calamares::Branding::UploadServerType::None;
```
adriaandegroot: Depending on the meaning of `sizeLimit: 0` (according to the docs, or according to the implementation), you could also check that here.

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -40,11 +42,16 @@ logFileContents()
         return QByteArray();
     }
     QFileInfo fi( pasteSourceFile );
-    if ( fi.size() > 16_KiB )
+    if( sizeLimitKiB < 0 )
+        sizeLimitKiB = 1024;
```
adriaandegroot: This deserves a test as well (in `TestPaste.cpp`)

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -40,11 +42,16 @@ logFileContents()
...
+    qint64 sizeLimitBytes = CalamaresUtils::KiBtoBytes( ( unsigned long long ) sizeLimitKiB );
```
adriaandegroot: Please use `static_cast<>` rather than C-style cast. The `sizeLimitBytes` can be const, as well.

src/libcalamaresui/utils/Paste.cpp
```diff
@@ -40,11 +42,16 @@ logFileContents()
...
+    cDebug() << "Log upload size limit was set to" << sizeLimitKiB << "KiB";
+    if ( fi.size() > sizeLimitBytes and sizeLimitBytes > 0 )
     {
-        pasteSourceFile.seek( fi.size() - 16_KiB );
+        cDebug() << "Only last" << sizeLimitBytes << "bytes of log file (sized" << fi.size() << "bytes) uploaded" ;
+        pasteSourceFile.seek( fi.size() - sizeLimitBytes + 1_KiB );
```
adriaandegroot: Changing the seek **and** changing the read (below) seems to defeat the purpose: suppose the log file was 64KiB and the sizeLimit is 16KiB, then:
- you seek to 49KiB
- if there's been no additional writes to the log file, you read 17KiB, and get 15KiB
- if there's been writes, less than 1KiB you read 17KiB and still get < 16KiB
Your original problem here was that that last debug-line (about how much was being uploaded) was itself not included in the web-paste. Did you figure out whether it is `fi.size()` that is wrong, here, or something else? Seeking to 48KiB and then reading 17KiB will get you at least the 16KiB requested -- but it's up to you what it does exactly, as long as it's documented :)

adriaandegroot (commented):
Also needs 1 pass with the coding style tool

src/libcalamaresui/utils/Paste.cpp
deprov447: You asked to convert KiB to bytes in `Calamares::Branding::uploadServerFromMap()` for _not having to_ care about units everytime UploadServerInfo object is made. But _utils_ in _libcalamaresui_ do not provide unit converter. So either I have to explicitly convert them (*1024) or call utils from _libcalamares_. I guess 1st one would be better ?

deprov447: fixed. used `pasteSourceFile.readAll()`

src/libcalamaresui/utils/Paste.cpp
deprov447: actually I found we just have to refresh QFileInfo fi. Now uploads the log upto "_Uploading x bytes of logs_". 
image

src/libcalamaresui/utils/Paste.cpp
adriaandegroot: `src/libcalamares/utils/Units.h` (due to the way include paths are, just `#include "utils/Units.h"` ), then `CalamaresUtils::Units::KiBtoBytes()`

Issue #847, opened 2017-10-27
Allow selecting  the FS from GUI


abucodonosor:
#### Submission type

  - [X] Feature Request

Right now we have a static 'defaultFileSystemType: ..' value in packages.conf
and fallback to ext4 in PartitionViewStep.

It would be nice to allow selecting the FS from a defined list from the Partition GUI.
I don't think we should force users to use 'Manual Partitioning' just to select the
some other FS.

So Distributors can have something like this:

```
   supportedFileSystems:
            - FS1
            - FS2
            - FS3
            ....
```

First FS would be default one and the user could switch to any supported 
FS from the partition GUI.

adriaandegroot:
This will probably require changes to the partition "bars view" widget, to allow changing the filesystem. I consider it fairly low-priority: it introduces a fair bit of complexity with -- for me -- no visible benefit to distro's. Sure, it empowers a certain kind of power user who knows and wants to set up something specific, but this leads us straight towards supporting nested plausibly-deniable LUKS volumes, which is something we recently refused under "not a goal with Calamares".

abucodonosor:
Well this is not meant for the poweruser but for normal users,

A poweruser won't care about what defauts are he'll go to manual partitioning anyway.

A normal user may want to use ext4 over btrfs ( as example ) but has no clue how to
use the 'manual partitioning' mode.

kkofler:
A normal user should just use `ext4` and not pick some unreliable file system just because he read the buzzword somewhere. (It follows that the distro default should be sane, defaulting to `btrfs` is a very bad idea.)

jayarmstrong:
@adriaandegroot Yes, there is a gradient of popularity for options. Perhaps you'll allow users to vote on feature requests in some manner? It'd be a shame, IMO, to produce another installer that is as equally limited as the other projects. As for the benefit to distros, I'm here today, choosing my distro based on how well its installer works at solving my needs. I gave up on Maui and Neon because of Calamares bugs. That said, I really want to see Calamares flourish, mature, and deliver a better experience for desktop Linux.

@kkofler "should" and "bad" are your opinions. I don't see why you're bringing those into someone else's feature request.

adriaandegroot:
[...]

PR #1667, opened 2021-03-31
[RFC] [partition] FS choice for autopartition options

LordTermor:
Allows to choose filesystem when _erase disk_ is chosen (and potentially other autopartition options) and configure available filesystems using `partition.conf`.
изображение
изображение

### DIFF
--- a/src/modules/partition/core/Config.cpp
+++ b/src/modules/partition/core/Config.cpp
@@ -207,6 +207,15 @@ Config::setSwapChoice( Config::SwapChoice c )
     }
 }
 
+void
+Config::setEraseFsTypeChoice(const QString& choice)
+{
+    if(choice != m_eraseFsTypeChoice){
+        m_eraseFsTypeChoice = choice;
+    }
+}
+
+
 bool
 Config::allowManualPartitioning() const
 {
@@ -239,6 +248,17 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
     Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
     gs->insert( "allowManualPartitioning",
                 CalamaresUtils::getBool( configurationMap, "allowManualPartitioning", true ) );
+
+    if ( configurationMap.contains( "availableFileSystemTypes" ) ) {
+        if(configurationMap.value(  "availableFileSystemTypes" ).type() == QVariant::List ){
+            m_eraseFsTypes.clear();
+            m_eraseFsTypes.append(configurationMap.value(  "availableFileSystemTypes" ).toStringList());
+            m_eraseFsTypeChoice = m_eraseFsTypes[0];
+        } else if ( configurationMap.value(  "availableFileSystemTypes" ).type() == QVariant::String  ){
+            m_eraseFsTypes.append(configurationMap.value(  "availableFileSystemTypes" ).toString());
+            m_eraseFsTypeChoice = m_eraseFsTypes[0];
+        }
+    }
 
     if ( configurationMap.contains( "requiredPartitionTableType" )
          && configurationMap.value( "requiredPartitionTableType" ).type() == QVariant::List )
--- a/src/modules/partition/core/PartitionActions.cpp
+++ b/src/modules/partition/core/PartitionActions.cpp
@@ -109,6 +109,10 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
         partType = isEfi ? PartitionTable::gpt : PartitionTable::msdos;
     }
 
+    FileSystem::Type type;
+    PartUtils::findFS(o.defaultFsType, &type);
+    core->initLayout( type == FileSystem::Unknown ? FileSystem::Ext4 : type);
+
     core->createPartitionTable( dev, partType );
 
     if ( isEfi )
--- a/src/modules/partition/gui/ChoicePage.cpp
+++ b/src/modules/partition/gui/ChoicePage.cpp
@@ -246,6 +246,14 @@ ChoicePage::setupChoices()
         m_eraseButton->addOptionsComboBox( m_eraseSwapChoiceComboBox );
     }
 
+    if ( m_config->eraseFsTypes().count() > 1)
+    {
+        m_eraseFsTypesChoiceComboBox = new QComboBox;
+        m_eraseFsTypesChoiceComboBox->addItems(m_config->eraseFsTypes());
+        connect(m_eraseFsTypesChoiceComboBox, &QComboBox::currentTextChanged, [this](const auto& currentString){m_config->setEraseFsTypeChoice(currentString); onActionChanged();});
+        m_eraseButton->addOptionsComboBox( m_eraseFsTypesChoiceComboBox );
+    }
+
     m_itemsLayout->addWidget( m_alongsideButton );
     m_itemsLayout->addWidget( m_replaceButton );
     m_itemsLayout->addWidget( m_eraseButton );
@@ -442,9 +450,8 @@ ChoicePage::applyActionChoice( InstallChoice choice )
     case InstallChoice::Erase:
     {
         auto gs = Calamares::JobQueue::instance()->globalStorage();
-
         PartitionActions::Choices::AutoPartitionOptions options { gs->value( "defaultPartitionTableType" ).toString(),
-                                                                  gs->value( "defaultFileSystemType" ).toString(),
+                                                                  m_config->eraseFsType(),
                                                                   m_encryptWidget->passphrase(),
                                                                   gs->value( "efiSystemPartition" ).toString(),
                                                                   CalamaresUtils::GiBtoBytes(
--- a/src/modules/partition/gui/ChoicePage.h
+++ b/src/modules/partition/gui/ChoicePage.h
@@ -139,6 +139,8 @@ private:
     Calamares::PrettyRadioButton* m_replaceButton;
     Calamares::PrettyRadioButton* m_somethingElseButton;
     QComboBox* m_eraseSwapChoiceComboBox;  // UI, see also m_eraseSwapChoice
+    QComboBox* m_eraseFsTypesChoiceComboBox;  // UI, see also m_eraseSwapChoice
+
 
     DeviceInfoWidget* m_deviceInfoWidget;
 
--- a/src/modules/partition/partition.conf
+++ b/src/modules/partition/partition.conf
@@ -137,7 +137,7 @@ initialSwapChoice: none
 # If nothing is specified, Calamares defaults to "ext4".
 #
 # Names are case-sensitive and defined by KPMCore.
-defaultFileSystemType:  "ext4"
+availableFileSystemTypes:  ["ext4", "btrfs", "f2fs"]
 
 # Show/hide LUKS related functionality in automated partitioning modes.
 # Disable this if you choose not to deploy early unlocking support in GRUB2
--- a/src/modules/partition/core/Config.h
+++ b/src/modules/partition/core/Config.h
@@ -54,6 +54,8 @@ public:
     static const NamedEnumTable< SwapChoice >& swapChoiceNames();
     using SwapChoiceSet = QSet< SwapChoice >;
 
+    using EraseFsTypesSet = QStringList;
+
     void setConfigurationMap( const QVariantMap& );
     void updateGlobalStorage() const;
 
@@ -94,6 +96,11 @@ public:
      */
     SwapChoice swapChoice() const { return m_swapChoice; }
 
+    EraseFsTypesSet eraseFsTypes() const {return m_eraseFsTypes; }
+
+    QString eraseFsType() const {return m_eraseFsTypeChoice; }
+
+
     ///@brief Is manual partitioning allowed (not explicitly disnabled in the config file)?
     bool allowManualPartitioning() const;
 
@@ -102,6 +109,7 @@ public Q_SLOTS:
     void setInstallChoice( InstallChoice );
     void setSwapChoice( int );  ///< Translates a button ID or so to SwapChoice
     void setSwapChoice( SwapChoice );
+    void setEraseFsTypeChoice( const QString& );
 
 Q_SIGNALS:
     void installChoiceChanged( InstallChoice );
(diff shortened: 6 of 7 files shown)


src/modules/partition/partition.conf
```diff
@@ -137,7 +137,7 @@ initialSwapChoice: none
 # If nothing is specified, Calamares defaults to "ext4".
 #
 # Names are case-sensitive and defined by KPMCore.
-defaultFileSystemType:  "ext4"
+availableFileSystemTypes:  ["ext4", "btrfs", "f2fs"]
```
Conan-Kudo, line 140: Does this imply order/preference?

Conan-Kudo (changes requested):

Conan-Kudo:
Also, if we're going to have _choices_ for autopart options, how do we define schemes for each of them?

src/modules/partition/partition.conf
abucodonosor: I don't think so but this should not be like this.

defaultFileSystemType should remain as is, IOW this is what the installer defaults too,
and availableFileSystemType should be called something like optionalFileSystemTypes.

abucodonosor:
> Also, if we're going to have _choices_ for autopart options, how do we define schemes for each of them?

Not possible with the current implementation. So this option should conflict with the schemes one, until
the scheme one gets UIX bits I guess.

src/modules/partition/partition.conf
Conan-Kudo: That makes more sense to me, yes.

LordTermor: Well it's still a draft and I wanted to discuss the name with the rest of contributors. I don't understand why there should be two config values? We can leave only one and use it as a string in cases when there should be only one option.

LordTermor:
> 
> 
> > Also, if we're going to have _choices_ for autopart options, how do we define schemes for each of them?
> 
> Not possible with the current implementation. So this option should conflict with the schemes one, until
> the scheme one gets UIX bits I guess.

Where can I read more about schemes?

abucodonosor:
> > > Also, if we're going to have _choices_ for autopart options, how do we define schemes for each of them?
> > 
> > 
> > Not possible with the current implementation. So this option should conflict with the schemes one, until
> > the scheme one gets UIX bits I guess.
> 
> Where can I read more about schemes?

See the partition.conf & partitionLayout section.

src/modules/partition/partition.conf
abucodonosor: If you want to remove the default option then you have to change a lot of logic all over the place,
including the forced usage of ext4 in the case of missing default FS. 
Not to mention by just removing it without deprecation or automagically replace,
all existing downstream calamares configuration packages will break, or do at least very unexpected things.

Just do something like 

git grep defaultFileSystemType
git grep ext4

in the calamares source code.

hphilm: > I don't think so but this should not be like this.
> 
> defaultFileSystemType should remain as is, IOW this is what the installer defaults too,
> and availableFileSystemType should be called something like optionalFileSystemTypes.

Isn't that not the exact request you did [2017](https://github.com/calamares/calamares/issues/847) regarding that topic? Default is the first FS you specify. If none is specified it defaults to Ext4.

hphilm:
Seems we might need adjustments to the [template schema](https://github.com/calamares/calamares/blob/calamares/src/modules/partition/partition.schema.yaml#L20) anyway, regarding default FS type ...

src/modules/partition/partition.conf
LordTermor: > If you want to remove the default option then you have to change a lot of logic all over the place,
> including the forced usage of ext4 in the case of missing default FS.
> Not to mention by just removing it without deprecation or automagically replace,
> all existing downstream calamares configuration packages will break, or do at least very unexpected things.
> 
> Just do something like
> 
> git grep defaultFileSystemType
> git grep ext4
> 
> in the calamares source code.

Well deprecation thing is ok. It's not too much places where defaultFileSystemType is used I think I can address them all. I just think that leaving and using two options is a bit dirty.

abucodonosor: At the time nothing like the partition scheme existed among many other changes in this area.
But what I requested there is an additional list of supported FSes by the distribution.

abucodonosor: > Well deprecation thing is ok. It's not too much places where defaultFileSystemType is used I think I can address them all. I just think that leaving and using two options is a bit dirty.

Or maybe hide it in the code.. Eg: we remove the configurationMap check for it, then set it to the first fs in the new_option_of_FSes, with a fallback to ext4 if everything fails ( config bad or whatever else ), that way not too 
much logic needs to change.

src/modules/partition/core/PartitionActions.cpp
```diff
@@ -109,6 +109,10 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
         partType = isEfi ? PartitionTable::gpt : PartitionTable::msdos;
     }
 
+    FileSystem::Type type;
+    PartUtils::findFS(o.defaultFsType, &type);
+    core->initLayout( type == FileSystem::Unknown ? FileSystem::Ext4 : type);
```
hphilm, line 114: I think the fallback filesystem is addressed here ...

src/modules/partition/partition.conf
adriaandegroot: The point is that **existing** configs which set *defaultFileSystem* now don't set anything at all. The configuration (`setConfigurationMap()`) code could check for that key and decide it means a one-element list, but that does need to be spelled out.

src/modules/partition/core/PartitionActions.cpp
LordTermor: It's for the case when fs string in config is invalid

src/modules/partition/core/Config.cpp
```diff
@@ -207,6 +207,15 @@ Config::setSwapChoice( Config::SwapChoice c )
     }
 }
 
+void
+Config::setEraseFsTypeChoice(const QString& choice)
```
adriaandegroot, line 211: Apply the coding-style tool (I'll just say that once to get it over with; `ci/calamaresstyle <dirname>` will do the trick, assuming you have astyle + clang-format10 or so.

src/modules/partition/core/Config.cpp
```diff
@@ -207,6 +207,15 @@ Config::setSwapChoice( Config::SwapChoice c )
...
+{
+    if(choice != m_eraseFsTypeChoice){
+        m_eraseFsTypeChoice = choice;
```
adriaandegroot, line 214: I realise that this "config" object does not follow the style of most other config classes, but normally there would be an `emit eraseFsTypeChoiceChanged()` here. It *might* make sense to start introducing that style to the config object.

src/modules/partition/core/Config.cpp
```diff
@@ -207,6 +207,15 @@ Config::setSwapChoice( Config::SwapChoice c )
     }
 }
 
+void
+Config::setEraseFsTypeChoice(const QString& choice)
+{
+    if(choice != m_eraseFsTypeChoice){
```
adriaandegroot, line 213: Is there an idea of "valid choices"? Will any string do?

src/modules/partition/core/Config.cpp
```diff
@@ -240,6 +249,17 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
     gs->insert( "allowManualPartitioning",
                 CalamaresUtils::getBool( configurationMap, "allowManualPartitioning", true ) );
 
+    if ( configurationMap.contains( "availableFileSystemTypes" ) ) {
+        if(configurationMap.value(  "availableFileSystemTypes" ).type() == QVariant::List ){
+            m_eraseFsTypes.clear();
+            m_eraseFsTypes.append(configurationMap.value(  "availableFileSystemTypes" ).toStringList());
```
adriaandegroot, line 255: These two lines can be one assignment instead of clear + append; you could use `getStringList()` although it is a bit wordy.

src/modules/partition/core/Config.cpp
```diff
@@ -240,6 +249,17 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
...
+            m_eraseFsTypeChoice = m_eraseFsTypes[0];
```
adriaandegroot, line 256: You can lift this one out of (both branches of the) if and put it after the is-this-a-list-or-a-single-string checks.

adriaandegroot, line 256: If the list is empty this is still UB (e.g. if the user has the terrible configuration `availableFileSystemTypes: []`)

src/modules/partition/core/Config.cpp
```diff
@@ -240,6 +249,17 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
...
+        } else if ( configurationMap.value(  "availableFileSystemTypes" ).type() == QVariant::String  ){
+            m_eraseFsTypes.append(configurationMap.value(  "availableFileSystemTypes" ).toString());
```
adriaandegroot, line 258: Why **isn't** the list cleared here, first? It could be an assignment, too.

src/modules/partition/core/Config.cpp
```diff
@@ -240,6 +249,17 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
...
+            m_eraseFsTypeChoice = m_eraseFsTypes[0];
```
adriaandegroot, line 259: .. come to think of it, calling the setter is probably better, since you might want to emit a signal; this depends a little bit on what's in the setter and whether everything is fully set-up for calling it yet.

src/modules/partition/gui/ChoicePage.h
```diff
@@ -139,6 +139,8 @@ private slots:
     Calamares::PrettyRadioButton* m_replaceButton;
     Calamares::PrettyRadioButton* m_somethingElseButton;
     QComboBox* m_eraseSwapChoiceComboBox;  // UI, see also m_eraseSwapChoice
+    QComboBox* m_eraseFsTypesChoiceComboBox;  // UI, see also m_eraseSwapChoice
```
adriaandegroot, line 142: Comment mismatch. Also, start to prefer `= nullptr` initializations for pointers (esp. relevant because you didn't initialize it in the constuctor).

src/modules/partition/gui/ChoicePage.cpp
```diff
@@ -246,6 +246,14 @@ ChoicePage::setupChoices()
         m_eraseButton->addOptionsComboBox( m_eraseSwapChoiceComboBox );
     }
 
+    if ( m_config->eraseFsTypes().count() > 1)
+    {
+        m_eraseFsTypesChoiceComboBox = new QComboBox;
+        m_eraseFsTypesChoiceComboBox->addItems(m_config->eraseFsTypes());
+        connect(m_eraseFsTypesChoiceComboBox, &QComboBox::currentTextChanged, [this](const auto& currentString){m_config->setEraseFsTypeChoice(currentString); onActionChanged();});
```
adriaandegroot, line 253: This looks like a case for:
- connecting directly to `&Config::setEraseFsTypeChoice`
- giving the config a signal to .. signal changes to the erase choice
- connecting the change signal to `onActivate()`

adriaandegroot:
Open conversation topic: **existing configurations**

- An existing configuration can (should) have a *defaultFileSystemType*.
- If that is set and no erase-fs-type is set, then what? (I'd interpret it as "one item erase-fs-type list")
- if it is set and also erase-fs-type is set, then what? (I'd interpret it as "use erase-fs-type list and set the selected one to the configured default, rather than the first in the list")

To land this before 3.3 (which is scheduled "sometime, eventually") you'll need to address backwards-compatibility like that. For 3.3 it's possible to "just" redefine the config meanings.

adriaandegroot:
Open conversation topic: **documentation**

The config file should be extensively documented; there should be examples. If the order of a list matters, there should be comments explaining that. There is a `partition.schema.yaml` which is the schema file for the YAML configuration. This needs to be updated to match the new config keys.

adriaandegroot:
Open conversation topic: **partition schema**

The key *partitionLayout* in the config file defines .. a configuration to apply when erasing. Each entry has a *name* and a *filesystem*. The interation with *defaultFileSystemType* is already a bit odd in **that** scheme, since it completely ignores the default FS type. For good reason: it's specified explicitly. 

- So the whole drop-down-choose-FS makes sense only if there **isn't** a layout specified; the layout overrides the need to pick a FS. A distro that specifies a layout will not set any choices anyway, and it's not needed.
- **or** the layout can have a special *filesystem* value, say "default", that picks up .. the default! This can be implemented without the erase-choice code, even, so it would be testable with existing layouts and configurations. In the presence of a choosable FS on erase, that changes every item in the layout that is told to use the default FS.
- **or** there could be multiple layouts, one for each chosen FS on erase. How to specify this in the config is unclear; possibly a *partitionLayout[defaultfs]* kind of key would work, but then we need to store multiple layouts and pick the right one on erase.

arcolinuxz:
Can this be incorporated in the next Calamares @adriaandegroot  or does it need further testing?
Always ready to be a betatester.

LordTermor:
> Can this be incorporated in the next Calamares @adriaandegroot or does it need further testing?
> Always ready to be a betatester.

Don't have time to finish it properly unfortunately

adriaandegroot:
This is now in branch *pr-1667* in the main repo, where I intend to resolve the issues that are still open in comments here.

LordTermor:
> This is now in branch _pr-1667_ in the main repo, where I intend to resolve the issues that are still open in comments here.

Thank you a lot for taking it in your hands. PR can be closed I think?

adriaandegroot:
> PR can be closed I think?

Let's leave it open until the branch is merged to *calamares*, otherwise I'll forget there's things pending (like the comments in this PR).

src/modules/partition/core/Config.cpp
adriaandegroot: I've decided to canonicalize (e.g. use the names that KPMCore uses internally).

adriaandegroot:
(All the code-style and minor technical nits have been addressed in *pr-1667*, leaving the three "open conversation topics". I'm not going to get to them today, there's the mysterious-partition-flags issue to chase.

adriaandegroot:
Topics **existing configuration** and **documentation** have been dealt with in the branch.

adriaandegroot:
I have chosen to extend the partition-layout code -- from @gportay -- such that partitions may be specified as *unknown* and that that means "use the default / user selection", e.g. defers the **actual** choice of FS until later.

PR #1680, opened 2021-04-22
Add Luet PackageManager support

jcfrosty:
Sabayon is being rebuilt into MocaccinoOS with a new packagemanager.

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -255,6 +255,22 @@ class PMEntropy(PackageManager):
         # Doesn't need to update the system explicitly
         pass
 
+    class PMLuet(PackageManager):
+    backend = "luet"
+
+    def install(self, pkgs, from_local=False):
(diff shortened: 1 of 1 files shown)


src/modules/packages/main.py
```diff
@@ -255,6 +255,22 @@ def update_system(self):
         # Doesn't need to update the system explicitly
         pass
 
+    class PMLuet(PackageManager):
```
adriaandegroot, line 258: That's an indentation error and the package manager won't be found.

adriaandegroot:
Good to see you wrestled through the package-manager lack-of-docs (I'll leave that issue open, though, it probably needs a few lines in the wiki somewhere). It'd be good if you also edited the default / example `packages.conf` file, to mention this package manager (you need to set *backend: luet*) and that Sabayon's *entropy* is deprecated.

adriaandegroot:
n/m, as I was going over the docs anyway I added the requested strings

PR #1684, opened 2021-05-01
Fix minor Typo on README

eamanu:
a `)` is missing.


adriaandegroot:
It's Markdown, so the notation `[thing](URL)` is important for making the link. You're right that there is a `)` missing, but it should go **after** the URL-part, so the parenthetical work-with-link becomes `([thing](URL))`. If you fix that up, it'd be grand.

eamanu:
Oh! so sorry!  I've just make the change without thinking. @adriaandegroot  thanks!

Issue #1630, opened 2021-01-31
Can not find DNS in WiFi connections after installation via Calamares due to copied /etc/resolv.conf


embar-:
Since https://github.com/calamares/calamares/commit/a45f78e9ac01c93cb4c178722fda54697e0fe050#diff-5960a6e192d1804ed3e3b84905264cff37275564879d6d87d0668369b495c21b Calamares writes `/etc/resolv.conf` file. However at least in _openSUSE_ it is just symlink into `/var/run/netconfig/resolv.conf` . Later after installation, as this file already exist, openSUSE Linux fails to properly open any regular internet address, i.e. it shows errors "could not resolve host", due to wrong DNS as it came hardcoded during installation.
Some distribution, like Debian, already intentionally disabled this Calamares functionality: 
https://git.parrotsec.org/packages/installer/calamares/-/blob/master/debian/patches/disable-resolvconf-creation.patch 
I recommend to remove lines about copying of /etc/resolv.conf file upstream.

embar-:
Or, if we need internet in chroot, please at least dont't forget to remove /etc/resolv.conf at the end of installation

kkofler:
Basically, there are 3 ways `/etc/resolv.conf` can be handled:
* manually written (or generated by the installer and potentially manually edited),
* automatically generated by NetworkManager,
* handled by `systemd-resolved`.

We need to make sure that we do the right thing in all cases. If `/etc/resolv.conf` is a symlink, overwriting it with a regular file is not a good idea.

kkofler:
There may be additional ways (which Calamares also needs to work with), but the 3 above are those that can easily be used on Fedora, the distro I am running. See also: https://fedoraproject.org/wiki/Changes/systemd-resolved

kkofler:
My `/etc/resolv.conf` is currently managed by NetworkManager, and it is a symlink to `/var/run/NetworkManager/resolv.conf`.

adriaandegroot:
The code has been there since Calamares 3.1, so it's a bit weird to have issues with it now (or I wish Debian would have reported it in a meaningful way). There's not much explanation *why* in the original commit. To me the whole code looks like something that assumes that the ISO (host system)  configures networking in such a way that that also applies in the target system.

But .. now I'm wondering why copying NetworkManager configurations is needed (e.g. in Debian) but the other bit is not.

Anyway, plan-of-attack would be:

[...]

PR #1692, opened 2021-05-12
copy resolvconf as link

raurodse:
Hi, maybe this patch will resolve this issue. https://github.com/calamares/calamares/issues/1630 
Basically copy /etc/resolv.conf as it is at the moment that you install to chroot.


adriaandegroot:
That looks sensible, and preserves absolute or relative symlinks as needed.

PR #1733, opened 2021-06-25
bootloader: add-support_for_aarch64

jiachao2130:
Install bootload on aarch64 machine

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -24,6 +24,8 @@ import os
 import shutil
 import subprocess
 
+import platform
+
 import libcalamares
 
 from libcalamares.utils import check_target_env_call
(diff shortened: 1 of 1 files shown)


jiachao2130:
In Debian, the /usr/bin/bootloader-config provides by calamares-settings-debian also need to modify to suit for aarch64, install grub-efi-arm64 for aarch64 or grub-efi-amd64 for x86_64
But I can't fork the project,,,,,  https://salsa.debian.org/live-team/calamares-settings-debian-packaging.git

adriaandegroot:
Thanks! I looked at the resulting code and then factored it out into its own little function, so that **if** there's even more variants to deal with, that won't cause deep indentation or complicated logic in the place where all we need is the id and the two file names.

PR #1734, opened 2021-06-26
[usersq] add inline messages, adjust layout

demmm:
warning messages implemented for user & host names
move root password option box inline with password section
adjust background color according to text.length
hardcode color instead of kirigami colors
usersq is now usable, sets user, hostname & password correctly on install
warning messages & color changes not fully done yet


demmm:
What usersq looks like now:
Screenshot_20210626_164857

PR #1738, opened 2021-07-06
usersq: use system colorscheme instead of hardcoded colors

milohr:
With these changes the built-in style for usersq module now picks the system color scheme, resulting in better visuals when using different color schemes, like dark ones.

### DIFF
--- a/src/modules/usersq/usersq.qml
+++ b/src/modules/usersq/usersq.qml
@@ -23,9 +23,6 @@ Kirigami.ScrollablePage {
     width: parent.width
     height: parent.height
 
-    Kirigami.Theme.backgroundColor: "#EFF0F1"
-    Kirigami.Theme.textColor: "#1F1F1F"
-
     header: Kirigami.Heading {
         Layout.fillWidth: true
         height: 50
@@ -57,8 +54,8 @@ Kirigami.ScrollablePage {
                 text: config.fullName
                 onTextChanged: config.setFullName(text)
 
-                palette.base: _userNameField.text.length ? "#f0fff0" : "#FBFBFB"
-                palette.highlight : _userNameField.text.length ? "#dcffdc" : "#FBFBFB"
+                palette.base: _userNameField.text.length ? Kirigami.Theme.backgroundColor : Kirigami.Theme.backgroundColor
+                palette.highlight : _userNameField.text.length ? Kirigami.Theme.positiveBackgroundColor : Kirigami.Theme.backgroundColor
             }
         }
 
@@ -89,8 +86,8 @@ Kirigami.ScrollablePage {
                 palette.base: _userLoginField.text.length
                     ? ( acceptableInput
                     ? ( _userLoginField.text === "root"
-                    ? "#ffdae0" : "#f0fff0" ) : "#ffdae0" ) : "#FBFBFB"
-                palette.highlight : _userLoginField.text.length ? "#dcffdc" : "#FBFBFB"
+                    ? Kirigami.Theme.negativeBackgroundColor : Kirigami.Theme.backgroundColor ) : Kirigami.Theme.negativeBackgroundColor ) : Kirigami.Theme.backgroundColor
+                palette.highlight : _userLoginField.text.length ? Kirigami.Theme.positiveBackgroundColor : Kirigami.Theme.backgroundColor
             }
 
             Label {
@@ -98,7 +95,6 @@ Kirigami.ScrollablePage {
                 text: qsTr("If more than one person will use this computer, you can create multiple accounts after installation.")
                 font.weight: Font.Thin
                 font.pointSize: 8
-                color: "#6D6D6D"
             }
         }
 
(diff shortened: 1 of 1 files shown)


demmm:
The reason for switching to hardcoded colors in https://github.com/calamares/calamares/commit/989d08c70aae9496788842143b1550379c281b64#diff-2170da30dfbbd8f3b4b3cc6cbfe7e981e5fa440521f6d530cc9c2fc1cacf3494 was twofold.
One, many distro's talking about adapting QML modules mention unfamiliarity with QML. To lower the threshold to use those modules, nothing is simpler then starting with adjusting colors after copying the available .qml file to the distro specific branding directory. Seeing just a numeric, short color code once again lowers the bar to read QML & adjust.
Second, kirgami colors & icons are often not used in testing environments (you need to have a proper theme setup in the sudo env), thus again making it much harder to see how to adjust QML

Big disadvantage of course is that it doesn't adapt to light or dark colored themes.

My preference is clear, otherwise I wouldn't have changed the settings in the mentioned commit, but I will let others decide/give their opinion too.

Regarding the pull request itself, see the rest of users.qml with line breaks for long variables, best to keep consistency there, and improve readability. Example https://github.com/calamares/calamares/blob/calamares/src/modules/usersq/usersq.qml#L89

milohr:
Got it.

Strit:
If usersq.qml uses hard-coded colors, why don't the rest of the qml modules do that?
Seems inconsistent to me.

demmm:
@Strit 
>why don't the rest of the qml modules do that?
Seems inconsistent to me.

As you can see from the commit, it is a recent change. I wanted to see the response to it (if any).  Should the response be positive, then gradually convert all QML colors, if negative, just one to revert.

adriaandegroot:
The QML shipped with upstream Calamares should be "ok". Something easy to adjust for a distro to their layout / colors / whatever, without the distro having an experienced QML designer on board. Downstreams with more ambition, like Nitrux, will be replacing the QML anyway with a complete branding module (that's why the QML modules all look in branding first).

There are two sort-of-independent aims here, which slightly conflict:
- the QML shipped with upstream Calamares should be educational and easy to adapt
- the QML should look ok in whatever environment

Why not both, though?

I'd like to suggest factoring out colors in a meaningful way (although I don't know if QML allows that); write an additional QML file that can be `import`ed by the QML files that are shipped with Calamares. Ship that additional thing **once**, e.g. as part of branding. Explain the options -- hardcoded colors, pulling colors from `branding.conf` (won't cover all the needed cases), using Kirigami theme -- in comments, show possible implementations.

Then bind useful names to whatever colors, and use those instead. The way I imagine this would look like in the end would be something like

```
palette.base: _userNameField.text.length ? calamares.colors.field_problem : calamares.colors.field_ok
```

Doing that consistently through the codebase would allow Camilo to pick "follow the theme" by picking the right bit of the colors-configuration file, while Alice and Bennie use the default hard-coded colors.

To do such a thing (again, if it is even possible), though would need:
- [ ] PR to add a colors-configuration file to the default Calamares branding
- [ ] PR to refactor colors from existing upstream QML code to use that colors-configuration
- [ ] closing this PR as "implemented differently"

demmm:
>Explain the options -- hardcoded colors, pulling colors from branding.conf (won't cover all the needed cases), using Kirigami theme -- in comments, show possible implementations.

To me that option makes the most sense, adding comments on hardcoded colors, not adding extra files (with fairly long coded names for colors).

adriaandegroot:
I think my main concern in the change to hard-coded colors scattered all over the code base is how to obtain consistency. That's a strength of Calamares branding (for QWidgets, anyway), there's one central place to twiddle colors. Having hard-coded colors scattered all over the QML codebase does not seem like an improvement to me -- for learning how to tweak QML I suppose it's ok, but then we get into the tension between the QML shipped with Calamares being a learning experience, and being the QML you'd want to put into production. 

All that said, I'd be much happier as well if we had an alternative-QML branding (e.g. other implementations of the QML for various modules) in the *calamares-extensions* repository, as a demonstraton of "you can ship unmodified upstream Calamares, plus a branding package to re-skin it quite thoroughly".

demmm:
>Having hard-coded colors scattered all over the QML codebase does not seem like an improvement to me

Makes sense.  So, for now, it would be best to accept this PR, then there is consistency again across all QML modules, and work on alternative-QML branding after.

adriaandegroot:
After the merge, the following issues show up:

- https://github.com/calamares/calamares/commit/e361f524156eb8994edf0a59518fb5269c5a791f#diff-c98e4ac58f0a01b32706e67226afd1a02d5b7dfe1a1dc5cf1f1db9a02f76d12cL60 text-replace error, there were two different colors, now the `?:` selects two identical colors
- now depends heavily on having a complete Kirigami theme in the environment; any missing colors will fall back to Kirigami defaults from the Material theme, https://api.kde.org/frameworks/kirigami/html/Material_2Theme_8qml_source.html

For a default-QML that has both an educational value and should look "ok", these are problematic. We'll back this out and re-work the way colors are named.

PR #1749, opened 2021-08-11
[partition] Disable delete button for extended partition

LordTermor:
Extended partition can't be removed when contains children and Calamares allows this, what causes installer crash. This PR
adds missing check.


adriaandegroot:
Looks good, thanks! There's some code-cleanups that I now see would be nice to apply, I'll do that next.

Issue #1754, opened 2021-08-16
Networkcfg copies live user with file copy


demmm:
> Hi! Thank you for helping improve Calamares. If you are seeing a problem in installing a specific distribution, you should **probably** report the problem in the distribution's bug tracker, first. That helps filter out issues with packaging, mis-configuration, etc. that Calamares has no control over. If you are a distribution packager or maintainer, this page is for you.

**Describe the bug**
As reported a few times over the years (IRC, telegram), the networkcfg module is not functional, file is copied when present, but on restarting newly installed system, new wifi credentials need to be added.

Issue is now found, not fixed yet in the module.
The .nmconection file that is copied contains a line for the user, thus it contains whatever name a distro uses for the live system, example KaOS line:
```
permissions=user:live:;
```
When you edit that file (as root) and set it to current user of the installed system, wifi works right away on rebooting into the newly installed system.
networkcfg/main.py will need some lines added to copy the correct target user.

philmmanjaro:
In other words: we have to replace those lines with `permissions=user:$loginName:;` of which `$loginName` gets replaced by the actual target user.

demmm:
Start of a fix:
```
    root_mount_point = libcalamares.globalstorage.value("rootMountPoint")
    user = libcalamares.globalstorage.value("username")

    if root_mount_point is None:
        libcalamares.utils.warning("rootMountPoint is empty, {!s}".format(root_mount_point))
        return (_("Configuration Error"),
                _("No root mount point is given for <pre>{!s}</pre> to use." ).format("networkcfg"))

    source_nm = "/etc/NetworkManager/system-connections/"
    target_nm = os.path.join(
        root_mount_point, "etc/NetworkManager/system-connections/"
        )

    # Sanity checks.  We don't want to do anything if a network
    # configuration already exists on the target
    if os.path.exists(source_nm) and os.path.exists(target_nm):
        for network in os.listdir(source_nm):
            # Skip LTSP live
            if network == "LTSP":
                continue

            source_network = os.path.join(source_nm, network)
            target_network = os.path.join(target_nm, network)

            if os.path.exists(target_network):
                continue

            try:
                shutil.copy(source_network, target_network, follow_symlinks=False)
                if 'live' in open(target_network).read():
                    text = []
                    with open(target_network, "r") as network_conf:
                        text = network_conf.readlines()
                        with open(target_network, "w") as network_conf:
                            for line in text:
                                if 'permissions=user:live:;' in line:
[...]

PR #1755, opened 2021-08-16
[networkcfg] add setting the correct target user in copied file

demmm:
see https://github.com/calamares/calamares/issues/1754

### DIFF
--- a/src/modules/networkcfg/main.py
+++ b/src/modules/networkcfg/main.py
@@ -7,6 +7,7 @@
 #   SPDX-FileCopyrightText: 2014 Teo Mrnjavac <teo@kde.org>
 #   SPDX-FileCopyrightText: 2017 Alf Gaida <agaida@siduction.org>
 #   SPDX-FileCopyrightText: 2019 Adriaan de Groot <groot@kde.org>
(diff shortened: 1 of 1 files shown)


hphilm:
We also could check for the `permissions` line in the files and replace it regardless, however this seems to work also.

demmm:
AFAIK, `permissions` is always there, plus it shouldn't be replaced, patch only works if the live user is listed in the file, thus that one needs replacing.
But I'm redoing an ISO (plan was to release today, but will postpone now with this possible fix in place), so lots of different options will be tested.

hphilm:
All good. keep us updated.

demmm:
Doesn't work as is, no crashes or install fail, but live user is not detected right, will need a different option for `live_user = getpass.getuser()`.  That one returns `root` when Cala runs.

saber716rus:
When will it be implemented? I saw only "FS label" in manjar, but not the mount option.

demmm:
What does mount option have to do with network settings being preserved on the target system?  Looks llke you commented on a wrong ticket.

saber716rus:
So you yourself tied this comit to my topic.

PR #1770, opened 2021-09-03
Dont set fsck check for BTRFS in fstab

dr460nf1r3:


adriaandegroot:
There's an opportunity to clean this up, factor out the calculate-the-check-number and in general, massage this module to have more readable code-flow. But that can be done some other time.

PR #1774, opened 2021-09-06
[keyboardq] add interactive keyboard preview

demmm:
rewrite of keyboardq.qml, reduce stackview to 2, use a combobox for
keyboard models list
colors set to configurable
.xml files used for keyboard layouts, about a dozen added now
builds, runs, actions record as intended, GS filled correctly


demmm:
Screenshot_20210906_130055

PR #1775, opened 2021-09-07
make sure locale and keyboard layout is set to us for india

ghost:
india is a large country, it has many languages in it.
but us english is used in most of computer and mobile devices.

### DIFF
--- a/src/modules/locale/LocaleConfiguration.cpp
+++ b/src/modules/locale/LocaleConfiguration.cpp
@@ -201,6 +201,8 @@ LocaleConfiguration::fromLanguageAndLocation( const QString& languageLocale,
                 // but nearly all its native speakers also speak English,
                 // and migrants are likely to use English.
                 { "IE", "en" },
+                //India has many languages even though hindi is known as national language but english is used in all computer,mobile devices.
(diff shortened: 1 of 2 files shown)


src/modules/locale/LocaleConfiguration.cpp
```diff
@@ -201,6 +201,8 @@ LocaleConfiguration::fromLanguageAndLocation( const QString& languageLocale,
                 // but nearly all its native speakers also speak English,
                 // and migrants are likely to use English.
                 { "IE", "en" },
+                //India has many languages even though hindi is known as national language but english is used in all computer,mobile devices.
+                { "IN", "en" },
```
adriaandegroot, line 205: Is this needed? 0e7982ec371246cc6e189ada32b2787f204479e5 already switched off Hindi-for-India AFAIK.

adriaandegroot: (and if it is needed -- you can test more easily than I can because I need to spoof locations -- then the comment should say "Hindi" and "English" etc.

adriaandegroot:
Merged by hand with textual fixes applied to the code comments.

PR #1776, opened 2021-09-07
[initcpiocfg] add consolefont to hooks

ghost:
consolefont needs to be added in hooks array for better console fonts in tty
https://wiki.archlinux.org/title/Linux_console#Persistent_configuration

### DIFF
--- a/src/modules/initcpiocfg/main.py
+++ b/src/modules/initcpiocfg/main.py
@@ -144,7 +144,7 @@ def find_initcpio_features(partitions, root_mount_point):
 
     :return 3-tuple of lists
     """
-    hooks = ["base", "udev", "autodetect", "modconf", "block", "keyboard", "keymap"]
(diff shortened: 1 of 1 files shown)


src/modules/initcpiocfg/main.py
```diff
@@ -115,7 +115,7 @@ def modify_mkinitcpio_conf(partitions, root_mount_point):
     btrfs = ""
     lvm2 = ""
     hooks = ["base", "udev", "autodetect", "modconf", "block", "keyboard",
-             "keymap"]
+             "keymap","consolefont"]
```
adriaandegroot: Does the consolefont hook **always** exist? Or is this something your distro has (maybe all Arches?) but others do not? It would need a detection mechanism (hooks in general, might be nice to check that they exist before enabling them).

adriaandegroot: Python style: space after `,`

adriaandegroot (commented):
Please squash this down to a single commit, without the 3-commits-and-a-revert history, and then force-push to the PR.

PR #1777, opened 2021-09-07
[fstab] Improve btrfs mount options

ghost:
this pr adds ssd optimization + asynchronous discard support and
makes zstd the default compression algorithm for btrfs

also adds additional subvolumes

### DIFF
--- a/src/modules/fstab/fstab.conf
+++ b/src/modules/fstab/fstab.conf
@@ -13,7 +13,7 @@
 # options from this mapping.
 mountOptions:
     default: defaults,noatime
-    btrfs: defaults,noatime,space_cache,autodefrag
+    btrfs: defaults,noatime,space_cache,autodefrag,compress=zstd
 
 # Mount options to use for the EFI System Partition. If not defined, the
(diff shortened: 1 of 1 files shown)


src/modules/mount/mount.conf
```diff
@@ -50,7 +50,13 @@ btrfsSubvolumes:
...
+    - mountPoint: /srv
+      subvolume: /@srv
     - mountPoint: /var/cache
       subvolume: /@cache
     - mountPoint: /var/log
-      subvolume: /@log
\ No newline at end of file
+      subvolume: /@log
+    - mountPoint: /var/tmp
+      subvolume: /@tmp
```
adriaandegroot: This is an example file, not particularly intended for use by actual distro's (if you're a distro, and have specific needs / a style for btrfs subvolumes, you should have a `mount.conf` for your distro). In particular `/srv` seems like a distro-specific thing.

adriaandegroot (changes requested):
Please squash this down to remove revert-commits and submit one thing at a time: this would be `[fstab] Improve btrfs mount options` because I don't think the *mount* module changes should be included.

PR #1783, opened 2021-09-13
[packagechooserq] start using prettyName

demmm:
this way it can be used for more instances

### DIFF
--- a/src/modules/packagechooserq/PackageChooserQmlViewStep.cpp
+++ b/src/modules/packagechooserq/PackageChooserQmlViewStep.cpp
@@ -29,7 +29,7 @@ PackageChooserQmlViewStep::PackageChooserQmlViewStep( QObject* parent )
 QString
 PackageChooserQmlViewStep::prettyName() const
 {
-    return tr( "Packages" );
+    return m_prettyName;
 }
 
 QString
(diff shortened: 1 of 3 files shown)


demmm:
Image showing the use of prettyName, 2 packagechooserq instances, packages & bootloader:
Screenshot_20210913_204517

adriaandegroot:
There is a `CalamaresUtils::Locale::TranslatedString` for this purpose; because the *prettyName* is human-readable, it should be translatable (e.g. "Bootloader" is fine in English, but it should *cough* be "Laarslader" in Dutch) so that it follows settings when the language is selected in the Welcome page.

See e.g. the *netinstall* module, where the Config object loads the sidebar label and top-text from the configuration file, and it is translated too. The *packagechooser* module does something similar. That should just move to the Config object, rather than living in the ViewStep; then it's shared between both versions, and they can both use the *labels* key to set sub-key *step* (and *step[nl]*) with relevant strings.

adriaandegroot:
.. so, while this specific code is OK, it could be better -- more flexible, resilient for translations -- both for this module and *packagechooser*.

demmm:
Best to close this one then, and get it (at whatever convenient time) from packagechooser.

demmm:
No need to move it to config in packagechooser, I have the translation working locally, and no-one else uses packagechoosrq, let alone extra instances, there is no need for prettyName in main Calamares.

Issue #1786, opened 2021-09-19
luksbootkeyfile job "cryptsetup luksAddKey" fails, timeout too short


whorfin:
This was observed attempting to install lubuntu 20.04.3 LTS, but as we shall see, I think it might be fundamental.

What was observed:
Ticking the "Encrypt" box for a "erase and install from scratch" installation eventually resulted in the popup:
```
Error
Installation Failed
Encrypted rootfs setup error
Could not configure LUKS key file on partition /dev/sda1.
```

After the installer bailed, I found that `/crypto_keyfile.bin` was in fact created
And I could manually (with a chroot)
```
# cryptsetup luksAddKey /dev/sda1 /crypto_keyfile.bin
```
That all worked just fine, although I found it took 8-12 seconds on an older computer with a 500G "spinning rust" hard-drive.

So what's the problem?

Running `calamares -d` again to see what was happening, I saw that yes indeed the creation of the keyfile worked fine, but the `cryptsetup` saw this:
```
Running "chroot" ("/tmp/calamares-root-3alg6xxq", "cryptsetup", "luksAddKey", "/dev/sda1", "/crypto_keyfile.bin")
16:59:01 [2]: WARNING: Timed out.  Output so far:

16:59:01 [2]: QProcess: Destroyed while process ("chroot") is still running.
16:59:01 [2]: WARNING: Could not configure LUKS keyfile on "/dev/sda1" : "" ( exit code -4 )
```

Having a look in https://github.com/calamares/calamares/blob/calamares/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp we see on line 122 that the `cryptsetup` is only given 15 seconds.  On my not-terribly-slow test computer, I regularly saw this take 8-12 seconds - so I speculate that's simply not enough time.  My tested range of 8-12 seconds is far too close to 15, and likely with the system in a "hot" state after the `gen-locale` step, we cross the threshold to failtown.  This should be at least 30 seconds, maybe a minute or more - there's no harm in waiting that long (or longer) for such an important operation, before blowing up and leaving the drive useless with no continuation mechanism.

I ran this for a fourth time and on the fourth try, the `cryptsetup` completed just barely in time, and everything else succeeded.

So, please increase this timeout.

adriaandegroot:
Thanks for the carefully-reasoned and explained report. I've bumped the timeout (like you did in the PR) and tidied up some attendant code.

PR #1787, opened 2021-09-19
Allow more conservative timeout in luksbootkeyfile for cryptsetup to address #1786

whorfin:
Fix for #1786


adriaandegroot:
Darn. I hadn't seen immediately that there was a PR for this, so I wrote my own -- taking advantage to tidy up some bits-and-bobs in the surrounding code. So I'm closing this one without merging, but the effect is there.

PR #1808, opened 2021-10-23
Fix for zfs partition deletion

dalto8:
kpmcore uses zfs to destroy the zpool on a zfs partition.

In the case of Calamares, most ISOs don't include zfs or the needed kernel modules.  This means that when Calamares tries to delete an existing zfs partition, it results in error.

This change checks to see if there is working zfs environment.  If there isn't, it uses `sfdisk` to manually delete the partition.  `sfdisk` was chosen since it is used by kpmcore and shouldn't add any additional dependencies.

### DIFF
--- a/src/modules/partition/jobs/DeletePartitionJob.cpp
+++ b/src/modules/partition/jobs/DeletePartitionJob.cpp
@@ -20,6 +20,44 @@
 #include <kpmcore/ops/deleteoperation.h>
 #include <kpmcore/util/report.h>
 
+#include <QCoreApplication>
+
+/** @brief Determine if the given partition is of type Zfs
+ *
+ * Returns true if @p partition is of type Zfs
+ *
+ */
+static bool
+isZfs( Partition* partition )
+{
+    return partition->fileSystem().type() == FileSystem::Type::Zfs;
+}
+
+/** @brief Remove the given partition manually
+ *
+ * Uses sfdisk to remove @p partition.  This should only be used in cases
+ * where using kpmcore to remove the partition would not be appropriate
+ *
+ */
+static Calamares::JobResult
+removePartition( Partition* partition )
+{
+    auto r = CalamaresUtils::System::instance()->runCommand(
+        { "sfdisk", "--delete", "--force", partition->devicePath(), QString::number( partition->number() ) },
+        std::chrono::seconds( 5 ) );
+    if ( r.getExitCode() !=0 || r.getOutput().contains("failed") )
+    {
+        return Calamares::JobResult::error(
+            QCoreApplication::translate( DeletePartitionJob::staticMetaObject.className(), "Deletion Failed" ),
+            QCoreApplication::translate( DeletePartitionJob::staticMetaObject.className(),
+                                         "Failed to delete the partition with output: " )
+                + r.getOutput() );
+    }
+    else
+    {
+        return Calamares::JobResult::ok();
+    }
+}
 
 DeletePartitionJob::DeletePartitionJob( Device* device, Partition* partition )
     : PartitionJob( partition )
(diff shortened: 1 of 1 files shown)


kkofler:
I think this belongs into kpmcore, not Calamares.

dalto8:
> I think this belongs into kpmcore, not Calamares.

It is an interesting question.  Should kpmcore check for zfs presence or should it rely on anyone leveraging the functionality to implement it if they call it?  I am not sure what their position on that is.

Either way, Calamares users will continue to have failed installs in the meantime.

kkofler:
@stikonas: What do you think?

(FYI, @stikonas is the kpmcore maintainer.)

src/modules/partition/jobs/DeletePartitionJob.cpp
```diff
@@ -49,6 +51,29 @@ DeletePartitionJob::prettyStatusMessage() const
 Calamares::JobResult
 DeletePartitionJob::exec()
 {
+    // Special handling for zfs
```
adriaandegroot: Typically I'd split this off into a separate function -- probably a static free function -- so that the overall flow is easier to read; something along the lines of

```
if (isZfs(m_partition)) { return removeZfs(m_partition); }
```

Then bung the rest of this code into those functions.

src/modules/partition/jobs/DeletePartitionJob.cpp
```diff
@@ -49,6 +51,29 @@ DeletePartitionJob::prettyStatusMessage() const
...
+
+        if ( r.getExitCode() != 0 )
+        {
+            r = CalamaresUtils::System::instance()->runCommand( { "sfdisk",
+                                                                  "--delete",
+                                                                  "--force",
+                                                                  m_partition->devicePath(),
+                                                                  QString::number( m_partition->number() ) },
+                                                                std::chrono::seconds( 5 ) );
+            if ( r.getExitCode() != 0 )
```
adriaandegroot: Please use `{ }` around every block (Calamares **has** had bugs because of suggestive indentation and missing braces, and it's made worse by mixing Python and C++ where "suggestive indentation" is "meaningful indentation" half the time).

src/modules/partition/jobs/DeletePartitionJob.cpp
```diff
@@ -49,6 +51,29 @@ DeletePartitionJob::prettyStatusMessage() const
...
+                return Calamares::JobResult::error( "message",
```
adriaandegroot: These are user-visible, so you need to apply `QObject::tr()` to them (if you move this code to a static free function, it becomes a bit wordy, as

```
QCoreApplication::translate( DeletePartitionJob::staticMetaObject.className(), "message" )
```

.. not that I think "message" is a good title for an error message describing the failure to delete a partition.

adriaandegroot (commented):
I share the feeling that this belongs in kpmcore -- you ought to be able to delete a partition even if the FS on that partition is unsupported .. um, hang on. This is weird **anyway** because a partition would typically belong to a pool, not to a single ZFS filesystem. OTOH, it'd be nice to have delete-ZFS-partition working even before a new kpmcore release.

There's comments on the code, though, regardless of whether this belongs here or not.

stikonas:
> @stikonas: What do you think?
> 
> (FYI, @stikonas is the kpmcore maintainer.)

Yes, I think it would be good to get this into KPMcore. KPMcore releases are fairly frequent these days. Next release is in December and feature freeze is 12 November.

dalto8:
> > @stikonas: What do you think?
> > (FYI, @stikonas is the kpmcore maintainer.)
> 
> Yes, I think it would be good to get this into KPMcore. 

Probably not the right place to have this discussion, but I don't think kpmcore should need to do anything this complicated.  Since kpmcore can't understand intention, it should alwyas just delete the partition and not worry about the fact that it is a zfs partition.  What it currently does, forcibly destroying a zpool based on the partition label is unbelievably dangerous.  What if someone is trying to delete a partition that is part of a mirror or raid set and not expecting the entire pool to be destroyed?  Worse, what if the partition is one that has been removed from the pool and still has the label of the pool?

I am not that familiar with kpmcore but I guess that would be just removing the `remove()` function from the zfs implementation and letting the base class handle the remove?

stikonas:
> > > @stikonas: What do you think?
> > > (FYI, @stikonas is the kpmcore maintainer.)
> > 
> > 
> > Yes, I think it would be good to get this into KPMcore.
> 
> Probably not the right place to have this discussion, but I don't think kpmcore should need to do anything this complicated. Since kpmcore can't understand intention, it should alwyas just delete the partition and not worry about the fact that it is a zfs partition. What it currently does, forcibly destroying a zpool based on the partition label is unbelievably dangerous. What if someone is trying to delete a partition that is part of a mirror or raid set and not expecting the entire pool to be destroyed? Worse, what if the partition is one that has been removed from the pool and still has the label of the pool?
> 
> I am not that familiar with kpmcore but I guess that would be just removing the `remove()` function from the zfs implementation and letting the base class handle the remove?

I don't remember why exactly there is zpool destruction. But It looks like it is indeed too much. But maybe there was an issue that if zpool if active then `remove()` failed. But I am not really familiar with ZFS, so I might be saying nonsense.

adriaandegroot:
This discussion suggests that what Calamares should do is **ignore** the current KPMCore implementation and always use the `sfdisk remove` approach here (e.g. we can lose the "zpool status" check). That will damage the pool, but avoid whatever thing kpmcore is doing with the pool as a whole.

dalto8:
Let me know if this is more inline with what you were thinking.

stikonas:
And `kpmcore` is now switched to just remove partition without messing up with pools.

PR #1809, opened 2021-10-23
Add isModuleEnabled function

dalto8:
This is a simple function to make checking if a modules is enabled in settings.conf easier.

### DIFF
--- a/src/libcalamares/Settings.cpp
+++ b/src/libcalamares/Settings.cpp
@@ -275,7 +275,7 @@ Settings::isModuleEnabled( const QString& module )
     // Iterate over the list of modules searching for a match
     for ( const auto& moduleInstance : qAsConst( m_moduleInstances ) )
     {
-        if ( moduleInstance.key().first == module )
+        if ( moduleInstance.key().module() == module )
         {
             return true;
         }
--- a/src/libcalamares/Settings.cpp
+++ b/src/libcalamares/Settings.cpp
@@ -269,14 +269,16 @@ Settings::Settings( const QString& settingsFilePath, bool debugMode )
     s_instance = this;
 }
 
-bool
-Settings::isModuleEnabled( QString module )
+const bool
+Settings::isModuleEnabled( const QString& module )
 {
     // Iterate over the list of modules searching for a match
-    for ( const auto& moduleInstance : Calamares::Settings::instance()->moduleInstances() )
+    for ( const auto& moduleInstance : qAsConst( m_moduleInstances ) )
     {
         if ( moduleInstance.key().first == module )
+        {
             return true;
+        }
     }
 
     return false;
@@ -288,7 +290,7 @@ Settings::reconcileInstancesAndSequence()
     // Since moduleFinder captures targetKey by reference, we can
     //   update targetKey to change what the finder lambda looks for.
     Calamares::ModuleSystem::InstanceKey targetKey;
-    auto moduleFinder = [ &targetKey ]( const InstanceDescription& d ) { return d.isValid() && d.key() == targetKey; };
+    auto moduleFinder = [&targetKey]( const InstanceDescription& d ) { return d.isValid() && d.key() == targetKey; };
 
     // Check the sequence against the existing instances (which so far are only custom)
     for ( const auto& step : m_modulesSequence )
--- a/src/libcalamares/Settings.cpp
+++ b/src/libcalamares/Settings.cpp
@@ -269,13 +269,26 @@ Settings::Settings( const QString& settingsFilePath, bool debugMode )
     s_instance = this;
 }
 
+bool
+Settings::isModuleEnabled( QString module )
+{
+    // Iterate over the list of modules searching for a match
+    for ( const auto& moduleInstance : Calamares::Settings::instance()->moduleInstances() )
+    {
+        if ( moduleInstance.key().first == module )
+            return true;
+    }
+
+    return false;
+}
+
 void
 Settings::reconcileInstancesAndSequence()
 {
     // Since moduleFinder captures targetKey by reference, we can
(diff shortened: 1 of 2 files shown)


src/libcalamares/Settings.h
```diff
@@ -152,8 +152,12 @@ class DLLEXPORT Settings : public QObject
      */
     bool isSetupMode() const { return m_isSetupMode; }
 
+    /** @brief Returns true if the named module is enabled */
+    bool isModuleEnabled( QString module );
```
adriaandegroot: - `const QString& module` is idiomatic (even if QString is cheap-to-copy, the reference thing is deeply ingrained)
- method should be `const`
- "Returns whether the named module is enabled" for brief
- Describe that this means "any instance of @p module is listed in the sequence (in a show or an exec part, that does not matter) in settings", make clear that this isn't the same as "@p module is present on the system and could conceivably be used"

src/libcalamares/Settings.cpp
```diff
@@ -269,13 +269,26 @@ Settings::Settings( const QString& settingsFilePath, bool debugMode )
...
+}
+
 void
 Settings::reconcileInstancesAndSequence()
 {
     // Since moduleFinder captures targetKey by reference, we can
     //   update targetKey to change what the finder lambda looks for.
     Calamares::ModuleSystem::InstanceKey targetKey;
-    auto moduleFinder = [&targetKey]( const InstanceDescription& d ) { return d.isValid() && d.key() == targetKey; };
+    auto moduleFinder = [ &targetKey ]( const InstanceDescription& d ) { return d.isValid() && d.key() == targetKey; };
```
adriaandegroot: This is a spurious change. Calamares uses clang-format 8 or clang-format 9 **only**. There are some really annoying changes between clang-format 9 and 10, so that it's not possible to keep a "clean" formatting while using both versions. You end up with endless ping-ponging of spaces in lambdas.

(That said, I am inclined to revise the coding style for Calamares 3.3 so we can use newer versions of the tooling, and have less custom-configuration involved)

src/libcalamares/Settings.cpp
```diff
@@ -269,13 +269,26 @@ Settings::Settings( const QString& settingsFilePath, bool debugMode )
     s_instance = this;
 }
 
+bool
+Settings::isModuleEnabled( QString module )
```
adriaandegroot: See comments on consts and references in the header file.

src/libcalamares/Settings.cpp
```diff
@@ -269,13 +269,26 @@ Settings::Settings( const QString& settingsFilePath, bool debugMode )
...
+{
+    // Iterate over the list of modules searching for a match
+    for ( const auto& moduleInstance : Calamares::Settings::instance()->moduleInstances() )
```
adriaandegroot: `isModuleEnabled()` is a member function, you don't have to re-direct through `instance()` (which might even be a **different** Settings object!). Just iterate over `moduleInstances()`, or better yet, `m_moduleInstances`. This looks like an artifact of copying a free function (which needs to get the list of instances in this fashion) into a member function.

src/libcalamares/Settings.cpp
```diff
@@ -269,13 +269,26 @@ Settings::Settings( const QString& settingsFilePath, bool debugMode )
...
+    {
+        if ( moduleInstance.key().first == module )
```
adriaandegroot: Please use `{ }` around all blocks (the style-tool should do that already, but astyle is sometimes a bit hard-of-learning).

You can use `moduleInstance.key().module() == module` to avoid the implementation knowledge that the module-key (e.g. *packagechooser*) is first.

src/libcalamares/Settings.h
adriaandegroot: This should probably emphasize that `@p module` is a module name, or module key (e.g. *packagechooser*) and not a full module key+id (e.g. *packagechooser@packagechooser* or *shellprocess@before_user*). It also doesn't consider QML and non-QML versions of a module "the same"; this is something that will eventually change when the Q-or-not-Q decision is not reflected in the module's identity, but in its configuration.

adriaandegroot (changes requested):
There's some const-bits and code-cleanup that needs doing, alongside some nits.

src/libcalamares/Settings.h
dalto8: > This should probably emphasize that `@p module` is a module name, or module key (e.g. _packagechooser_) and not a full module key+id (e.g. _packagechooser@packagechooser_ or _shellprocess@before_user_).

Although I don't need it in my implementation, would it be useful to add a bool to allow to check either case?  The module key or the key+id?

> It also doesn't consider QML and non-QML versions of a module "the same";

Should it?  If so, is the only difference a leading "q" in the name?

adriaandegroot: These were more "bear in mind that.." kinds of comments from me. I don't think that checking for *module@id* is useful; the *id* part serves to distinguish different instances of the same module, when you need it more than once (e.g. multiple netinstalls, or multiple shellprocesses, those are the most common cases). I can't usefully imagine a module checking for a specific instance of a different module -- at some point the problem becomes "fix your distro's configuration". (e.g. what would checking for *shellprocess@you_can_call_me_al* even **mean** in a general context?)

This is probably something that is up to the caller, not the function implementing the check; is a module satisfied if *packagechooser* is enabled? How about *packagechooserq*? Either? Both? That's up to the caller.

src/libcalamares/Settings.cpp
dalto8: Yep, I was testing it elsewhere and then copied here.  Fixed now.

dalto8:
This should be ready for review again.

PR #1814, opened 2021-10-28
feat(displaymanager): greetd support

boredland:
this PR aims at adding support for the light [greetd](https://sr.ht/~kennylevinsen/greetd/) login manager and some/most of its [greeters](https://wiki.archlinux.org/title/Greetd#Greeters) (the ones that I deemed easily configurable). I just discovered the great [testing guide](https://github.com/calamares/calamares/wiki/Test-Guide), so that I will try before this is ready for review - early feedback is highly welcome though.

### DIFF
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -844,13 +844,17 @@ class DMgreetd(DisplayManager):
         return os.path.join(self.root_mount_point, path)
 
     def config_path(self):
-        return self.os_path("etc/greetd/config.toml")
+        path = self.os_path("etc/greetd/config.toml")
+        if not os.path.exists(path):
+            open(path, 'a').close()
+        return path
 
     def environments_path(self):
         return self.os_path("etc/greetd/environments")
 
     def config_load(self):
         self.config_data = toml.loads(self.config_path())
+        return self.config_data
 
     def config_write(self):
         toml.dump(self.config_data, self.config_path())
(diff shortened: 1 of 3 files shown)


src/modules/displaymanager/main.py
```diff
@@ -17,7 +17,7 @@
 
 import abc
 import os
-import re
```
boredland, line 20: this is an unused import

src/modules/displaymanager/main.py
```diff
@@ -16,8 +16,9 @@
 #
 
 import abc
+from genericpath import exists
 import os
-import re
+import toml
```
hphilm, line 20: This might introduce a 3rd-party lib. Can this be handled also without the need of toml?

boredland: I think it could be possible, its just text after all. will it be similarly nice to interface with? I bet not.

adriaandegroot: I seem to have *toml* as something-I-have-installed-already on my developer machine, let's see if some random ALCI-based distro has it on the ISO image .. nope. But it's 140kB or so, so not a heavy dependency. You could bung the `import toml` in the methods of the class, and then document that enabling greetd will require `toml`.

src/modules/displaymanager/main.py
```diff
@@ -835,6 +835,84 @@ def greeter_setup(self):
...
+
+        if not os.path.exists(self.config_data['default_session']):
+            self.config_data['default_session'] = {}
+
+        self.config_data['default_session']['user'] = self.greeter_user
+
+        return self.config_data
+
+    def config_write(self):
+        toml.dump(self.config_data, self.config_path())
```
adriaandegroot, line 868: The version of toml I have wants a file-like object as second parameter, not just a string.

adriaandegroot (commented):
Overall I like this -- nice clean code. The extra dependency can be wrangled out-of-sight. Most of the other DM-configuration thingies do text-replacement, rather than reading-and-writing the config file -- that is more fragile in many ways, but doesn't need any extra code.

src/modules/displaymanager/main.py
boredland: according to the (api reference)[https://github.com/uiri/toml#api-reference]:

```
toml.dump(o, f, encoder=None)
o: An object to be converted into TOML
f: A File descriptor where the TOML-formatted output should be stored
```

boredland:
> that is more fragile in many ways

thanks for the feedback. I guess I could find a nice way to do this however. the overall complexity of those files is quite low. on the other hand: using a sane parser library for the three or four typical configuration languages (json, yaml, toml, ini?) could make all of this less cumbersome.

src/modules/displaymanager/main.py
adriaandegroot: That's what I said :) It needs a file-descriptor, not a path. You can check that with

```
>>> import toml
>>> toml.dump(dict(), "/tmp/x")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "python3.8/site-packages/toml-0.10.2-py3.8.egg/toml/encoder.py", line 27, in dump
    if not f.write:
AttributeError: 'str' object has no attribute 'write'
```

adriaandegroot:
If you can move the `import toml` into the methods for greetd, then we don't have an extra dependency unless the distro wants to support greetd; we can merge that and think about useful configuration-parsing-equipment later.

adriaandegroot:
Since there didn't seem to be much movement, I ended up merging this and applying the requested changes -- moving the import -- myself. Then it turned out that my earlier reviews had missed a whole lot of other issues, which have mostly been fixed now. I'm pretty sure the `desktop_environment_setup()` method is broken, since you can't write a desktop-environment object. I'm going to leave that for an issue report / PR that fixes up greetd support.

PR #1822, opened 2021-11-07
zfs support - Phase 1

dalto8:
I thought this was far enough along that it was worthy of sharing to ensure I am going in the right direction.

My goal with phase 1 was to bring it to feature parity with btrfs support.  The idea for phase 2 would add a UI to allow the user to manually configure the pools and datasets.

Here is what is done so far:
* Changes to partition to allow zfs partitions to be selected and created
* A new module for zfs where pools and datasets are configured and defined
* Support for zfs in the initcpiocfg, mount, umount, fstab, grubcfg and bootloader modules
* Supports zfs encryption

What still needs to be done for phase 1:
* <s>Support for encryption</s>
* <s>Add support for multiple zpools</s>
* <s>Export the pools in unmount</s>
* <s>Ensure zpools can't be re-used</s>
* <s>Much more testing, only the changes to partition, mount and and the zfs modules have been tested</s>


dalto8:
There are ~10 lines of duplicate code between `partition/gui/CreatePartitionDialog.cpp` and `partition/core/PartitionLayout.cpp`.  I am not sure what the easiest way to remove that duplication is.

src/modules/bootloader/main.py
```diff
@@ -92,6 +92,32 @@ def get_kernel_line(kernel_type):
...
+    zfs = libcalamares.globalstorage.value("zfs")
+
+    if not zfs:
+        libcalamares.utils.warning("Failed to locate zfs dataset list")
+        return None
+
+    # Find the root dataset
+    for dataset in zfs:
+        try:
+            if dataset['mountpoint'] == '/':
```
adriaandegroot: Inconsistent quotes (use `"`)

src/modules/bootloader/main.py
```diff
@@ -133,12 +159,21 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
-    # about the root subvolume.
     for partition in partitions:
+        # systemd-boot with a BTRFS root filesystem needs to be told
+        # about the root subvolume.
         if partition["mountPoint"] == "/" and partition["fs"] == "btrfs":
             kernel_params.append("rootflags=subvol=@")
 
+        # zfs needs to be told the location of the root dataset
+        if partition["mountPoint"] == "/" and partition["fs"] == "zfs":
+            zfs_root = get_zfs_root
```
adriaandegroot: Probably missing `()` to actually **call** the function?

src/modules/bootloader/main.py
```diff
@@ -133,12 +159,21 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+            if zfs_root is not None:
+                kernel_params.append("root=ZFS=" + zfs_root)
```
adriaandegroot: This makes me think a better name for the variable would be `zfs_root_path`

src/modules/bootloader/main.py
```diff
@@ -133,12 +159,21 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
...
+            else:
+                # Something is really broken if we get to this point
+                libcalamares.utils.error("Internal error handling zfs dataset")
```
adriaandegroot: If it's really broken, how about returning some kind of error (so the installation ends semi-cleanly) or raising an exception (less clean).

src/modules/bootloader/main.py
```diff
@@ -314,6 +349,37 @@ def get_grub_efi_parameters():
...
+    partitions = libcalamares.globalstorage.value("partitions")
+    if not partitions:
+        libcalamares.utils.error("Failed to run grub-mkconfig, no partitions defined in global storage")
+        return
+
+    # check for zfs
+    is_zfs = False
+    for partition in partitions:
+        if partition["mountPoint"] == "/" and partition["fs"] == "zfs":
+            is_zfs = True
```
adriaandegroot: There's possibly a more-Pythonic `is_zfs = any([partition["mountPoint"] == "/" and partition["fs"] == "zfs" for partition in partitions])` .. and since this particular check for root-filesystem-is-zfs shows up several times, maybe it should be a free function (e.g. `is_zfs_root(partition)` .. and that alongside a `is_btrfs_root(partition)` for symmetry). Your call, it just strikes me as a duplicated condition that can have a meaningful name attached.

src/modules/initcpiocfg/main.py
```diff
@@ -198,6 +202,9 @@ def find_initcpio_features(partitions, root_mount_point):
     if uses_lvm2:
         hooks.append("lvm2")
 
+    if uses_zfs:
+        hooks.append("zfs")
```
adriaandegroot, line 206: Since there's several of these, maybe a more-Pythonic use of `any()` in initializing the variable (because something similar can be applied to the other two `uses_*`)

src/modules/mount/main.py
```diff
@@ -61,6 +61,97 @@ def get_btrfs_subvolumes(partitions):
     return btrfs_subvolumes
 
 
+def parse_global_storage(gs_value):
+    """ Something in the chain is converting on and off to true and false.  This converts it back.
```
adriaandegroot: It's not really parsing, though, is it? More like "munge". Or `bool_to_zfs_command_boolean()`

src/modules/mount/main.py
```diff
@@ -61,6 +61,97 @@ def get_btrfs_subvolumes(partitions):
...
+        raise Exception("Internal error mounting zfs datasets")
+
+    # Find the zpool matching this partition
+    for zfs_pool in zfs_pool_list:
+        if zfs_pool["mountpoint"] == partition["mountPoint"]:
+            pool_name = zfs_pool["poolName"]
+            ds_name = zfs_pool["dsName"]
+
+    # import the zpool
+    import_result = subprocess.run(['zpool', 'import', '-R', root_mount_point, pool_name])
```
adriaandegroot: These calls to `subprocess.run()` should be converted to use libcalamares api -- e.g. `host_run_command()` or the like. Not because these are wrong **now**, but at some point Calamares will lose the setuid-root bit, and we will add privilege-escalation as needed on the commands that are run (rather than on the whole UI).

src/modules/mount/main.py
```diff
@@ -61,6 +61,97 @@ def get_btrfs_subvolumes(partitions):
...
+    if import_result.returncode != 0:
+        raise Exception("Failed to import zpool")
```
adriaandegroot: These exceptions seem rather .. generic exceptions. Is there a more specific type you can use here? If these exceptions can escape from the module, it's going to create an ugly failure message. If you can catch the specific type and turn the exception into a failure return from the module -- with a human-readable message -- then it becomes much nicer for the user. ("nicer" is relative: the installation still failed then :( )

adriaandegroot (commented):
The Python bits look pretty ok, from a Python-code perspective. I have not examined how they interact with the *partition* module, though.

src/modules/zfs/ZfsJob.cpp
```diff
@@ -0,0 +1,326 @@
...
+        }
+    }
+    return false;
+}
+
+
+ZfsResult
+ZfsJob::CreateZpool( QString deviceName, QString poolName, QString poolOptions, bool encrypt, QString passphrase ) const
+{
+    // zfs doesn't wait for the devices so pause for 2 seconds to ensure we give time for the device files to be created
```
adriaandegroot, line 138: Why not `sleep(2)` in the C code instead? There does not seem to be another command running here, so doing the delay outside the shell is simpler in many ways -- it doesn't need a shell to interpret multiple commands.

src/modules/zfs/ZfsJob.cpp
```diff
@@ -0,0 +1,326 @@
...
+ZfsJob::~ZfsJob() {}
+
+QString
+ZfsJob::prettyName() const
+{
+    return tr( "Create ZFS pools and datasets" );
+}
+
+QString
+ZfsJob::AlphaNumeric( QString input ) const
```
adriaandegroot: Usual Qt / C++ camel-casing would make that `alphaNumeric` .. but does this need to be a member function? You could make the function `static` (private to this translation unit) unless it's important to use from elsewhere, too.

src/modules/zfs/ZfsJob.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+
+ZfsResult
+ZfsJob::CreateZpool( QString deviceName, QString poolName, QString poolOptions, bool encrypt, QString passphrase ) const
+{
+    // zfs doesn't wait for the devices so pause for 2 seconds to ensure we give time for the device files to be created
+    QString command;
+    if ( encrypt )
+    {
+        command = "sleep 2 ; echo \"" + passphrase + "\" | zpool create " + poolOptions
```
adriaandegroot: Sending input to a command can be done through the QProcess API (or, preferably, the Calamares run-a-command API). Then you can avoid the echo, the pipeline, and the need to run this through a shell -- instead, run `zpool create` with a suitable stdin set.

src/modules/zfs/ZfsJob.cpp
```diff
@@ -0,0 +1,326 @@
...
+            + " -O encryption=aes-256-gcm -O keyformat=passphrase " + poolName + " " + deviceName;
+    }
+    else
+    {
+        command = "sleep 2 ; zpool create " + poolOptions + " " + poolName + " " + deviceName;
+    }
+
+    // We use a qProcess instead of runCommand so the password will not end up in the logs
```
adriaandegroot: If the password is ending up in the logs, then `RedactedCommand` needs some more work; but I think that using `Runner` or the simplified API and setting stdin is a better approach. (Discussion welcome -- I really don't like commands that do not end up in the logs, but that **can** cause errors, because they're hard to debug)

src/modules/zfs/ZfsJob.cpp
```diff
@@ -0,0 +1,326 @@
...
+        {
+            poolName += AlphaNumeric( mountpoint );
+        }
+
+        // Check to ensure the list of zfs info from the partition module is available and convert it to a list
+        if ( !gs->contains( "zfsInfo" ) && gs->value( "zfsInfo" ).canConvert( QVariant::List ) )
+        {
+            return Calamares::JobResult::error( tr( "Internal data missing" ), tr( "Failed to create zpool" ) );
+        }
+        QVariantList zfsInfoList = gs->value( "zfsInfo" ).toList();
```
adriaandegroot: This chunk is done for every iteration of the loop, but isn't it supposed to be constant throughout the loop? If so, it can be lifted. That would also let you do the work, once, of collecting just the mount-points from the zfs info, so that the check below becoms a `QStringList.contains()` (or define a struct with mountpoint and bool encrypted, so you copy it from GS into a map / set / list just once).

src/modules/zfs/ZfsJob.cpp
```diff
@@ -0,0 +1,326 @@
...
+                                                    Calamares::JobResult::InvalidConfiguration );
+    }
+
+    const CalamaresUtils::System* system = CalamaresUtils::System::instance();
+
+    QVariantList poolNames;
+
+    for ( auto& partition : qAsConst( partitions ) )
+    {
+        QVariantMap pMap;
```
adriaandegroot, line 198: From here to L176 looks like a function -- ok, I like short functions that tell a story -- that returns `QPair<QString, QString>` of device-name and mount-point. Maybe some more data-structure wrangling is needed there.

src/modules/zfs/ZfsJob.cpp
```diff
@@ -0,0 +1,326 @@
...
+                gs->insert( "zfsDatasets", datasetList );
+            }
+        }
+        else
+        {
+            // This is a zpool with a single dataset We again set canmount=no regardless of the desired setting.
+            // It is modified to the correct value in the mount module to ensure mount order is maintained
+            QString dsName = mountpoint;
+            dsName = AlphaNumeric( mountpoint );
+            auto r = system->runCommand( { "sh",
```
adriaandegroot: Sort of a general question, why all the `sh -c` calls? To me it looks like this could be done with less string-manipulations and more building-a-command-line-as-a-QStringList.

eg. here,
```
QStringList cmd{ "zfs", "create" };
cmd << m_datasetOptions; // Probably needs actual work, since this might be a bunch of space-separated `-o opt=val` strings
cmd << "-o" << "canmount=off" << "-o" << QStringLiteral("mountpoint=%1").arg(mountpoint);
cmd << (poolName + "/" + dsName);
```

Hm, now that I write it like that I'm not **really** keen on how readable it becomes with streaming operators :(

src/modules/zfs/ZfsJob.h
```diff
@@ -0,0 +1,96 @@
...
+#include "CppJob.h"
+
+#include "utils/PluginFactory.h"
+
+#include "DllMacro.h"
+
+struct ZfsResult
+{
+    bool success;
+    QString failureMessage;
```
adriaandegroot: Can you indicate if that is intended to be a user-visible string (which needs translation) or not? It might not even be necessary to make this visible is `CreateZpool()` is made internal as well.

src/modules/mount/main.py
dalto8: In that case, I think I am doing the wrong thing here.  I am raising those exceptions to provide a human readable error to the user.

If the proper way to do that is for the module to return an error string of some sort, I will look into making it do that instead.

src/modules/zfs/ZfsJob.h
```diff
@@ -0,0 +1,96 @@
...
+     * @p poolOptions are the options to pass to zpool create
+     * @p encrypt is a boolean which determines if the pool should be encrypted
+     * @p passphrase is a string continaing the passphrase
+     *
+     */
+    ZfsResult CreateZpool( QString deviceName,
+                           QString poolName,
+                           QString poolOptions,
+                           bool encrypt,
+                           QString passphrase = QString() ) const;
```
adriaandegroot, line 67: Is encrypted-with-an-empty-passphrase a valid configuration? (e.g. is there ever a situation where `encrypt != !passphrase.isEmpty()` )

src/modules/zfs/ZfsJob.cpp
adriaandegroot: .. looking at the config file, much later, there's string-valued configuration options that get substituted in, which might be anything; that makes the shell-because-command-line-parsing necessary (or we need to invest in `QStringList::split(' ')`)

src/modules/zfs/ZfsJob.cpp
dalto8: > Sending input to a command can be done through the QProcess API 

It can if I do the sleep as recommended above.  I didn't think of that, which caused me to not be able to use stdin.  I suppose I could have user `QProcess.write()`, but that seemed more fragile.

Either way, I think I can clean it up now.

src/modules/zfs/ZfsJob.h
dalto8: > Is encrypted-with-an-empty-passphrase a valid configuration? 

No, if encrypt is false, the passphrase is irrelevant.

src/modules/zfs/ZfsJob.cpp
dalto8: > Sort of a general question, why all the `sh -c` calls? 

I tried to only use `sh -c` when it was needed.  i.e. When I was using shell pipelines.

However, in this case and the one above at line 254, it doesn't look needed to me. I will convert it and try to see if I can make it readable.

PR #1824, opened 2021-11-10
Remove space_cache from btrfs mount options

dalto8:
Since the most release when installing with btrfs the `systemd-remount-fs` service fails to run properly.

Upon investigation it appears the cause of this issue is that the root filesystem is mounted initially with `space_cache=v2` and when it tries to remount it with `space_cache`, it fails.

<s>Although I was unable to find any specific commit which would cause this behaviour, removing `space_cache` from the mount options avoids this conflict.  Additionally, since `space_cache=v1` has been the default since 4.5, I don't think removing it would be impactful.</s>

Based on the below, it seems the problem is caused by the default in kernel 5.15 being changed to space_cache=v2.

### DIFF
--- a/src/modules/fstab/fstab.conf
+++ b/src/modules/fstab/fstab.conf
@@ -12,11 +12,9 @@
 # is listed here, use those options, otherwise use the *default*
 # options from this mapping.
 #
-# With kernels 5.15 and newer be cautious of adding the option
-# space_cache to the btrfs mount options.  If the root volume is initially
-# mounted with one version of space_cache and the mount option specifies
-# a different version or implies v1 when v2 is in use, it may fail to
-# remount properly.
+# With kernels 5.15 and newer be cautious of adding the option space_cache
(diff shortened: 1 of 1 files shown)


demmm:
A few links to explain what changed with btrfs 5.15 that might explain the errors:
https://btrfs.wiki.kernel.org/index.php/Changelog#btrfs-progs_v5.15_.28Nov_2021.29
https://wiki.tnonline.net/w/Btrfs/Space_Cache

From the above it seems that removing is not the right option, rather moving to space_cache=v2.

dalto8:
> From the above it seems that removing is not the right option, rather moving to space_cache=v2.

I think removing it is the safer option given that different distros are on different linux kernel versions.  Setting it to v2 will probably cause the same problem for people using kernels older than 5.15.  By leaving it omitted, the default version of space_cache specified by the kernel will be applied.

Issue #1821, opened 2021-11-07
systemd-boot has hard-coded btrfs root subvolume


dalto8:
While doing some other work in Calamares I noticed this:
```
        if partition["mountPoint"] == "/" and partition["fs"] == "btrfs":
            kernel_params.append("rootflags=subvol=@")
```

Since the root subvolume is configurable in the mount module, this seems bad.

I wanted to log this since I am in the middle of some other development at the moment.

PR #1827, opened 2021-11-18
Ensure root subvolume is set correctly for systemd-boot

dalto8:
This fixes issue https://github.com/calamares/calamares/issues/1821

The root subvolume is currently being hard-coded to `@` which is problematic for a couple of reasons:
* The actual root subvolume is configurable in mount.conf
* Sometimes, distros choose to not have a separate subvolume for root

This should address both those issues.

NOTE: I must have accidentally triggered my editor to fix format issues in the mount module.  I took a look at the changes and they all look sensible so I left them.  If you would prefer them removed from this PR just let me know.

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -178,10 +178,13 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
                                   + partition["luksMapperName"]]
 
     for partition in partitions:
-        # systemd-boot with a BTRFS root filesystem needs to be told
-        # about the root subvolume.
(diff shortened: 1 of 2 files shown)


adriaandegroot:
The style fixes on the Python code are fine. We used to have fairly stringent PEP8-checks, but that ran headlong into mindless PEP8-ery which is not the point of that document. The C++ coding style is enforced with tooling and very spaced-out and this leaks into the Python code, too (I sometimes terminate lines with a `;` in Python if I've been doing too much C++ for a while).

I do think this deserves a follow-up PR documenting some of the global settings values and how they are used by different modules (there are more special-secret-GS keys for the code modules, and they're fragile).

PR #1828, opened 2021-11-18
Remove setState call to resolve compat issue with older kpmcore

dalto8:
It seems that `Partition::State::None` isn't available in older versions of kpmcore.

However, after performing some testing, this call is non-critical to the function of the application.

Removing it is probably the most sensible way to resolve this issue.


adriaandegroot:
```
src/modules/partition/core/KPMHelpers.h
30:#define KPM_PARTITION_STATE( x ) Partition::State::x
34:#define KPM_PARTITION_STATE( x ) Partition::State##x
```

That's some compatibility-code that we already have, but if the call is superfluous anyway, no worries.

PR #1833, opened 2021-11-26
Additional pacman options in the packages module

dalto8:
This PR adds 3 options for pacman to the packages module
* Allows you to set a number of retries that the calls to pacman will attempt if it fails
* Supports optionally sending the `--disable-download-timeout` flag to pacman
* Adds config to optionally add the `--needed` flag to pacman calls

By default, all values are set at their previous defaults.

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -442,9 +442,8 @@ class PMPacman(PackageManager):
 
         command += pkgs
 
-        self.run_pacman(command, True)
-
         self.reset_progress()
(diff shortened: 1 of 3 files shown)


adriaandegroot:
I'm going to merge this, and then follow it up with changing the way the job configuration is done. Instead of keys *pacman_something*, let's use sub-keys: one key *pacman* (for configuring pacman) with *something* as a sub-key.

killajoe:
we will test all this as soon as we got the current release out of the box ;)

PR #1836, opened 2021-11-30
Improve Vietnamese translation

th1nhhdk:
Also, who is Vigo, and why should i give him my name? (at line 2539)

### DIFF
--- a/lang/calamares_vi.ts
+++ b/lang/calamares_vi.ts
@@ -655,27 +655,27 @@ Trình cài đặt sẽ thoát và tất cả các thay đổi sẽ bị mất.<
     <message>
       <location filename="../src/modules/partition/gui/ChoicePage.cpp" line="1641"/>
       <source>No Swap</source>
-      <translation>Không hoán đổi</translation>
+      <translation>Không tạo phân vùng hoán đổi(swap)</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


demmm:
Translations are done on Transifex, might consider joining & working there:

adriaandegroot:
@th1nhhdk Going through Transifex makes things a lot easier for us, because the tools are nicer. We **can** upload this change to Transifex, but it might be faster to re-do it there (your call).

I was confused by the question about Vigo, but the translation files look like this:
```
      <source>What is your name?</source>
      <translation>Hãy cho Vigo biết tên đầy đủ của bạn?</translation>
```
That .. is an interesting translation (I don't speak Vietnamese at all, but just plugging it into online translators is enough). I have no idea why that was ever chosen.

th1nhhdk:
Online translater and linux in my language _sucks_, there isn't enough Vietnamese people that had good enough programming, linux and english skills, Most Vietnamese people that i know don't really know much about technology in general, most of them don't even know what is linux and usually **they only know the basics** of their computers (how to turn it on, off ...), so linux programs in my language sucks and confusing, so much so using linux in vietnamese is rated beginner mistakes.

My mission is to make linux more accessible to Vietnamese in general, because i think the linux community deserve it and the current state of linux in Vietnam sucks

Issue #1820, opened 2021-11-07
using grub creates the same grub-id if install the same OS alongside on the same partition (UEFI dualboot)


killajoe:
Just got a use case where we test installing EndeavourOS using the alongside option on partition slide:
1. install EOS on UEFI system using automatic partition using grub a s bootloader
2. install EOS using alongside option in partition slide
3. reboot
Grub shows only the second installed system and seems to have overwritten the boot entry from first installed system because we use the same grub-id on grub-install command. dir-name is the same as it uses grub-id naming inside /boot/efi (ESP)

I can fix it by chrooting into the first installed system and reinstall grub files using a different grub-id, which will generate an entry for both installed EOS systems. 

using current calamares 3.2.45 and latest EOS calamares configs:
https://github.com/endeavouros-team/EndeavourOS-calamares

demmm:
From the link I don't see that the option to set bootloaderID is used:
https://github.com/calamares/calamares/blob/calamares/src/modules/bootloader/bootloader.conf#L38
If you just set a name without any versioning/ID, not celar how this could be a Calmares bug:
https://github.com/endeavouros-team/EndeavourOS-calamares/blob/main/calamares/modules/bootloader.conf#L14

The option is available right now in Calamares to set versioned name/ID, seems that is what you are looking for, so each ISO will have a unique GRUB ID.

killajoe:
i see we use `bootloaderEntryName:` instead of  `efiBootloaderId:` so it will use the `bootloaderEntryName` from branding.desc  but i do not see any option describing that we can use versioning there.

So could you hint me on:
_**The option is available right now in Calamares to set versioned name/ID**_

killajoe:
![grub-efi-fix-path](https://user-images.githubusercontent.com/16797647/141687569-18475b53-cbab-4080-9d6c-7939b26aabca.png)
we solved that by adding randomized postfix to the pathname:
`   # change bootloaderEntryName in bootloader.conf:
    sed -i $cfolder/modules/bootloader.conf \
        -e "s|^bootloaderEntryName:.*|bootloaderEntryName: \"endeavouros-$(shuf -i 1000-9999 -n1)\"|" `

apt-ghetto:
What is the use case to have the same binaries more then once on the ESP?

If Grub doesn't show the other system, then the used grub.cfg is wrong.

killajoe:
> What is the use case to have the same binaries more then once on the ESP?

[...]

PR #1842, opened 2021-12-04
[bootloader] Fix error with systemd-boot when path exists in the ESP

### LINKED ISSUE #1840: Calamares fails when installing over an existing installation, while using systemd-boot and reusing the ESP
**Describe the bug**
The bootloader module encounters an error when installing, if you have previously installed the same distribution using the same ESP, and are using systemd-boot.

**To Reproduce**
1. Configure Calamares to use systemd-boot.
2. Install to an EFI machine.
3. After installation finishes, reboot back into the live environment and try to install again.
4. Partition in any way, as long as the ESP is kept and reused (as is typical when dual-booting with Windows, for instance).

**Expected behavior**
Calamares should overwrite existing bootloader files, or at least ask the user about it, instead of completely failing.

**Screenshots and Logs**
Screenshot of the error that appears: https://i.imgur.com/YJwJJbk.png

**Additional context**
Currently, the only way to work around this is to either delete your ESP (breaking Windows in a dual-boot setup) or to open the partition and delete your distribution's files manually.

dalto8:
Fixes https://github.com/calamares/calamares/issues/1840

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -224,7 +224,7 @@ def create_systemd_boot_conf(install_path, efi_dir, uuid, entry, entry_name, ker
 
     # Copy kernel and initramfs to a subdirectory of /efi partition
     files_dir = os.path.join(install_path + efi_dir, entry_name)
-    os.mkdir(files_dir)
+    os.mkdir(files_dir, exist_ok=True)
 
     kernel_path = install_path + kernel
     kernel_name = os.path.basename(kernel_path)


hphilm:
LGTM

demmm:
Overwriting without warning? Not a good idea imo, better make the file name unique by adding a date stamp or something similar.

dalto8:
> Overwriting without warning? Not a good idea imo, better make the file name unique by adding a date stamp or something similar.

That would create a mess the way this is currently written.  I think the case where this would overwrite something important is an edge case.  It would require both the entry name and the kernel names to match.

Honestly, what it is doing here is pretty strange to being with.  This is really just a band-aid until 3.3 where this has already been addressed.

That being said, it wouldn't be hard to create unique directory names, I am just not confident that is really the right thing.  If there is a consensus that it would be a better approach, I can make that change.

demmm:
Just checked more, I see now it is adjusting the kernel image location, not the conf files (not clear from the bug report this addresses if the error is related to the kernel image, but seems so). This means, ignore my previous comment.

adriaandegroot:
`ok.mkdir()` is not documented to have an `exist_ok` named parameter. That is an `os.makedirs()` thing.

adriaandegroot:
Uniqueing of the name is something requested in #1820 , and there's a branch to playfully do that.

adriaandegroot:
I fixed this up and cherry-picked to the *calamares* branch.

PR #1843, opened 2021-12-05
Rework of error dialog

LordTermor:
Currently Calamares uses plain QMessageBox to show an error. This PR gives some visual uplift by creating a dedicated dialog with a text browser for error details thus visually separating them from the actual message.
image


adriaandegroot:
Landing first, then shuffling around a bit.

PR #1850, opened 2021-12-11
Comment : preservefiles on Sat Dec 11 06:11:45 PM CET 2021

erikdubois:
Add these extra lines to move the two major files to /var/log to keep them together.
+
If choosing encrypted then /var/log/Calamares.log will not be there unless you add it to preserve files
tested out see telegram channel and here in this image
cal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ CMakeLists.txt.user
 # Kate
 *.kate-swp
 tags
+git*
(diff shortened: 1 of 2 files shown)


adriaandegroot:
I merged with modifications:
- the change to .gitignore seems extra
- The change to the permissions used was undone; the log is sanitized, hopefully well-enough, but it doesn't seem like something that every user of the target system should be able to read. The JSON file is a dump of global storage, which contains fun stuff like usernames and passwords; that should definitely not be world-readable.

PR #1864, opened 2021-12-30
[locale] Update language always

tjpark88:
onActivate of locale updates the language only when currentLocation changed
or when onActivate of locale is called for the first time.
However, It is irrelevant solution since the language is set by the welcome.
So language should be updated always.
The language is used by keyboard module to guessing a layout of keyboard.
Once you face the locale, you can't change language in the welcome
if you don't change the timezone.

### DIFF
--- a/src/modules/locale/Config.cpp
+++ b/src/modules/locale/Config.cpp
@@ -252,15 +252,21 @@ Config::setCurrentLocation( const QString& regionName, const QString& zoneName )
 void
 Config::setCurrentLocation( const CalamaresUtils::Locale::TimeZoneData* location )
 {
-    if ( location != m_currentLocation )
+    bool updateLocation = ( location != m_currentLocation );
+    if ( updateLocation )
     {
         m_currentLocation = location;
-        // Overwrite those settings that have not been made explicit.
-        auto newLocale = automaticLocaleConfiguration();
-        if ( !m_selectedLocaleConfiguration.explicit_lang )
-        {
-            setLanguage( newLocale.language() );
-        }
+    }
+
+    // Always lang should be updated
+    auto newLocale = automaticLocaleConfiguration();
+    if ( !m_selectedLocaleConfiguration.explicit_lang )
+    {
+        setLanguage( newLocale.language() );
+    }
+
+    if ( updateLocation )
+    {
         if ( !m_selectedLocaleConfiguration.explicit_lc )
         {
             m_selectedLocaleConfiguration.lc_numeric = newLocale.lc_numeric;


src/modules/locale/Config.cpp
```diff
@@ -252,15 +252,21 @@ Config::setCurrentLocation( const QString& regionName, const QString& zoneName )
 void
 Config::setCurrentLocation( const CalamaresUtils::Locale::TimeZoneData* location )
 {
-    if ( location != m_currentLocation )
+    bool updateLocation = ( location != m_currentLocation );
```
adriaandegroot, line 255: That should be a `const bool`.

adriaandegroot:
The scenario you have described is rather unclear. I can find the following peculiar behavior, please confirm if this is the problem you are tying to solve:

- start Calamares in debug mode (`calamares -d`) and then open the debug-window
- notice *localeConf* has one key: *LANG*, set to "en" (or whatever the default language is for you when you start Calamares)
- in the *welcome* page, pick Belarussian as language
- notice *LANG* now says "be"
- click through to the locale module, notice now there is also a global key *locale*, set to "be"
- click back to welcome, notice that *LC_* keys are now filled in in *localeConf* (depending on other configurations, based on your timezone; I get nl), but *LANG* there is "be" and in the global *locale*, also "be"
- change language to Danish, notice *LANG* is now "da" but *locale* is still "be"
- click next to get back to the locale page. Notice that (still)  *LANG* is "da" and *locale* is "be". notice also that the UI indicates that the system language will be Belarussian.
- clicking past the locale module sets *LANG* to "be" (which is what the UI indicated)

So is the issue "sometimes the language selected on the welcome page affects the overall installation, and sometimes (the second time you visit the locale page) it doesn't" ?

adriaandegroot:
I think your PR fixes some *other* problem than what I've described -- since my issue is still there. I agree it is confusing. In branch *issue-1864* I have taken your change and am looking to fix some other bits so that it ends up consistent.

@tjpark88 I mention people in the changelog for Calamares when they make a contribution, like you have. Should I mention you as "Taejun Park" (list of names) and "Thanks Taejun" (in text), or would "PARK Taejun" and "Thanks Taejun" be more appropriate? Or something else.

tjpark88:
> The scenario you have described is rather unclear. I can find the following peculiar behavior, please confirm if this is the problem you are tying to solve:
> 
> * start Calamares in debug mode (`calamares -d`) and then open the debug-window
> * notice _localeConf_ has one key: _LANG_, set to "en" (or whatever the default language is for you when you start Calamares)
> * in the _welcome_ page, pick Belarussian as language
> * notice _LANG_ now says "be"
> * click through to the locale module, notice now there is also a global key _locale_, set to "be"
> * click back to welcome, notice that _LC__ keys are now filled in in _localeConf_ (depending on other configurations, based on your timezone; I get nl), but _LANG_ there is "be" and in the global _locale_, also "be"
> * change language to Danish, notice _LANG_ is now "da" but _locale_ is still "be"
> * click next to get back to the locale page. Notice that (still)  _LANG_ is "da" and _locale_ is "be". notice also that the UI indicates that the system language will be Belarussian.
> * clicking past the locale module sets _LANG_ to "be" (which is what the UI indicated)
> 
> So is the issue "sometimes the language selected on the welcome page affects the overall installation, and sometimes (the second time you visit the locale page) it doesn't" ?

Your explanation is what I want to describe.

tjpark88:
> I think your PR fixes some _other_ problem than what I've described -- since my issue is still there. I agree it is confusing. In branch _issue-1864_ I have taken your change and am looking to fix some other bits so that it ends up consistent.

My PR intends to fix the problem what you've described.
- Sometimes the language selected on the welcome page affects the overall installation, and sometimes (the second time you visit the locale page) it doesn't

What should I do? should I change bool to const bool and PR again?

> 
> @tjpark88 I mention people in the changelog for Calamares when they make a contribution, like you have. Should I mention you as "Taejun Park" (list of names) and "Thanks Taejun" (in text), or would "PARK Taejun" and "Thanks Taejun" be more appropriate? Or something else.

"Taejun Park" (list of names) and "Thanks Taejun" (in text) would be grateful.

adriaandegroot:
Merged with some minor additions. Thank you!

PR #1865, opened 2021-12-30
Changes to mount module for 3.3

dalto8:
This PR has a number of changes to the mount module.

Here is the TLDR version with more detailed explanations below:
* Moved mount options to the mount module
* Refactored the mount config settings
* Added support for separate options for both hdds and ssds
* Added an option to set the name of the subvol for the brtrs swapfile
* Refactored the remainder of the mount config to simply and avoid duplication
* Added documentation

The first and most significant change is that the mount options were moved from the fstab module to the mount module.  This has the advantage of applying mount options to the files in the install.  It will apply all mount options but the most notable is probably for filesystems like btrfs that support compression via a mount option.  Currently, the files from the install aren't being compressed in that scenario.

In moving the mount options config, it has also been refactored.  There are 3 primary changes here:
* Group the config together by filesystem
* Use arrays to store lists instead of comma separated strings
* Add support for additional options for HDDs in addition to SSDs

Since major changes were being made to the mount module config the opportunity was taken to refactor the remaining settings.  Most notably:
* `extraMountsEfi` and `extraMounts` were combined
* A setting for the name of the subvol for the brtrs swapfile was added so it was no longer hardcoded to `@swap`
* As above, it now uses arrays to store lists instead of comma separated strings

Lastly, additional documentation was added throughout the mount module.

It should be noted that since the fstab module needs a rewrite, the changes were to fstab were made as minimal as possible.

### DIFF
--- a/src/modules/fstab/fstab.conf
+++ b/src/modules/fstab/fstab.conf
@@ -5,57 +5,8 @@
 # Also creates mount points for all the filesystems.
 #
 # When creating fstab entries for a filesystem, this module
-# uses the options for the filesystem type to write to the
-# options field of the file.
+# uses the options previously defined in the mount module
 ---
-# Mount options to use for all filesystems. If a specific filesystem
-# is listed here, use those options, otherwise use the *default*
-# options from this mapping.
-#
-# With kernels 5.15 and newer be cautious of adding the option space_cache
-# to the btrfs mount options.  The default in 5.15 changed to space_cache=v2.
-# If space_cache or space_cache=v1 are specified, it may fail to remount.
-#
-# btrfs_swap options are used when a swapfile is chosen with a btrfs root
-# the options are applied to the subvolume which holds the swap partition
-#
-# The settings shown here apply only the btrfs defaults; these
-# are generally the right ones. Commented-out lines show other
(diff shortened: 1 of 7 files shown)


adriaandegroot:
This looks good overall, and reminds me of two things:

- the fstab-module-creates-mountpoints comment (it's not something this PR touches, but is at the top of `fstab.conf`) is a bit weird; shouldn't those already be created by the mount module when it does whatever mounting it is going to do?
- overall, better documentation of global storage keys and how they couple different modules together is needed

dalto8:
> * the fstab-module-creates-mountpoints comment (it's not something this PR touches, but is at the top of `fstab.conf`) is a bit weird; shouldn't those already be created by the mount module when it does whatever mounting it is going to do?

It does try to create them if they don't exist(using a pointless wrapper to `os.makedirs()`).  However, to your point, I can't think of any situation where they won't already exist.

> * overall, better documentation of global storage keys and how they couple different modules together is needed

Since I am guilty of creating some(many?) of those I can start by documenting the keys I have added.  Where do we want that documentation?  It seems like it should be centralized somewhere to make it easy to find.  Add a section to the end of the modules readme?

PR #1879, opened 2022-01-22
Allow the packagechooser module to modify the netinstall module

dalto8:
This adds two new optional methods to the packagechooser module.  Each of these methods are used to modify the netinstall module.  The netinstall module is also modified to read and act on these optional values.

#### The first is netinstall-add
This allows for dynamically adding groups/packages to the netinstall tree.  There are many use cases for this functionality but the way it is being used initially is to give the user a choice between several custom editions which are mutually exclusive.  Adding them to the netinstall has two advantages.  It adds clarity to the end-user so they aren't wondering why the things they selected earlier in the process aren't showing up.  It allows them to refine the selection, for example if they want to remove some of the optional packages.

#### The second is netinstall-select
This allows you to preselect certain groups.  The use case for this is when you have a situation where you want to provide a "pretty" way to describe a group of packages and make selections easy but still allow the user to refine those selections.  For example, you could put a DE picker before the netinstall module and give the user a lot more information about each DE.

### DIFF
--- a/src/modules/netinstall/NetInstallPage.cpp
+++ b/src/modules/netinstall/NetInstallPage.cpp
@@ -76,6 +76,6 @@ NetInstallPage::onActivate()
     const QVariantList groups = gs->value( "NetinstallAdd" ).toList();
     if ( !groups.isEmpty() )
     {
-        m_config->model()->appendModelData( groups, "packageChooser" );
+        m_config->model()->appendModelData( groups );
     }
 }
--- a/src/modules/netinstall/NetInstallPage.cpp
+++ b/src/modules/netinstall/NetInstallPage.cpp
@@ -15,6 +15,7 @@
 #include "PackageModel.h"
 #include "ui_page_netinst.h"
 
+#include "GlobalStorage.h"
 #include "JobQueue.h"
 
 #include "network/Manager.h"
@@ -62,4 +63,19 @@ void
 NetInstallPage::onActivate()
 {
     ui->groupswidget->setFocus();
+
+    Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
+    if ( gs->contains( "NetinstallSelect" ) && gs->value( "NetinstallSelect" ).canConvert( QVariant::StringList ) )
+    {
+        const QStringList selectNames = gs->value( "NetinstallSelect" ).toStringList();
+
+        static_cast< PackageModel* >( ui->groupswidget->model() )->setSelections( selectNames );
+    }
+
+    if ( gs->contains( "NetinstallAdd" ) && gs->value( "NetinstallAdd" ).canConvert( QVariant::List ) )
+    {
+        const QVariantList groups = gs->value( "NetinstallAdd" ).toList();
+
+        static_cast< PackageModel* >( ui->groupswidget->model() )->appendModelData( groups );
+    }
 }
--- a/src/modules/netinstall/NetInstallPage.cpp
+++ b/src/modules/netinstall/NetInstallPage.cpp
@@ -64,20 +64,18 @@ NetInstallPage::onActivate()
 {
     ui->groupswidget->setFocus();
 
-    // The NetInstallSelect global sotrage value can be used to make additional items selected by default
+    // The NetInstallSelect global storage value can be used to make additional items selected by default
     Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
-    if ( gs->contains( "NetinstallSelect" ) && gs->value( "NetinstallSelect" ).canConvert( QVariant::StringList ) )
+    const QStringList selectNames = gs->value( "NetinstallSelect" ).toStringList();
+    if ( !selectNames.isEmpty() )
     {
-        const QStringList selectNames = gs->value( "NetinstallSelect" ).toStringList();
-
-        static_cast< PackageModel* >( ui->groupswidget->model() )->setSelections( selectNames );
+        m_config->model()->setSelections( selectNames );
     }
 
     // If NetInstallAdd is found in global storage, add those items to the tree
-    if ( gs->contains( "NetinstallAdd" ) && gs->value( "NetinstallAdd" ).canConvert( QVariant::List ) )
+    const QVariantList groups = gs->value( "NetinstallAdd" ).toList();
+    if ( !groups.isEmpty() )
     {
-        const QVariantList groups = gs->value( "NetinstallAdd" ).toList();
-
-        static_cast< PackageModel* >( ui->groupswidget->model() )->appendModelData( groups, "packageChooser" );
+        m_config->model()->appendModelData( groups, "packageChooser" );
     }
 }
--- a/src/modules/netinstall/NetInstallPage.cpp
+++ b/src/modules/netinstall/NetInstallPage.cpp
@@ -78,6 +78,6 @@ NetInstallPage::onActivate()
     {
         const QVariantList groups = gs->value( "NetinstallAdd" ).toList();
 
-        static_cast< PackageModel* >( ui->groupswidget->model() )->appendModelData( groups );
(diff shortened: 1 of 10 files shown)


src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -62,4 +63,21 @@ void
 NetInstallPage::onActivate()
 {
     ui->groupswidget->setFocus();
+
+    // The NetInstallSelect global sotrage value can be used to make additional items selected by default
```
adriaandegroot: Typo

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -62,4 +63,21 @@ void
...
+    Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
+    if ( gs->contains( "NetinstallSelect" ) && gs->value( "NetinstallSelect" ).canConvert( QVariant::StringList ) )
```
adriaandegroot: This could probably be shorter by just *doing* the conversion to stringlist, and then checking the list if it is empty?

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -62,4 +63,21 @@ void
...
+    {
+        const QStringList selectNames = gs->value( "NetinstallSelect" ).toStringList();
+
+        static_cast< PackageModel* >( ui->groupswidget->model() )->setSelections( selectNames );
```
adriaandegroot: This is weird. There should be "better" pointers to the model, that don't need casting. `m_config` is there in the `NetInstallPage` class for this reason.

src/modules/netinstall/PackageModel.cpp
```diff
@@ -170,6 +170,38 @@ PackageModel::headerData( int section, Qt::Orientation orientation, int role ) c
...
+ *
+ * Recursively traverses the tree pointed to by m_rootItem and
+ * checks if a group name matches any of the items in @p selectNames.
+ * If a match is found, set check the box for that group and it's children.
+ *
+ * Individual packages will not be matched.
+ *
+ */
+void
+PackageModel::setSelections( QStringList selectNames )
```
adriaandegroot: `const QStringList&`, even though QStringList is implicit-shared and cheap to copy.

src/modules/netinstall/PackageModel.cpp
```diff
@@ -170,6 +170,38 @@ PackageModel::headerData( int section, Qt::Orientation orientation, int role ) c
...
+{
+    if ( m_rootItem )
+    {
+        setSelections( selectNames, m_rootItem );
+    }
+}
+
+void
+PackageModel::setSelections( QStringList selectNames, PackageTreeItem* item )
```
adriaandegroot: `const QStringList&`, even though QStringList is implicit-shared and cheap to copy.

src/modules/netinstall/PackageModel.cpp
```diff
@@ -309,3 +341,27 @@ PackageModel::setupModelData( const QVariantList& l )
     setupModelData( l, m_rootItem );
     emit endResetModel();
 }
+
+void
+PackageModel::appendModelData( const QVariantList& groupList, const QString source )
```
adriaandegroot: `const QString&`

src/modules/netinstall/NetInstallPage.cpp
```diff
@@ -62,4 +63,21 @@ void
...
+    }
+
+    // If NetInstallAdd is found in global storage, add those items to the tree
+    if ( gs->contains( "NetinstallAdd" ) && gs->value( "NetinstallAdd" ).canConvert( QVariant::List ) )
+    {
+        const QVariantList groups = gs->value( "NetinstallAdd" ).toList();
+
+        static_cast< PackageModel* >( ui->groupswidget->model() )->appendModelData( groups, "packageChooser" );
```
adriaandegroot: You might need `QStringLiteral()` there around the string, for people who compile with `QT_NO_CAST_ASCII` (I should add that to the test-suite)

src/modules/netinstall/PackageModel.cpp
```diff
@@ -170,6 +170,38 @@ PackageModel::headerData( int section, Qt::Orientation orientation, int role ) c
     return QVariant();
 }
 
+/** @brief Sets the checked flag on matching groups in the tree
```
adriaandegroot: Documentation would typically go in the header. I know for modules which don't install the headers at all this is less of an issue, but I'd typically look in the header **first** for explanatory comments.

src/modules/netinstall/PackageModel.cpp
```diff
@@ -309,3 +341,27 @@ PackageModel::setupModelData( const QVariantList& l )
...
+            if ( child->source() == source )
+            {
+                m_rootItem->removeChild( i );
+            }
+        }
+
+        // Add the new data to the model
+        setupModelData( groupList, m_rootItem );
+
+        emit endResetModel();
```
adriaandegroot: Please use `Q_EMIT` (I think it's ugly, but it is the way forward to "unclaim" `emit` as a special-for-Qt keyword)

src/modules/netinstall/PackageModel.h
```diff
@@ -54,9 +54,25 @@ class PackageModel : public QAbstractItemModel
     int rowCount( const QModelIndex& parent = QModelIndex() ) const override;
     int columnCount( const QModelIndex& parent = QModelIndex() ) const override;
 
+    void setSelections( QStringList selectNames );
+    void setSelections( QStringList selectNames, PackageTreeItem* item );
```
adriaandegroot: This is a private implementation detail, isn't it?

src/modules/netinstall/PackageModel.cpp
adriaandegroot: This is an implementation detail, and doesn't need to be a member function at all (it does not refer to any member variables). It could be a static free function instead.

src/modules/netinstall/PackageTreeItem.cpp
```diff
@@ -248,6 +249,19 @@ PackageTreeItem::setChildrenSelected( Qt::CheckState isSelected )
...
+void
+PackageTreeItem::removeChild( int row )
+{
+    if ( row < m_childItems.count() )
+    {
+        m_childItems.removeAt( row );
+    }
+    else
+    {
+        cWarning() << "Attempt to remove invalid child in removeChild() at row " + QString::number( row );
```
adriaandegroot: `<< row`, there's no need to do QString manipulations here.

adriaandegroot:
I threw a bunch of nits over the wall there, but my underlying question is "under what circumstances would you mix netinstall and packagechooser?" Because that seems like a really weird mix of UI styles to me. I get the "power users want to tweak every bit" but this seems like it's creating cross-coupling between modules for very little benefit.

src/modules/netinstall/NetInstallPage.cpp
dalto8: Ahh.....why do I have hardcoded reference to the packagchooser in the netinstall module?  I am pretty sure I meant to pass that data in.  I should definitely fix that.

dalto8:
> my underlying question is "under what circumstances would you mix netinstall and packagechooser?"

The packagechooser module allows you to include a lot more information in a more approachable way than the netinstall module.  On the other hand, the netinstall module allows an easy way to make lots of detailed changes.  By using both, you get the best of both worlds.

For a more real world example, the EndeavourOS community mode installer lets you select between a set of mutually exclusive editions using the packagechooser module.  Then you later get to the netinstall module to refine other packages selections.  The fact that the packages from the netinstall module don't show up there is confusing to end users.  As is the fact that they can refine almost every package in the system *except* the ones for the community edition.

dalto8:
This should be ready for another review.

PR #1882, opened 2022-01-27
[grubcfg] Add configuration for default kernel_params

dalto8:
Currently the default kernel params in the grubcfg module are hard-coded to "quiet"

This makes that an option in the conf file.


adriaandegroot:
Something we do try to avoid is suprises for distro's when configuration keys change. The distro's shouldn't **have** to update their configuration files for a new Calamares release unless we shout it from the rooftops (e.g. the change to umount module). So for distro's that have this module, but don't have the key in their configuration, this would change the kernel parameters from "quiet" (old fixed value) to empty. I've changed the default value for when the key isn't set and updated the documentation a bit.

PR #1899, opened 2022-03-06
allow viewing live log during execution

### LINKED ISSUE #1896: Option to see the live log during install
Currently whenever there are problems with an installation you're supposed to look at the log files afterwards.

It would be nice if there was a button in the installation to see the log of the installation itself within the installer. This makes it easier to see what is taking long or why the progress bar isn't progressing.

This can give users more information what the setup is doing at the moment. In addition, it gives developers more tools to debug issues.

The installer of elementary OS has such a button during installation:

bobvanderlinden:
Resolves https://github.com/calamares/calamares/issues/1896

This adds a button during execution to switch between the slideshow and a log viewer. The log viewer reads (and follows) the log file.

image

### DIFF
--- a/src/libcalamaresui/viewpages/ExecutionViewStep.cpp
+++ b/src/libcalamaresui/viewpages/ExecutionViewStep.cpp
@@ -30,6 +30,14 @@
 #include <QLabel>
 #include <QProgressBar>
 #include <QVBoxLayout>
+#include <QHBoxLayout>
+#include <QToolBar>
+#include <QAction>
+#include <QToolButton>
+#include <QTabWidget>
+#include <QPlainTextEdit>
+#include <QTabBar>
+#include "utils/Logger.h"
 
 static Calamares::Slideshow*
 makeSlideshow( QWidget* parent )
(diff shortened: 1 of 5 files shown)


bobvanderlinden:
Currently I'm a bit stuck. The log is read from a separate thread (`logThread`). It reads new chunks from the log file. Every time it does so, it'll call into the `ExecutionViewStep::logChunkAvailable` method where the new chunk of log text is appended to the log `QPlainTextEdit` widget.

However, whenever the `logChunkAvailable` calls into any of the widget's methods the application crashes without any indication why. The chunk string seems valid, because I can log that chunk to stdout correctly. The thread id seems to be the same one as `onActivate` is called on, so I think it's the main thread. The `this` pointer seems correct, as it is the same value as the `this` of `onActivate`.

Any ideas on how to move forward?

bobvanderlinden:
Found out what went wrong. I've moved the log reading/following to a separate LogWidget. The widget could be reused. I see potential reusage in the debug window and/or a 'show log' in the finish step.

That said, `session.log` is still a bit too spammy. I'm still looking into whether it's possible to have a cleaner log, like only showing the (shell) commands that are executed. Suggestions are welcome.

src/libcalamaresui/viewpages/ExecutionViewStep.cpp
```diff
@@ -200,10 +225,18 @@ ExecutionViewStep::updateFromJobQueue( qreal percent, const QString& message )
     }
 }
 
+void
+ExecutionViewStep::toggleLog()
+{
+    m_tab_widget->setCurrentIndex((m_tab_widget->currentIndex() + 1) % m_tab_widget->count());
```
adriaandegroot, line 231: This is clever. Possibly too clever by half, but so long as it is hidden behind the well-named `toggleLog()` it's an implementation detail. I think I would have written `m_tab_widget->currentIndex() ? 0 : 1` which is still a bit opaque.

bobvanderlinden: That would've been sufficient as well. I had cycling through the tabs in mind :sweat_smile:

bobvanderlinden:
Thanks for the review and merge!

arcolinuxz:
Great addition to Calamares indeed.

Can I make two remarks/improvements?

1. Is it possible to keep the slider to the right fixed? Every time there is a new slider qml instance we get dragged to the bottom. - https://youtu.be/RFYXP-763Lg?t=886
2. Can you add a button at end in the finished module to have a look at the log again.
https://youtu.be/RFYXP-763Lg?t=1138

adriaandegroot:
I wsa wondering about the separate thread though -- let's test what we've got first . I was thinking there ought to be a `QSocketNotifier`-like thing for files, but `QFile` in particular notes it does not emit `readyRead()`.If the threading is annoying, a `QTimer`-based loop might do the trick instead.

I'll do a little bit of opinionated polishing now it's merged, and then we can see it in action in 3.2.54 (in particular, I'll see if I can easily implement @arcolinuxz 's requests)

arcolinuxz:
added the time in the videolink - so you do not have to look too long

bobvanderlinden:
Great feedback. The additions you mention are extra obvious when seeing it in a real installation.

Interesting that the icon for the toggle log button didn't show up. I guess the theme misses the terminal icon. I'm wondering what name should be used to get a bit more compatibility with other themes.

I was also thinking that the next slide log lines aren't that useful. It might be better to have them as debug/verbose logs, so that by default the log stays a bit cleaner. Unless I'm overlooking an additional purpose for the next slide log lines?

> I wsa wondering about the separate thread though -- let's test what we've got first . I was thinking there ought to be a QSocketNotifier-like thing for files, but QFile in particular notes it does not emit readyRead().If the threading is annoying, a QTimer-based loop might do the trick instead.

I initially considered using a file watcher, however it thought it might get spammy on the main loop when individual lines are flushed to the log. I also read that file watching isn't guarenteed to trigger due to the inotify API. With that in mind I opted for polling the file size of the log. I haven't heard of socket notifing, so I'm interested how it fares for files.

I have a bit more time next weekend to work on this 👍

adriaandegroot:
I'm doing a little polishing and looking-into-Erik's requests. I'll merge that shortly, and then you, @bobvanderlinden , can carry on from the *calamares* branch when you feel like it. I've added a Copyright statement to `LogWindow.h` and `.cpp`.

PR #1903, opened 2022-03-14
Avoid Reinitializing partition layout.

santoshmahto7:
Partition layout are determined from config file - partition.conf
Calling init will cause to clear layout entries, thus loosing
partition layout specified in conf file.
Without this, calamares end up creating partitions layout that is made
of only boot and root partition

Signed-off-by: Santosh Mahto <santosh.mahto@collabora.com>

### DIFF
--- a/src/modules/partition/core/PartitionActions.cpp
+++ b/src/modules/partition/core/PartitionActions.cpp
@@ -108,12 +108,6 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
         partType = isEfi ? PartitionTable::gpt : PartitionTable::msdos;
     }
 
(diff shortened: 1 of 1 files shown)


adriaandegroot:
Hi Santosh! Thanks for looking into this. Your PR made me look, and wonder why `initLayout()` was being called there. The reason it's called is that the user may have picked a different default filesystem; the layout can have "Unknown" partitions, which get the default FS (whatever that is). So the layout does need to be told what the new default is.

I did a small refactor to expose setting-the-default-FS to the rest of the partition module, and removed the call to `initLayout()`. That is why this PR conflicts, but the underlying issue is fixed based on your change.

So I'm closing this PR with "applied, with changes".

PR #1905, opened 2022-03-14
Avoid setting rootfs partition name to "root" by default.

### LINKED ISSUE #1904: Root partition label is always "root"  even if partition layout in conf names it differently
**Describe the bug**

In target image, root partition label are not picked up from parition layout specified in partition.conf instead always set to "root"
e.g if parition.conf specify layout as below where we name root partition as rootfs-A then it is ignored and target rootfs is named "root" which is incorrect
```
partitionLayout:
    - name: "efi"
      filesystem: "fat32"
      mountPoint: "/tmp/A/efi"
      size: 256MiB
    - name: "rootfs-A"
      type: "4f68bce3-e8cd-4db1-96e7-fbcaf984b709"
      filesystem: "ext4"
      mountPoint: "/tmp/A"
      size: 10GiB
    - name: "var-A"
```

**Expected behavior**
Target rootfs should be named as specified in partition layout in settings

santoshmahto7:
By default, calamares renames the label of root partition
to "root" overriding the name specified in partiton.conf

Signed-off-by: Santosh Mahto <santosh.mahto@collabora.com>

### DIFF
--- a/src/modules/partition/core/PartitionCoreModule.cpp
+++ b/src/modules/partition/core/PartitionCoreModule.cpp
@@ -982,10 +982,6 @@ PartitionCoreModule::layoutApply( Device* dev,
         {
             part->setLabel( "boot" );
         }
-        if ( is_root( part ) )
-        {
-            part->setLabel( "root" );
-        }
         if ( ( separate_boot_partition && is_boot( part ) ) || ( !separate_boot_partition && is_root( part ) ) )
         {
             createPartition(


src/modules/partition/core/PartitionCoreModule.cpp
```diff
@@ -982,10 +982,6 @@ PartitionCoreModule::layoutApply( Device* dev,
         {
             part->setLabel( "boot" );
         }
```
evelikov, line 984: The same principle applies for all partitions right? It so, we should also drop the `boot` hunk, correct?

Disclaimer: I know very little about the calamares internals :-P

adriaandegroot:
I'd suggest applying this to
- both boot, **and** root, as @evelikov says
- expanding the checks `is_*()` to handle trailing slashes (I don't know if anything normalizes the mountpoint)
- checking if the existing label is **empty** and then setting it, rather than just deleting the setting-of-the-label

adriaandegroot:
The existing code with various lambdas and such might be refactored to add an `void applyDefaultLabel(Partition* partition, const QString& mountPoint, const QString& label){ ... }`, leading to two calls to that function (one for boot, one for root). It can have the smarts to check for trailing `/` etc.

adriaandegroot:
I'm merging so this has your name on it, but I'm going to apply the various changes listed above.

PR #1906, opened 2022-03-15
Added spaces and line-breaks to fix formatting.

ladysharkyshark:
- Fixes a formatting issue by adding some spaces and a line-break. Increases readability.


adriaandegroot:
Heh, I rarely look at the rendered version of those docs, I would never have noticed.

PR #1909, opened 2022-03-22
packages: Add support for the AUR via yay

kennystrawnmusic:
Although pamac has AUR support itself, it uses Polkit instead of sudo for authentication. This in turn renders it incapable of suspending password prompts even with NOPASSWD: in sudoers and therefore also renders it unsuitable when trying to install AUR packages upon system installation.

With yay, on the other hand, sudo itself is supported — which in turn makes yay much easier to configure for use in a system-preinstallation environment. So it's time to add it as an option to the packages module, because it should make a lot more things possible through calamares than are currently the case.

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -222,8 +222,6 @@ def run():
     Mount all the partitions from GlobalStorage and from the job configuration.
     Partitions are mounted in-lexical-order of their mountPoint.
     """
-    import os
-
     partitions = libcalamares.globalstorage.value("partitions")
 
     if not partitions:
@@ -232,7 +230,6 @@ def run():
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
 
     root_mount_point = tempfile.mkdtemp(prefix="calamares-root-")
-    os.system("mount -t tmpfs tmpfs " + root_mount_point + "/etc/sudoers.d")
 
     # Guard against missing keys (generally a sign that the config file is bad)
     extra_mounts = libcalamares.job.configuration.get("extraMounts") or []
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -222,6 +222,8 @@ def run():
     Mount all the partitions from GlobalStorage and from the job configuration.
     Partitions are mounted in-lexical-order of their mountPoint.
     """
+    import os
+
     partitions = libcalamares.globalstorage.value("partitions")
 
     if not partitions:
@@ -230,6 +232,7 @@ def run():
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
 
     root_mount_point = tempfile.mkdtemp(prefix="calamares-root-")
+    os.system("mount -t tmpfs tmpfs " + root_mount_point + "/etc/sudoers.d")
 
     # Guard against missing keys (generally a sign that the config file is bad)
     extra_mounts = libcalamares.job.configuration.get("extraMounts") or []
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -475,6 +475,107 @@ class PMPacman(PackageManager):
 
         self.run_pacman(command)
 
+class PMyay(PackageManager):
+    backend = "yay"
+
+    def __init__(self):
+        import re
+        progress_match = re.compile("^\\((\\d+)/(\\d+)\\)")
+
+        def line_cb(line):
+            if line.startswith(":: "):
+                self.in_package_changes = "package" in line or "hooks" in line
+            else:
+                if self.in_package_changes and line.endswith("...\n"):
+                    # Update the message, untranslated; do not change the
+                    # progress percentage, since there may be more "installing..."
+                    # lines in the output for the group, than packages listed
+                    # explicitly. We don't know how to calculate proper progress.
+                    global custom_status_message
+                    custom_status_message = "yay: " + line.strip()
+                    libcalamares.job.setprogress(self.progress_fraction)
+            libcalamares.utils.debug(line)
+
+        self.in_package_changes = False
+        self.line_cb = line_cb
+
+        yay = libcalamares.job.configuration.get("yay", None)
+        if yay is None:
+            yay = dict()
+        if type(yay) is not dict:
+            libcalamares.utils.warning("Job configuration *yay* will be ignored.")
+            yay = dict()
+        self.yay_num_retries = yay.get("num_retries", 0)
+        self.yay_disable_timeout = yay.get("disable_download_timeout", False)
+        self.yay_needed_only = yay.get("needed_only", False)
+
+    def reset_progress(self):
+        self.in_package_changes = False
+        # These are globals
+        self.progress_fraction = (completed_packages * 1.0 / total_packages)
+
+    def run_yay(self, command, callback=False):
+        """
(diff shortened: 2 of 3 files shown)


kennystrawnmusic:
Unfortunately there is one major hurdle here that needs to be overcome before this can be merged (exactly why "allow edits by maintainers" is specified): AUR packages need to be installed as an ordinary user, not as root.

So, is there a way for one module to hook into another in order to retrieve someone's username?

kennystrawnmusic:
Tested with AUR-specific packages and can confirm that these changes definitely work.

adriaandegroot:
From a system-installer perspective it seems really weird to compile / build packages on the target system during installation. It's going to take longer, and be more annoying, than either installing the binaries, or queueing up a build-and-install to do upon reboot. On the other hand, an OEM-configuration tool might want to use this. It's a toss-up: this is not a lot of code and we don't promise maintainence on package managers anyway unless the distro's that use them do the work.

The code seems to write `/etc/sudoers` in the host system, and executes with `sudo` (oh, wait, not because you want to run as root, but because you want to run as *nobody*) in the target system. It also copies some useless code from the pacman bits about progress reporting (which isn't used, is broken right now, and is looking for pacman output anyway).

kkofler:
Why would you use `sudo` to `su` from `root` to `nobody`? Just use `su nobody -c …`.

kkofler:
That said, as I understand it, the sudoers entry is actually to allow `yay` to `sudo` back to `root` to actually install the package with `pacman`. That is not something you want to retain on the installed system though, it is a big gaping security hole.

demmm:
Some notes on this:
>maintenance on package managers

This is not a PR for a package manager, this is for a package builder-helper (wrapper for pacman & makepkg commands), so imo it is making some false claims as to where it is added.

This will start adding support for Calamares as a source based installer,  completely out of the scope of the Calamares intentions.   To top it off, it tries to build from unsupported, not maintained by any distribution sources, where there is zero control during the installation process of what the quality/security of the package is.

So, to put it bluntly, this has no place in Calamares in my opinion.

kkofler:
I shall also add that building the package as an unprivileged user provides a false sense of security when the package then gets installed as root (so it can run pre-/post-install scripts as root) and the user used for building is even allowed to run `pacman` as root with no password during the whole build.

kennystrawnmusic:
> The code seems to write `/etc/sudoers` in the host system, and executes with `sudo` (oh, wait, not because you want to run as root, but because you want to run as *nobody*) in the target system.

It writes to `sudoers.d`, *not* to sudoers itself. That's because of this: https://wiki.archlinux.org/title/Makepkg#Usage

yay depends on makepkg to do the package building, and makepkg will throw an error if run as root. And that very Arch Wiki page recommends using `nobody` as a workaround to this.

To be fair, that's an easy fix — all you'd need to do is `rm` that `sudoers.d` file after the operation is complete. Adding that in, hold on

kennystrawnmusic:
Updated to delete the `sudoers.d` entry after the operation is complete. That should fix the security hole, because it should go without saying that, unlike `/etc/sudoers` itself, `sudoers.d` entries can be created and deleted at will without affecting the ability for any other users to sudo the way modifying sudoers itself would.

Finally, the reason why I opted to add this by forking the pacman code is because yay is a pacman wrapper — syntactically identical to pacman in terms of how it's invoked. Which, for the record, is exactly why I prefer it over other AUR helpers like pamac and paru: consistency.

kennystrawnmusic:
And the reason why I opted to add this by forking the pacman code is because yay is a pacman wrapper — syntactically identical to pacman in terms of how it's invoked.

kennystrawnmusic:
Well it wasn't working as expected doing it the way I thought initially, so trying a new approach: mounting `/etc/sudoers.d` in the target system as `tmpfs` before writing to it.

kennystrawnmusic:
By doing this, it ensures that the `sudoers.d` entry gets deleted the instant the `umount` module runs.

dalto8:
My personal opinion is this is somewhat strange for a number of reasons.
* `yay` is one of many AUR wrappers, if support is going to be added for an AUR wrapper, why not make it configurable?
* Building AUR packages on the ISO is going to painful for a distro support.  Not only is the process somewhat fragile, memory constraints are going to be a problem.
* The packages module writing to `/etc/sudoers` on the host seems quite odd to me.  Does that work?
* Same for the environment variables you are setting, are those getting passed into the target somehow?

To me, this seems like a support nightmare.  Does this make sense upstream in Calamares or should it be handled as a distro-specific module?

kennystrawnmusic:
> * `yay` is one of many AUR wrappers, if support is going to be added for an AUR wrapper, why not make it configurable?

Because that would necessitate modifying the code at a much deeper level

> * Building AUR packages on the ISO is going to painful for a distro support.  Not only is the process somewhat fragile, memory constraints are going to be a problem.

The only alternative is bundling AUR packages directly on the ISO image which is even worse for licensing reasons in some cases.

> * The packages module writing to `/etc/sudoers` on the host seems quite odd to me.  Does that work?

No it doesn't because it again writes to the sudoers.d include-directory, not to sudoers itself.

> * Same for the environment variables you are setting, are those getting passed into the target somehow?

Only temporarily. They're unset after reboot, and I know this because I've tested it, both in a VM and as a full install.

> To me, this seems like a support nightmare.  Does this make sense upstream in Calamares or should it be handled as a distro-specific module?

Handling it as distro-specific necessitates forking Calamares itself at the moment.

dalto8:
> Because that would necessitate modifying the code at a much deeper level

Not substantially, you only need to add a little bit of config to describe the options needed for each operation

> The only alternative is bundling AUR packages directly on the ISO image which is even worse for licensing reasons in some cases.

The other alternative is providing an easy way for them to be installed post-install.

> No it doesn't because it again writes to the sudoers.d include-directory, not to sudoers itself.

That is an implementation detail that doesn't really change my comment any.  It still seems quite strange.  Don't you need the sudoers drop in on the target?

> Handling it as distro-specific necessitates forking Calamares itself at the moment.

No it doesn't.  You can just make some small changes to your Calamares PKGBUILD.  Either add a custom module or patch/overwrite the existing one.  For a Python module there is really nothing to it other than put the files where they need to be.

kkofler:
> The only alternative is bundling AUR packages directly on the ISO image

Uh, no. You can build the packages you want to install and ship the binary packages in an overlay repository. Then you just set up the binary overlay repository directly on the ISO and you can do the install in Calamares with just `pacman`.

But licensing can be an issue, depending on what you want to ship.

kennystrawnmusic:
> But licensing can be an issue, depending on what you want to ship.

Exactly my point. AUR helper usage is the only way around some license restrictions, like the ones associate with the various Google-encumbered Anbox images in the AUR (my use case) for instance.

demmm:
>Exactly my point. AUR helper usage is the only way around some license restrictions

That is why there is the License module, added exactly for situations like this.  Users will have to agree to the License, otherwise the selected package won't be installed. So, certainly not the case that you can't provide a binary of a restricted package.

kkofler:
Well, even making the user agree to a license does not magically make it legal to distribute things for which you have no distribution license. Android images containing proprietary Google code are one such case.

kennystrawnmusic:
Closing this PR as I’ve opted to use a patch containing these changes along with my PKGBUILD instead. Much easier, more productive solution.

PR #1910, opened 2022-03-24
Enable Replace Partition FS selection

DarkXero-dev:
A valid question was brought to my attention...

"Why when selecting the replace partition in cala does not allow me to change filesystem type and stuck at Distro's default choice?"

Could the FS selection stay activated for that ?


hphilm:
I think we can close this one ...

PR #1918, opened 2022-04-11
[libcalamares] Fix file-reading from target system

vtriolet:
`readTargetFile` was not fully reading files because of an incorrect
EOF check. This could cause /etc/openswap.conf files to be
truncated and hibernation to fail on installed systems.

Here's one external bug report about the issue:
- https://forum.manjaro.org/t/empty-etc-openswap-conf-manjaro-xfce-gnome-21-2-4-5-minimal-220301-linux515-iso-default-install-fde-nohibernation-on-efi-baremetal-or-kvm-boot-device-does-not-exists/106452

### DIFF
--- a/src/libcalamares/utils/CalamaresUtilsSystem.cpp
+++ b/src/libcalamares/utils/CalamaresUtilsSystem.cpp
@@ -169,7 +169,7 @@ System::readTargetFile( const QString& path ) const
 
     QTextStream in( &f );
     QStringList l;
-    while ( !f.atEnd() )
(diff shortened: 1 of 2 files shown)


adriaandegroot:
Thanks! That goes to show how limited unittests (the one-line test was ok) can be misleading :(

vtriolet:
> Thanks! That goes to show how limited unittests (the one-line test was ok) can be misleading :(

Yeah, it is difficult to develop a representative set of tests that ensures correctness :).  Thank you for the review.

PR #1928, opened 2022-04-19
Module to fix unencrypted boot partition configuration

abalfoort:
Module to fix unencrypted /boot partition issues.
Includes configuring previously encrypted partitions that are mounted without formatting first (like /home).

### DIFF
--- a/src/modules/luksboot-fix/luksboot-fix.conf
+++ b/src/modules/luksboot-fix/luksboot-fix.conf
@@ -1,0 +1,38 @@
+# SPDX-FileCopyrightText: no
+# SPDX-License-Identifier: CC0-1.0
+#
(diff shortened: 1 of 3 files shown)


kkofler:
At first glance, I see at least 2 issues with this module (there are probably more though, if I were to look closer):
1. conceptual: a "module to fix …" is always a bad idea. If something is broken, it needs to be fixed in the existing module where it is broken, not worked around in another, new module.
2. implementation: the module uses GTK (through its Python bindings), in Calamares, which is otherwise a Qt application. Bringing up GUI dialogs in Python modules is just generally not supported, and dialogs also do not fit the wizard user interface well. The module should probably be a view step in C++ or QML instead.

adriaandegroot:
Closing this as "superceded by #1931"

PR #1931, opened 2022-04-21
Unencrypted boot

abalfoort:
My [previous pull request](https://github.com/calamares/calamares/pull/1928) had some critique. So, with this pull request I adapted these files to let Calamares support unencrypted /boot partition:

- src/modules/fstab/main.py
- src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp

I am new to C and C++ and I just started recently getting into Calamares. So, be patient.

Feel free to delete the previous pull request (and branch luksboot-fix).

### DIFF
--- a/src/modules/fstab/main.py
+++ b/src/modules/fstab/main.py
@@ -142,21 +142,13 @@ class FstabGenerator(object):
         with open(crypttab_path, "w") as crypttab_file:
             print(CRYPTTAB_HEADER, file=crypttab_file)
 
-            # Check if /boot is unencrypted
-            unencrypted_separate_boot = False
             for partition in self.partitions:
-                if (partition["mountPoint"] == "/boot"
-                    and "luksMapperName" not in partition):
-                        unencrypted_separate_boot = True
-                        break
-
-            for partition in self.partitions:
-                dct = self.generate_crypttab_line_info(partition, unencrypted_separate_boot)
+                dct = self.generate_crypttab_line_info(partition)
 
                 if dct:
                     self.print_crypttab_line(dct, file=crypttab_file)
 
-    def generate_crypttab_line_info(self, partition, unencrypted_separate_boot):
+    def generate_crypttab_line_info(self, partition):
         """ Generates information for each crypttab entry. """
         if "luksMapperName" not in partition or "luksUuid" not in partition:
             return None
@@ -166,13 +158,17 @@ class FstabGenerator(object):
         if not mapper_name or not luks_uuid:
             return None
 
-        # Set crypttab password for partition to none and remove crypttab options
-        # on root partition when /boot is unencrypted
         password = "/crypto_keyfile.bin"
         crypttab_options = self.crypttab_options
-        if partition["mountPoint"] == "/" and unencrypted_separate_boot:
-            password = 'none'
-            crypttab_options = ''
+
+        # Set crypttab password for partition to none and remove crypttab options
+        # on root partition when /boot is unencrypted
+        if partition["mountPoint"] == "/":
+            if any([p["mountPoint"] == "/boot"
+                   and "luksMapperName" not in p
+                   for p in self.partitions]):
+                password = "none"
+                crypttab_options = ""
 
         return dict(
             name=mapper_name,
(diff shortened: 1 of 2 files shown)


src/modules/fstab/main.py
```diff
@@ -142,13 +142,21 @@ def generate_crypttab(self):
         with open(crypttab_path, "w") as crypttab_file:
             print(CRYPTTAB_HEADER, file=crypttab_file)
 
+            # Check if /boot is unencrypted
+            unencrypted_separate_boot = False
```
adriaandegroot: The diff is hard to read here (and that's not your fault). Can we do this in a more Pythonic way? You want to know if there's a `/boot` partition and it is unencrypted. How about `any([p["mountPoint"]=="/boot" and "luksMapperName" not in p for p in self.partitions])` ? That's a one-liner, and using `any()` makes it clear you're interested in finding one, not in the details.

src/modules/fstab/main.py
```diff
@@ -158,11 +166,19 @@ def generate_crypttab_line_info(self, partition):
         if not mapper_name or not luks_uuid:
             return None
 
+        # Set crypttab password for partition to none and remove crypttab options
+        # on root partition when /boot is unencrypted
+        password = "/crypto_keyfile.bin"
+        crypttab_options = self.crypttab_options
+        if partition["mountPoint"] == "/" and unencrypted_separate_boot:
+            password = 'none'
+            crypttab_options = ''
```
adriaandegroot: All the other strings used in Calamares Python modules (rare exceptions where they must contain a `"`) are `"` delimited. For consistency, please change that (but this isn't a blocker).

src/modules/fstab/main.py
```diff
@@ -264,6 +280,10 @@ def generate_fstab_line_info(self, partition):
             options = "subvol={},".format(partition["subvol"]) + options
 
         if has_luks:
+            # Check if user mounted a previously encrypted partition
```
adriaandegroot: From only the context here I don't see the match between the comment and the code that follows. In any case, `has_luks` is set:

```
has_luks = "luksMapperName" in partition
```
.. aha, so this check is for *luksMapperName* being set, but empty. Perhaps move the check earlier in the function?  Before `options` is set, that's where most of the sanity-checks are.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -172,6 +172,22 @@ hasUnencryptedSeparateBoot()
     return false;
 }
 
+static bool
+hasEncryptedRoot()
```
adriaandegroot, line 176: I see you've just copy-pasted an existing function; I'm going to Have Opinions about that, but only after we've merged your work, then I'll hit it with a more-modern-C++ hammer.

adriaandegroot (commented):
This looks pretty good -- it's easy enough to explain. I have some non-blocking comments if you'd like to polish things up a bit. To discuss changes before filing PRs (e.g. to avoid that GTK issue) you might want to join one of the chat channels for Calamares. They're low-traffic.

abalfoort:
I had some problems with git on my end. Solved now but that's why there are several identical commits.

adriaandegroot:
(Hm, maybe I should have squashed that or done it by hand .. doesn't fundamentally matter, we still have a mainline *calamares* branch that contains mostly merges-of-new-things)

abalfoort:
I'd like to know how you want to adapt my solution in LuksBootKeyFileJob.cpp. Are you going to use this branch for it?

abalfoort:
Also, maybe these issues can be closed?
#1311
#1073

adriaandegroot:
You can take a look at recent commits in the *calamares* branch, for instance 1752dd573b36d3cc27c37d05c0124c1d779ffc49 . That's not something I wanted to bother you with if you're a relative newcomer to C++ -- and also I don't have a really strong sense right now of what is the best / right way of doing things.

In my mind, we had two functions -- `hasEncryptedRoot()` and `hasUnencryptedSeparateBoot()` which both do one thing: check all the partitions for one that matches a particular idea; if there is one, return `true`, otherwise `false`. So I did some refactoring, so that there is a handful of one-liner functions. These give us a vocabulary to write down those two important functions in a way that is immediately readable and expresses exactly what they check for. I leave it to the compiler to expand all of the static inline functions to get back to the for-loop version that there was, but for human readers, the code is better.

(I'll add "better" also means, to me, "uses C++ algorithms rather than writing out raw loops" -- that's opinionated, but I feel C++ would be better if we had a richer shared vocabulary that included the standard algorithms, rather than having to fall back every time. See Python for instance: list comprehensions are part of its power, and we can use C++ algorithms to do the same)

friedy10:
How do you enable unencrypted /boot in Calamares configuration?

abalfoort:
> How do you enable unencrypted /boot in Calamares configuration?

You simply create a partition (about 500MB) and mount that as /boot.
And check https://github.com/calamares/calamares/blob/calamares/src/modules/partition/partition.conf

Issue #1531, opened 2020-10-12
Give in-config translations consistent top-level key


adriaandegroot:
Various modules support translations in the configuration files -- e.g. netinstall which can be instantiated multiple times and can have different names set in the sidebar. This works with a `CalamaresUtils::Locale::TranslatedString` (something that you can grep for), loaded from the configuration. The key used for the translation is not consistent: netinstall uses *label* and packagechooser uses *labels* and shellprocess uses *i18n* (since today).

Let's make that consistent: all the translations in configuration files should use top-level key *i18n*.

Existing code for processing the configurations should be modified to complain about the "old" keys, but to use them nonetheless. Mark them with a `TODO` that in Calamares 3.3 they should bail out with the old keys.

adriaandegroot:
@abalfoort 

For 3.3, the idea is to give all the modules *Config* classes a common base class: `Calamares::ModuleSystem::Config`. This lives in `src/libcalamares/modulesystem/`. There's a TODO that mentions consistent i18n for labels / step-names in the header file. Here's a (partial) checklist of what I think should be done:

- [ ] Add a `QString prettyName() const` to the base class; add a `virtual QString basePrettyName() = 0` for the no-special-setting-in-config.
- [ ] Add a `TranslatedString` instance -- like you just did to the *packagechooser* Config -- to the nested  `Private` class.
- [ ] Implement `prettyName()` by checking for the translation, or calling `basePrettyName()`. This looks a lot like your implementation you just did.
- [ ] In `setConfigurationMap()`, look up key *i18n*, look up some suitable subkey -- I suggest *step* -- for the translation of the step name. Store that in the `TranslatedString` instance. Again, this looks just like the code you already wrote.
- [ ] Finally, go through the modules looking for a translated step-name; make necessary config-file and config-schema changes to shuffle the keys around. Don't worry about backwards compatibility: that's why this is a 3.3 ticket.
- [ ] Make the module's `Config` class inherit from the common base class. Remove any `TranslatedString` instance that is there for the step-name. Call the base-class `setConfigurationMap()` at the end of the sub-class implementation.

All of that together makes the labels and the underlying code consistent. There's probably some bits missing: the type of `config()` for view modules probably needs to change as well so callers know they're getting a (sub-class) pointer to the common base class Config.

abalfoort:
Are you sure you meant to send me this message? I was not the one who worked on packagechooser or any translation code (yet).

adriaandegroot:
[...]

PR #1932, opened 2022-04-22
[packagechooserq]: allow changing step name

vlinkz:
Adds a configurable step name to the `packagechooserq` module, similar to the `packagechooser` module.

### DIFF
--- a/src/modules/packagechooser/Config.cpp
+++ b/src/modules/packagechooser/Config.cpp
@@ -238,6 +238,12 @@ Config::setPackageChoice( const QString& packageChoice )
 }
 
 QString
+Config::prettyName() const
+{
+    return m_stepName ? m_stepName->get() : tr( "Packages" );
+}
+
+QString
 Config::prettyStatus() const
 {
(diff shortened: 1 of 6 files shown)


adriaandegroot:
This should be in the `Config` class, so that it is (or can be) immediately shared between *packagechooser* and *packagechooserq*. That just means shuffling some code over to the other class, and making `prettyName()` call `m_config->prettyName()`.

See also #1531, I think it'd be nice to chase some consistency here. I think using *i18n* as top-level key (and then *steps* as the translated key, as you've done) is better naming in the long run.

adriaandegroot:
(Don't worry about the CI failure -- that's still fallout from GH / git being stupid in the face of a security update)

vlinkz:
Ok, I added the `prettyName()` function to `packagechooser/Config.cpp` and made both `PackageChooserViewStep` and `PackageChooserQmlViewStep` call that function. Seems to work as intended when I built it.

adriaandegroot:
It's weird .. 0eae72e10f adds relevant config-keys but doesn't seem to have an implementation in the *packagechooser* module. So this resolves that issue. If you want to continue to chase this, for 3.3 we could land a PR for #1531.

Issue #1922, opened 2022-04-13
About button on Welcome page confusing


jtojnar:
**Describe the bug**

The welcome page shows several buttons referring to the OS distribution being installed. However, when I click the “About” button, I am shown info about Calamares, not the OS.

https://github.com/calamares/calamares/blob/baf8ef2c33879511bcc168cae03ab0c2aee1f505/src/modules/welcomeq/welcomeq.qml#L66

And to make matters worse, languages like Czech translate this as “About the project”, which makes sense in regular applications but not so much in an installer, where it is not clear what the project is.

**To Reproduce**

1. Start Calamares
2. Click the “About” button
3. See that “About Calamares installer” pops up

**Expected behavior**
The button should be labelled “About the installer”, maybe moved away from the distribution-related buttons. Something like the following might work:

![Mockup of proposed Welcome page with Powered by Calamares moved away from the rest of the buttons](https://user-images.githubusercontent.com/705123/163074627-6f7b0595-bc2f-48a8-b38b-fe4af448e462.png)

**Screenshots and Logs**

![Current Welcome page](https://user-images.githubusercontent.com/705123/163074898-a3e5baa9-b532-4f93-80c2-b5f25aeeefa2.png)

demmm:
You can wait for upstream to decide to implement this feature (or decide not to), but you can already do this by using welcomeq instead of the welcome module.  Any of the QML modules will read complete custom version of a .qml file when placed in the branding directory of your distribution.
https://github.com/calamares/calamares/blob/calamares/src/modules/welcomeq/welcomeq.qml#L54

adriaandegroot:
I kinda like the "powered by Calamares" label; in terms of UI elements, though, that's in the "navigation bar" area, so it might be tricky to implement putting it exactly **there**. (There is a QML navbar which is easier to edit without re-compiling, and a Widget navbar in `CalamaresWindow.cpp`, function `getWidgetNavigation()`) Doable, though: I'd suggest showing that button only on the first step (maybe the last one, too), which is something you can arrange by connecting to the view manager's *currentStepChanged* signal.

You could also leave it inside the welcome page area (e.g. not the nav bar or the progress list).

For non-coding changes to the UI: the welcome page can be done in QML (as @demmm points out) or in Widgets, but the widgets part is generated from a Qt Designer .ui file. You could fiddle a bit with the UI with designer and see what you can achieve. The relevant file is `src/modules/welcome/WelcomePage.ui`. After editing needs a recompile.

[...]

PR #1934, opened 2022-04-23
[libcalamaresui] Provide a new API to add widgets to navigation panel

LordTermor:
* Adds a new ViewStep property for exposing it's extra navigation widget to ViewManager
* Connected CalamaresWindow to ViewManager's signal to update navbar layout on widget change
* `nullptr` means no widget is exposed, this is default behavior
* As an example About button in [welcome] was moved to navigation area module as https://github.com/calamares/calamares/issues/1922 suggests

image

### DIFF
--- a/src/calamares/CalamaresWindow.cpp
+++ b/src/calamares/CalamaresWindow.cpp
@@ -166,10 +166,6 @@ getWidgetNavigation( Calamares::DebugWindowManager*,
                       bottomLayout,
                       [ bottomLayout ]( QWidget* wgt )
                       {
-                          if ( !bottomLayout )
-                          {
-                              return;
-                          }
                           auto prevWgt = bottomLayout->parentWidget()->findChild< QWidget* >( "view-custom-item" );
 
                           if ( prevWgt )
--- a/src/calamares/CalamaresWindow.cpp
+++ b/src/calamares/CalamaresWindow.cpp
@@ -161,6 +161,31 @@ getWidgetNavigation( Calamares::DebugWindowManager*,
     QBoxLayout* bottomLayout = new QHBoxLayout;
     bottomLayout->addStretch();
 
+    QObject::connect( viewManager,
+                      &Calamares::ViewManager::navigationPanelWidgetChanged,
+                      bottomLayout,
+                      [ bottomLayout ]( QWidget* wgt )
+                      {
+                          if ( !bottomLayout )
+                          {
+                              return;
+                          }
+                          auto prevWgt = bottomLayout->parentWidget()->findChild< QWidget* >( "view-custom-item" );
+
+                          if ( prevWgt )
+                          {
+                              prevWgt->setObjectName( "" );
+                              prevWgt->setParent( nullptr );
+                              bottomLayout->removeWidget( prevWgt );
+                          }
+
+                          if ( wgt )
+                          {
+                              wgt->setObjectName( "view-custom-item" );
+                              bottomLayout->insertWidget( 0, wgt );
+                          }
+                      } );
+
     // Create buttons and sets an initial icon; the icons may change
     {
         auto* back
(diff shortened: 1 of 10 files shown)


LordTermor:
It's drafted because I'm not sure if it's ok implementation-wise and if it is I'll need to add some comments. Also maybe there will be some more fitting function names.

src/calamares/CalamaresWindow.cpp
```diff
@@ -161,6 +161,31 @@ getWidgetNavigation( Calamares::DebugWindowManager*,
     QBoxLayout* bottomLayout = new QHBoxLayout;
     bottomLayout->addStretch();
 
+    QObject::connect( viewManager,
+                      &Calamares::ViewManager::navigationPanelWidgetChanged,
+                      bottomLayout,
+                      [ bottomLayout ]( QWidget* wgt )
+                      {
+                          if ( !bottomLayout )
```
adriaandegroot: I'm not sure what this `if()` is supposed to protect against: it's a copy of the layout, which is also the object to which the signal is sent .. how is this ever going to be *nullptr* without lots of other UB happening already?

src/calamares/CalamaresWindow.cpp
```diff
@@ -161,6 +161,31 @@ getWidgetNavigation( Calamares::DebugWindowManager*,
...
+                          {
+                              return;
+                          }
+                          auto prevWgt = bottomLayout->parentWidget()->findChild< QWidget* >( "view-custom-item" );
```
adriaandegroot: prefer `auto*` to emphasize it's a pointer, and longer names: `previousCustomWidget` or so.

src/libcalamaresui/ViewManager.h
```diff
@@ -232,6 +232,8 @@ public Q_SLOTS:
     void quitVisibleChanged( bool ) const;
     void quitTooltipChanged( QString ) const;
 
+    void navigationPanelWidgetChanged( QWidget* );
```
adriaandegroot, line 236: Needs API dox.

src/calamares/CalamaresWindow.cpp
```diff
@@ -161,6 +161,31 @@ getWidgetNavigation( Calamares::DebugWindowManager*,
...
+
+                          if ( prevWgt )
+                          {
+                              prevWgt->setObjectName( "" );
```
adriaandegroot: `QString()` instead of `""`

src/calamares/CalamaresWindow.cpp
```diff
@@ -161,6 +161,31 @@ getWidgetNavigation( Calamares::DebugWindowManager*,
...
+                              prevWgt->setParent( nullptr );
+                              bottomLayout->removeWidget( prevWgt );
+                          }
+
+                          if ( wgt )
+                          {
+                              wgt->setObjectName( "view-custom-item" );
```
adriaandegroot: Please use `QStringLiteral()` where the API expects a QString.

src/libcalamaresui/viewpages/ViewStep.h
```diff
@@ -83,6 +83,8 @@ class UIDLLEXPORT ViewStep : public QObject
      */
     virtual QWidget* widget() = 0;
 
+    virtual QWidget* navigationPanelWidget();
```
adriaandegroot, line 94: Needs API dox: "a view module can return a widget that should be inserted .."

adriaandegroot:
After writing all this, I realise that it puts a new imbalance between QML modules and Widgets modules, one that I don't think we should support (in this form). A distro should be free to pick nav and progress panels: what if we mix a QML nav panel into the whole? These Widget methods will not display on the QML nav panel, no matter what we do. A QML-based ViewModule has no way of providing a Widget to the nav panel (well, it does, by returning a Widget, but that's a kind of weird thing to do if you're focused on QML).

So perhaps we need to take a step back and focus on "what is this extra area for?"

demmm:
I am working on getting this in the calamaras-navigation.qml, see:
Screenshot_20220424_213320

So maybe it isn't an issue to use this PR for the widgets only?

LordTermor:
> I am working on getting this in the calamaras-navigation.qml, see: Screenshot_20220424_213320
> 
> So maybe it isn't an issue to use this PR for the widgets only?

I actually extended the support to qml using `QQuickPaintedItem` already. Will push when will finish polishing.

image

LordTermor:
Closing, need to look into a better implementation that will support both QML and Widgets.

PR #1948, opened 2022-05-13
[partition] Don't offer whole disk encryption when zfs is selected

dalto8:
Although zfs encryption works with manual partitioning, it has not been implemented for whole disk encryption via erase disk.

Until such time as it can be implemented, hide the encryption option when zfs is selected.

### DIFF
--- a/src/modules/partition/gui/ChoicePage.cpp
+++ b/src/modules/partition/gui/ChoicePage.cpp
@@ -457,7 +457,6 @@ ChoicePage::continueApplyDeviceChoice()
     Q_EMIT deviceChosen();
 }
 
-
 void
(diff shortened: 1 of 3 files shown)


adriaandegroot:
There seems to be an unrelated and not-used change in this PR? What's that new `setChecked()` function for?

dalto8:
> What's that new setChecked() function for?

Oops, let me remove that.  The original purpose was to make it uncheck the encryption checkbox if you checked it before switching to zfs.  However, that caused Calamares to crash on some unrelated code and I didn't have time to track it all the way through so I removed it.

It seemed like an edge case and since this currently causes breakage I thought it was more important to get it fixed.

PR #1949, opened 2022-05-13
Remove unneeded emits

dalto8:
`beginResetModel()` and `endResetModel()` are not signals.  They are protected functions that emit signals.

There is no reason to emit() them.


adriaandegroot:
There's never a requirement to emit (or `Q_EMIT` in modern code so it doesn't stomp on a potential keyword), it's just a marker for the reader. I'm merging this although it's going to cause annoying conflicts in the work-3.3 branch.

dalto8:
I noticed it because clang-tidy complained about it.

>  I'm merging this although it's going to cause annoying conflicts in the work-3.3 branch.

It is too late now but it probably could have been merged to 3.3 instead.

Issue #1935, opened 2022-04-23
Calamares crashes when a previously encrypted partition is mounted


abalfoort:
**Describe the bug**
If you manually partition your system and you mount a previously encrypted partition without formatting it first, Calamares will error out just before the end of the installation process. This is usually done with the /home partition which you would like to preserve when upgrading the OS. The user is not able to provide an existing passphrase for the partition.

**To Reproduce**
1. Select manual partitioning
2. Click on the previously encrypted /home partition and mount as /home
3. Start the installation
4. Error: Could not configure LUKS key file on partition /dev/sda5.

**Expected behavior**
When mounting a previously encrypted partition, ask for an existing passphrase. Check by mapping the partition and use that for crypttab and fstab. Add the luks key to the key file.

**Screenshots and Logs**
Setup screenshot:
![Screenshot_20220423_150338](https://user-images.githubusercontent.com/63945418/164896094-f7a791e3-678d-4385-b3f6-96b01d70ac12.jpeg)

Error screenshot:
![Screenshot_20220423_150831](https://user-images.githubusercontent.com/63945418/164896104-ace24b5f-acdc-45dc-a092-c6a80d820fdd.jpeg)

Log:
[session.log](https://github.com/calamares/calamares/files/8546997/session.log)

**Additional context**
None

abalfoort:
At first glance this should only be possible for /home and manual mount points, or am I missing others?

abalfoort:
Can somebody explain to me how to save the luksMapperName for a specific partition in globalStorage (in C++)?

abalfoort:
[Also posted on Matrix calamares channel]

I'm working on this issue but because of my limited knowledge of C++ and the project I currently hit a wall. In /modules/partition/gui/EditExistingPartitionDialog.cpp I need to add a luksPassphrase for the current (and previously encrypted) partition. There is no partitions list in globalStorage at this point (I checked). So, how do I proceed?

adriaandegroot:
The dialog is setting up information about a specific partition, so you would add a text box for the passphrase, then store it in the partition data. The dialog knows which partition it is. `PartitionInfo` and `KPMHelpers` are the things that modify the partition; see `CreatePartitionDialog` for more examples -- there is a `createNewEncryptedPartition()` function. If there are no suitable helpers for setting-passphrase without creating a new partition, you'll have to add them.

abalfoort:
I got something like this:
```
[...]

PR #1952, opened 2022-05-16
#1935 - Reuse luks partitions

abalfoort:
This should fix issue #1935.

When editing an existing luks partition the user can provide the corresponding passphrase limited by the mount point chosen: custom or /home.

Adapting the existing partition objects is done in src/modules/partition/core/KPMHelpers.cpp - updateLuksDevice.

updateLuksDevice is called EditExistingPartitionDialog.cpp and when updateLuksDevice fails, a message is shown.

EncryptWidget was adapted for this change.


abalfoort:
Unfortunately the branch I created for this change was not created.
Closing to try again.

abalfoort:
I don't know how to fix this failed check:
fatal: Remote branch 1952/merge not found in upstream origin

If anyone can help me with that, please.

adriaandegroot:
The CI failure is a problem of the CI action, not a problem of your pull request.

> For context: GitHub has a "checkout" action that does a gazillion things, and is generic, and broke when git released a security fix. That was later fixed in the GH action, but any *other* use of git in CI was still broken. So Calamares grew a really **simple** "checkout" action, but that doesn't handle branches-from-forks at all. If you want to get lost in the weeds of GH CI -- I'd suggest against it, it's just another bunch of vendor lock-in -- take a look at https://github.com/calamares/actions

adriaandegroot:
Some retroactive comments:

- please *do* apply the common Calamares formatting tool to the source before committing. Feel free to format differently while editing, whatever is your jam, but `ci/calamaresstyle` on the sources should happen always. Now I did it after merging, so we have some ping-ponging of `KPMHelpers.cpp` which is indented and then outdented by 4 spaces.
- `int updateLuksDevice( Partition* partition, const QString& passphrase )` is missing API documentation (what does this do? in what sense "update"?), and the return value is mysterious; what does it indicate? what is success? are there different failure modes? Prefer an enum class so that we can use the compiler to help us.

abalfoort:
Sorry about that. I will  run `ci/calamaresstyle` in the future. I had to compile clang-format 12 from source because it is not available in Debian Bullseye. For future reference:

```
git clone https://github.com/llvm/llvm-project --branch release/12.x --depth 1
cd llvm-project
mkdir build
cd build
cmake ../llvm -DCMAKE_BUILD_TYPE=Release -G"Unix Makefiles" -DLLVM_ENABLE_PROJECTS=clang -DCMAKE_INSTALL_PREFIX=/usr/local/clang-format-12
make -j$(nproc) clang-format
sudo make install-clang-format
```
and add /usr/local/clang-format-12/bin to PATH.

I will create an enum for the return values.
How do you prefer your documentation (where can I find the documentation on that)?

PR #1955, opened 2022-05-19
Kpmhelpers updateLuksDevice name change and enum

abalfoort:
Rename updateLuksDevice method to savePassphrase to better reflect its purpose.

Create an enum with return values for savePassphrase.

### DIFF
--- a/src/modules/partition/core/KPMHelpers.cpp
+++ b/src/modules/partition/core/KPMHelpers.cpp
@@ -20,9 +20,9 @@
 #include <kpmcore/backend/corebackendmanager.h>
 #include <kpmcore/core/device.h>
 #include <kpmcore/core/partition.h>
+#include <kpmcore/fs/filesystem.h>
 #include <kpmcore/fs/filesystemfactory.h>
 #include <kpmcore/fs/luks.h>
-#include <kpmcore/fs/filesystem.h>
 #include <kpmcore/util/externalcommand.h>
 
 using CalamaresUtils::Partition::PartitionIterator;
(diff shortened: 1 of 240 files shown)


src/modules/partition/core/KPMHelpers.h
```diff
@@ -43,6 +43,17 @@ class PartitionRole;
 namespace KPMHelpers
 {
 
+enum SavePassphraseValue
```
adriaandegroot, line 46: If you make this `enum class` then in C++ using the class-name becomes mandatory, and you avoid the implicit-conversion-to-int of a C-style enum. That is useful, because it keeps names more widely separated (no chance of collision with any **other** `NoError`).

src/modules/partition/gui/EditExistingPartitionDialog.cpp
```diff
@@ -254,8 +254,8 @@ EditExistingPartitionDialog::applyChanges( PartitionCoreModule* core )
         const QString passphrase = m_ui->encryptWidget->passphrase();
         if ( !passphrase.isEmpty() )
         {
-            int retCode = KPMHelpers::updateLuksDevice( m_partition, passphrase );
-            if ( retCode != 0 )
+            KPMHelpers::SavePassphraseValue ret = KPMHelpers::savePassphrase( m_partition, passphrase );
```
adriaandegroot, line 257: You could use `auto` here; that's acceptable in Calamares code, although it's entirely up to you. Since `ret` isn't used in the rest of this block, you could also put the function-call in the `if()` comparison directly.

src/modules/partition/core/KPMHelpers.cpp
```diff
@@ -167,24 +167,27 @@ testPassphrase( FS::luks* fs, const QString& deviceNode, const QString& passphra
 }
 #endif
 
-// Adapted from luks cryptOpen which always opens a dialog to ask for a passphrase
-int
-updateLuksDevice( Partition* partition, const QString& passphrase )
+// Adapted from src/fs/luks.cpp cryptOpen which always opens a dialog to ask for a passphrase
+/**
```
adriaandegroot, line 171: This documentation style doesn't match the rest of Calamares; this looks an awful lot like Python-style docs. I could kind of live with that, but this looks a lot like docs-to-tick-the-docs-box, not docs-that-explain-the-thing. Having empty `@param` lines just doesn't help at all. Something like

```
/** @brief Saves an existing @p passphrase for a previously encrypted @p partition
 *
 * Tries to apply the passphrase to the partition; this checks if the
 * @p partition is one that can have a passphrase applied, and
 * runs `cryptsetup` to check that the passphrase actually works
 * for the partition. Returns `NoError` on success, or an explanatory
 * other value if it fails.
 */
```

adriaandegroot (changes requested):
Please squash down and drop the merge-from-calamares things. If you're developing a feature in a branch, don't merge into the branch: leave the branch "clean" with just the changes you need to make for your feature. That makes the sequence of commits clearer.

abalfoort:
I will have to study the "why" of your comments (steep learning curve here, that's good).

If you say "leave the branch "clean" with just the changes you need to make for your feature." Do you mean to create a new branch as described [here](https://stackoverflow.com/questions/34100048/create-empty-branch-on-github) with the use of  "git switch --orphan <new branch>"?

Do you know of a way to remove all unchanged files from a branch leaving only the tracked files in that branch (to clean up the existing branches)?

adriaandegroot:
I'm going to massage this PR with a "counter-proposal" and document what I did, because answering your questions -- and they are good questions! -- off the top of my head will not work.

adriaandegroot:
- I followed the manual instructions to fetch your branch (of this PR), then used `qgit` to take a look at what actually happens in the branch
- Commit 8613dcfe59acf351bff413030134a4d268195061 is the one that does the work: the rest is all merges and catching-up, so I cherry-picked that one commit into the *calamares* branch

When working on a branch in Calamares, we try -- there are exceptions -- to follow a "git alligator" workflow, which is described [here on my blog](https://euroquis.nl/blabla/2019/08/09/git-alligator.html). It is a lot like a rebase workflow, except it preserves a merge-commit as a kind of marker for "this was a thing". There's no point in merging *calamares* or another development branch into your branch; exceptions are only the really-really-long lived branches, like *work-3.3*. That's going to change Real Soon, though.

As for cleaning up a branch, if you don't merge, then `git rebase -i` is really the only tool you need, because you can squash, reorder, and rename commits. I used it after cherry-picking to adjust the commit message a little. (After a rebase, you may need to `git push --force` onto a branch to make it update, but since that's a work-branch in a fork, no harm is done)

adriaandegroot:
For comments and API-documentation:

- documentation goes in the headers, not in the implementation, for functions and methods that are intended to be used by another translation unit (e.g. functions in `KPMHelpers` have their documentation in the `.h` file, because they are used from other files in the *partition* module)
- .. an exception is `static` helper methods in an implementation. Ideally they would have sufficiently meaningful names to not need extra documentation, but you can see e.g. `welcome/Config.cpp` and the method `jobOrBrandingSetting()` for an implementation detail that needs explanation.

In Calamares, we mostly try to make new code look like older code; that's mentioned in passing in the `CONTRIBUTING` file, but not really-really spelled out. There's no talk about documentation style, for instance.

abalfoort:
Thanks for this explanation. I will use --no-ff with my future merges. Funny that it's not default but ff is.

You've closed this PR but some issues were still open. Do you want me to create a new branch for those changes?

kkofler:
> Funny that it's not default but ff is.

It is all a matter of taste. Some people (like me, in my projects) like linear histories. But the merge commits do have their merits, especially for larger merges consisting of many individual (but related) commits.

adriaandegroot:
> Thanks for this explanation. I will use --no-ff with my future merges. Funny that it's not default but ff is.

It's a project-specific taste. For PRs that go through GitHub, a merge commit is added when the PR is merged, so the whole git-alligator idea happens automatically. Recent history is a bit convoluted in Calamares, but you could look at PR #1923 , with commits cc3ab110eed1f9762e0f264f065c92ceccfba6c9 (actual work) and a67adf77025a0a8ba6365e48055f50d72113bcae (merge). The only time someone would use `--no-ff` is when merging things by hand in the git repo; something I do when merging my own local branches.

> You've closed this PR but some issues were still open. Do you want me to create a new branch for those changes?

Yes, please file issues and separate PRs as you notice problems / make more changes.

PR #1958, opened 2022-05-20
#1818 tmpfs on tmp

### LINKED ISSUE #1818: Lubuntu + Calamares mount tmpfs on /tmp
Hi, 

I've reported this bug to Ubuntu/Lubuntu some time ago, https://bugs.launchpad.net/ubuntu/+source/calamares/+bug/1888526 but they see it rather as a calamares specific bug. Still exists in Lubuntu 21.10. 

**Describe the bug**

After installing Lubuntu Desktop (which uses Calamares), a tmpfs is mounted on /tmp. 

This is not just a violation of Linux standards, it also breaks lots of functionalities, since it limits the space on /tmp, consumes RAM, and other problems. This really breaks many functions of a running Linux system.  

**To Reproduce**

Install Lubuntu and do a df. 

**Expected behavior**

Just nothing mounted on /tmp, unless manually chosen by admin during installation time.

abalfoort:
This fixes #1818 

Configure mounting /tmp as tmpfs in fstab can be configured in fstab.conf.
Old behavior (mounting /tmp as tmpfs) is still default if not configured differently in fstab.conf.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,1 +1,5 @@
+#   SPDX-FileCopyrightText: no
+#   SPDX-License-Identifier: CC0-1.0
+#
 18fef8dfe5d926ec0bc979562553adf4db8db2e9
+874a0c1f38b0da4e5bc83083b13a63b1c7eed935
--- a/CHANGES-3.2
+++ b/CHANGES-3.2
@@ -8,14 +8,35 @@ contributors are listed. Note that Calamares does not have a historical
 changelog -- this log starts with version 3.2.0. The release notes on the
 website will have to do for older versions.
 
-# 3.2.57 (unreleased) #
+# 3.2.58 (unreleased) #
+
+This release contains contributions from (alphabetically by first name):
+ - Enrique Medina Gremaldos
+
+## Core ##
+ - Internal improvements to translations-setup means that Catalan (in the
+   Valencian dialect), Occitan (Lenga d'Oc) and Serbian (in Latin script)
+   are all better supported. Thanks Enrique.
+
+## Modules ##
+ - *users* module now has a structured *user* key with settings specific
+   to the user (shell, in particular). This maintains backwards compatibility
+   with the *userShell* key.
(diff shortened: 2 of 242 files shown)


adriaandegroot:
Please squash these PR's into something that follows the regular coding style and commit style. You can use `git rebase -i` to go through the commits to tidy them up.

adriaandegroot:
The **code** is good, and I appreciate your attention to getting defaults and not crashing on missing configuration keys (something other Python modules do a lot). Also updating the schema, thanks.

abalfoort:
Could you eleborate on: "squash these PR's"?
I don't know how to tidy up the PRs (using git rebase -i).

adriaandegroot:
To tidy this up, I started by following the manual instructions for merging this branch. That leads to a situation that looks like this (in *qgit*, my preferred git-history-viewer; it looks slightly different in *gitk*).

image

The red nodes are your branch, which are mostly not needed: the "Configure.." and "Remove print.." commits are the useful ones. So I run `git rebase -i calamares` to re-do the branch on top of the *calamares* branch. This puts me in an editor where I can interactively edit the history:

image

I can replace the word *pick* there by a letter -- documentation is shown in the edit window -- to change the history. I delete the first two lines, since those are merges / noise. Then "Fix unencrypted.." gets an `r` because I want to edit the commit message slightly, and the last one gets an `s` to squash it with the previous commit (and I'll edit the commit message).

image

When squashing, you get all the commit messages together, and then you can edit all you like:

image

abalfoort:
Just a recap to see if I get it now:

1. Rebase with calamares: git rebase -i calamares.
2. Look for first commit you want to keep and label it "r".
3. Remove commits above that one.
4. The commit below the "r" labeled one does not change and keeps label "pick".
5. Anything below that is squased by labelling them with "s".

Correct?

kkofler:
For what it's worth, on Fedora, `/tmp` is on `tmpfs` by default without any `/etc/fstab` entry, they decided to set this up with a native systemd unit instead.

Issue #1954, opened 2022-05-19
Idea/Discussion: Porting partition module to QML


AIIX:
#### **Is your feature request related to a problem? Please describe.**

Hi, I have been working on customising Calamares for the Plasma Bigscreen project which requires using QML modules over Qt Widget modules to enable navigation via Arrow Keys and USB connected remotes, Currently the only module that seems to be lacking a QML equivalent seems to be the main Partition Module. 

#### **Describe the solution you'd like**

I have been working on porting the Partition Module to QML, progressing slowly as the module is quite complex and has nested functionality embedded into the gui widgets logic. 

My current approach to porting the module to QML has been to test and study how the current widget UI works and how it connects to its widgets backend, then additionally testing how much of the widget based functionality and data models can be accessed in QML, if if not creating proxy functionality that can supply the required data to QML objects.

#### How the partitionq module is designed:

- Follows a sub page design where the main partitionq qml provides a loader object to load the subpages of the partition module depending on the current process flow.
- The functionality on the cpp backend is divided as per the subpages, so a sub page called ChoosePage.qml will have a cpp backend class called ChoicePageQml that will handle all the functionality within the context of that page. Follows the similar approach to the widget module except it should house all the functions required by the page in the class itself.

#### Where the above port is at currently:

- Initializing the Partition Core Module
- Displaying the Initial install choose page with a working device selector combo box
- Displaying the install choices supported by the device
- Displaying swap choices supported when selecting the "Erase" install option
- Encryption selection and setting the encryption passphrase

#### To Do:

- Everything else, this is not a functional port yet

#### Some road blocks I have come across till now and possible solutions I see to them:

- Tree View models: The PartitionModel currently provides a tree view interface for Qt Widgets, QML Quick 2 does not have any native support for displaying Tree Views as such, Possible Solution: Add a proxy model to partitionq to flatten the Tree structure to be able to use it in QML listview or repeaters

- Nested functionality of install options: Options such as Install Alongside and Replace partition depend on accessing PartitionSplitterWidget, Possible Solution: Copy the minimal needed functionality from the widget UI to perform the same steps in the partitionq module, this might also be applicable to more parts of the module because of the nested functionality but the idea is probably going to be the same is try copying the minimal amount of functionality needed for it work with the QML port

demmm:
Thanks for posting this.  I have been working off & on for about a year to port the last module used in KaOS not ready for QML, which is ofc the partition module (good to hear all other modules you need are ported).
[...]

PR #1960, opened 2022-05-23
Initial upload of WIP partitionq module

AIIX:
Add support for WIP partitionq module, All details at https://github.com/calamares/calamares/issues/1954

This request provides the following:
- Initializing the Partition Core Module
- Displaying the Initial install choose page with a working device selector combo box
- Displaying the install choices supported by the device
- Displaying swap choices supported when selecting the "Erase" install option
- Encryption selection and setting the encryption passphrase
- Flat partition list mode derived from partition model  (to be use able with qml list, repeater and combo box components)


src/modules/partitionq/CMakeLists.txt
```diff
@@ -0,0 +1,107 @@
+# === This file is part of Calamares - <https://calamares.io> ===
+#
+#   SPDX-FileCopyrightText: 2020 Adriaan de Groot <groot@kde.org>
+#   SPDX-License-Identifier: BSD-2-Clause
+#
+if( NOT WITH_QML )
+    calamares_skip_module( "localeq (QML is not supported in this build)" )
+    return()
```
demmm, line 9: Seems a copy/paste error, should be partitionq, not localeq, copyright header seems to have wrong name, certainly wrong year

src/modules/partitionq/partitionq.conf
```diff
@@ -0,0 +1,255 @@
...
+# value is used instead. It is still a problem if there is no required
+# size set at all, and the replace and resize options will not be offered
+# if no required size is set.
+#
+# The value is in Gibibytes (GiB).
+#
+# BIG FAT WARNING: except for OEM-phase-0 use, you should be using
+#                  the welcome module, **and** configure this value in
+#                  `welcome.conf`, not here.
+# requiredStorage: 3.5
```
demmm: Usually the .conf for a QML references the corresponding widgets .conf, only adds QML specific settings, not duplicate all instructions

AIIX: Fixed in latest commit, have removed all the partitionq.conf file settings, but left the file in there with a to do message, as I am not sure yet what settings actually need to be in QML module itself, will remove it if no settings live in qml in the future

src/modules/partitionq/CMakeLists.txt
AIIX: Fixed in latest commit

Issue #1643, opened 2021-02-19
Allow to select between luks and luks2 when creating encrypted partition


Bleuzen:
**Is your feature request related to a problem? Please describe.**
Yes, I want to use luks2 for encryption. This is currently not possible as any encrypted partition created with Calamares ends up being luks1.

**Describe the solution you'd like**
Add a simple checkbox in the manual partition creation dialog for "Use luks2":
![image](https://user-images.githubusercontent.com/12885163/108502204-bae13a80-72b2-11eb-9195-3654dec3dcdb.png)

Also display a warning message like this when the user first clicks this checkbox:

**"WARNING: Luks2 is currently not fully supported by grub. Only use this option for pure data partitions or if you are using unencrypted /boot, otherwise your system won't boot!"**

**Describe alternatives you've considered**
For implementation of this feature:
I don't think there would be a much better alternative to implement this.

To achieve a setup with luks2 partitions now:
The only option is to not use Calamares.

**Additional context**
Nobody should be forced to use an old technology if there is no reason to.
Users should be able to enjoy the benefits of luks2 on any partition which does not hold encrypted /boot.
This is useful for example:
- for data partitions (like /home and other disks)
- when using unencrypted /boot
- when using separate encrypted /boot and encrypted root

Therefore it should be possible to select which partitions really still need luks1 and which should be luks2.

philmmanjaro:
See https://github.com/calamares/calamares/issues/1099

Bleuzen:
Hi @philmmanjaro , the issue you linked is about the changed / current default of cryptsetup to use luks2.
However, my feature request here has nothing to do with the default.
This is about giving the users an option to manually select the luks version.

adriaandegroot:
There are three parts to this request:
- luks2 needs to work **at all**; the issue Phil refers to discusses that and it's not clear that all the parts are in place to do that yet; this will also depend on the distro and the other parts of that stack that it select
- calamares needs to be able to put the luks2 parts together; that is something it probably does **not** do right now, because of the above^^ (although there is code that checks for / does things with Luks2; it requires KPMCore 4 which most distro's have now)
- The UI needs to be adjusted to pick one or the other (or maybe just "switch to luks2 by default" -- once the moving parts are in place, the selection is not interesting since, as you put it, there's not much reason to stick with outdated technology)

Bleuzen:
@adriaandegroot 
> luks2 needs to work **at all** [...]

[...]

PR #1964, opened 2022-05-25
WIP: Add luks generation selection

S0PEX:
Hey,

as described in https://github.com/calamares/calamares/issues/1643 the installer automatically creates `luks1` partitions, if system encryption is enabled. This is currently done as `grub` only supports `luks2 + pbkdf2`.
Other bootloaders, i.e systemdboot and lilo, do support `luks2 + argon2` and are able to **fully** handle `luks2`.
With this PR an option to manually configure `cryptsetup --type` argument and therefore enable the user to manually use `luks2` is added.

I have already done the necessary modifications to `partition/Config` but haven't finished to actually pass the selected option down to `FileSystemFactory::create`. https://github.com/calamares/calamares/blob/72f25f24ef8c0969f97759338ac0641fa00f8add/src/modules/partition/core/KPMHelpers.cpp#L91

As of now I am not sure if all changes, i.e `luksGenerationChoiceChanged`, are really required but I still included them. Secondly I am currently not sure how the whole qt thing is working, that's why I have updated the signature of `updateGlobalStorage( Config::InstallChoice installChoice, Config::SwapChoice swapChoice, Config::LuksGenerationChoice luksGenerationChoice )` but haven't used the argument there.
Should I just add `luks generation` to `partitionChoices` as it is somehow related to partitioning?

Thanks.

Regards,
Artur

### DIFF
--- a/src/modules/partition/Config.cpp
+++ b/src/modules/partition/Config.cpp
@@ -46,7 +46,7 @@ Config::swapChoiceNames()
     return names;
 }
 
-const NamedEnumTable< Config::LuksGeneration >&
+const NamedEnumTable< Config::LuksGenerationChoice >&
 Config::luksGenerationChoiceNames()
 {
     static const NamedEnumTable< LuksGenerationChoice > names { { QStringLiteral( "luks1" ), LuksGenerationChoice::Luks1 },
@@ -193,7 +193,7 @@ Config::setInstallChoice( InstallChoice c )
     {
         m_installChoice = c;
         Q_EMIT installChoiceChanged( c );
-        ::updateGlobalStorage( c, m_swapChoice );
+        ::updateGlobalStorage( c, m_swapChoice, m_luksGenerationChoice );
     }
 }
 
@@ -215,17 +215,17 @@ Config::setSwapChoice( Config::SwapChoice c )
     {
         m_swapChoice = c;
         Q_EMIT swapChoiceChanged( c );
-        ::updateGlobalStorage( m_installChoice, c );
+        ::updateGlobalStorage( m_installChoice, c, m_luksGenerationChoice );
     }
 }
 
 void
 Config::setLuksGenerationChoice( int c )
 {
-    if ( ( c < LuksGenerationChoice::Luks1 ) || ( c > LuksGenerationChoice::Luks2 ) )
+    if ( ( c != LuksGenerationChoice::Luks1 ) && ( c != LuksGenerationChoice::Luks2 ) )
     {
         cWarning() << "Invalid luks generation choice (int)" << c;
-        c = SwapChoice::Luks1;
+        c = LuksGenerationChoice::Luks1;
     }
     setLuksGenerationChoice( static_cast< LuksGenerationChoice >( c ) );
 }
--- a/src/modules/partition/Config.cpp
+++ b/src/modules/partition/Config.cpp
@@ -46,6 +46,15 @@ Config::swapChoiceNames()
     return names;
 }
 
+const NamedEnumTable< Config::LuksGeneration >&
+Config::luksGenerationChoiceNames()
+{
+    static const NamedEnumTable< LuksGenerationChoice > names { { QStringLiteral( "luks1" ), LuksGenerationChoice::Luks1 },
+                                                      { QStringLiteral( "luks2" ), LuksGenerationChoice::Luks2 } };
+
+    return names;
+}
+
 Config::SwapChoice
 pickOne( const Config::SwapChoiceSet& s )
 {
@@ -211,6 +220,29 @@ Config::setSwapChoice( Config::SwapChoice c )
 }
 
 void
+Config::setLuksGenerationChoice( int c )
+{
+    if ( ( c < LuksGenerationChoice::Luks1 ) || ( c > LuksGenerationChoice::Luks2 ) )
+    {
+        cWarning() << "Invalid luks generation choice (int)" << c;
+        c = SwapChoice::Luks1;
+    }
+    setLuksGenerationChoice( static_cast< LuksGenerationChoice >( c ) );
+}
+
+void
+Config::setLuksGenerationChoice( Config::LuksGenerationChoice c )
+{
+    if ( c != m_luksGenerationChoice )
+    {
+        m_luksGenerationChoice = c;
+        Q_EMIT luksGenerationChoiceChanged( c );
+        // TODO(Artur): This doesn't seem right but is also done at line 217
+        ::updateGlobalStorage( m_installChoice, c );
+    }
+}
+
+void
 Config::setEraseFsTypeChoice( const QString& choice )
 {
     QString canonicalChoice = PartUtils::canonicalFilesystemName( choice, nullptr );
@@ -355,6 +387,12 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
         m_initialSwapChoice = pickOne( m_swapChoices );
     }
     setSwapChoice( m_initialSwapChoice );
+
+    m_initialLuksGenerationChoice = luksGenerationChoiceNames().find(
+        CalamaresUtils::getString( configurationMap, "luksGeneration" ), nameFound );
+    setLuksGenerationChoice( m_initialLuksGenerationChoice );
+
+    setInstallChoice( m_initialInstallChoice );
 
     m_allowManualPartitioning = CalamaresUtils::getBool( configurationMap, "allowManualPartitioning", true );
     m_requiredPartitionTableType = CalamaresUtils::getStringList( configurationMap, "requiredPartitionTableType" );
--- a/src/modules/partition/Config.cpp
+++ b/src/modules/partition/Config.cpp
@@ -163,7 +163,7 @@ getSwapChoices( const QVariantMap& configurationMap )
 }
 
 void
-updateGlobalStorage( Config::InstallChoice installChoice, Config::SwapChoice swapChoice )
+updateGlobalStorage( Config::InstallChoice installChoice, Config::SwapChoice swapChoice, Config::LuksGenerationChoice luksGenerationChoice )
 {
     auto* gs = Calamares::JobQueue::instance() ? Calamares::JobQueue::instance()->globalStorage() : nullptr;
     if ( gs )
@@ -238,7 +238,7 @@ Config::setLuksGenerationChoice( Config::LuksGenerationChoice c )
         m_luksGenerationChoice = c;
         Q_EMIT luksGenerationChoiceChanged( c );
         // TODO(Artur): This doesn't seem right but is also done at line 217
-        ::updateGlobalStorage( m_installChoice, c );
+        ::updateGlobalStorage( m_installChoice, m_swapChoice, c );
     }
 }
 
--- a/src/modules/partition/Config.h
+++ b/src/modules/partition/Config.h
@@ -63,6 +63,15 @@ public:
 
     using EraseFsTypesSet = QStringList;
 
+    /** @brief Choice of luks disk encryption generation */
+
+    enum class LuksGenerationChoice {
+        Luks1,
+        Luks2
+    }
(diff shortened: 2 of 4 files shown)


src/modules/partition/Config.cpp
```diff
@@ -46,6 +46,15 @@ Config::swapChoiceNames()
     return names;
 }
 
+const NamedEnumTable< Config::LuksGenerationChoice >&
+Config::luksGenerationChoiceNames()
+{
+    static const NamedEnumTable< LuksGenerationChoice > names { { QStringLiteral( "luks1" ), LuksGenerationChoice::Luks1 },
```
adriaandegroot, line 52: clang-format has trouble making this kind of table "easy to read"; most of the other named enum tables have something like `// clang-format OFF` around them and some manual layout. Remember, too, that this table is a cheap way to introduce aliases if you feel that more names are needed (e.g. you could use `luks` as an alias of whatever you feel is appropriate)

src/modules/partition/Config.cpp
```diff
@@ -154,7 +163,7 @@ getSwapChoices( const QVariantMap& configurationMap )
 }
 
 void
-updateGlobalStorage( Config::InstallChoice installChoice, Config::SwapChoice swapChoice )
+updateGlobalStorage( Config::InstallChoice installChoice, Config::SwapChoice swapChoice, Config::LuksGenerationChoice luksGenerationChoice )
```
adriaandegroot, line 166: Huh. This function should totally be `static` (not necessarily part of your PR though). It's local to this implementation.

adriaandegroot, line 166: Is which-generation-of-LUKS something that is needed by other modules in Calamares (e.g. *bootloader*)? Then it makes sense to add it to Global Storage. If there isn't a consumer, then just leave out. Within the *partition* module pieces should be sharing information through the `Config` object (granted, not all of them do and it's a bit of a historical mess).

src/modules/partition/Config.cpp
```diff
@@ -206,7 +215,30 @@ Config::setSwapChoice( Config::SwapChoice c )
     {
         m_swapChoice = c;
         Q_EMIT swapChoiceChanged( c );
-        ::updateGlobalStorage( m_installChoice, c );
+        ::updateGlobalStorage( m_installChoice, c, m_luksGenerationChoice );
+    }
+}
+
+void
+Config::setLuksGenerationChoice( int c )
```
adriaandegroot, line 223: This -- as a method of the `Config` object -- is really only needed if you envision the setting **changing** at runtime. If it's just read-from-config-file, there's no special reason to make it a method of its own. Changing at runtime only makes sense if you expect the end-user to be able to select it somehow. Given the amount of machinery needed to support arbitrary LUKS choice (e.g. which packages to install, bootloader config, etc) it might not be worth it.

adriaandegroot, line 223: Why `int` rather than `LuksGenerationChoice` ? (The answer might be: "because of signal / slots, this needs to be a type that moc understands and I don't want to bother with all the Qt metatype stuff") Using a typed argument makes it easier on the compiler to check the values that are passed in.

adriaandegroot: .. i only see now that this is a wrapper to pass a value on to a more strongly-typed method, below.

You could still use a more strongly-typed approach here, and turn this into a `switch` and have the compiler help you check that all the cases are covered. Now, if you naively introduce a `Luks3`, you need to find this `if` through manual testing.

src/modules/partition/Config.cpp
```diff
@@ -206,7 +215,30 @@ Config::setSwapChoice( Config::SwapChoice c )
...
+    setLuksGenerationChoice( static_cast< LuksGenerationChoice >( c ) );
+}
+
+void
+Config::setLuksGenerationChoice( Config::LuksGenerationChoice c )
+{
+    if ( c != m_luksGenerationChoice )
+    {
+        m_luksGenerationChoice = c;
+        Q_EMIT luksGenerationChoiceChanged( c );
```
adriaandegroot, line 239: Again, this is only needed if you expect some (UI?) code to be hooked up that needs to know the LUKS generation **and** you expect it to become chageable at runtime.

src/modules/partition/Config.cpp
```diff
@@ -206,7 +215,30 @@ Config::setSwapChoice( Config::SwapChoice c )
...
+        // TODO(Artur): This doesn't seem right but is also done at line 217
+        ::updateGlobalStorage( m_installChoice, m_swapChoice, c );
```
adriaandegroot, line 241: If it feels weird, you might add additional overloads of `::updateglobalStorage()` that take just the new generation as argument. A different approach entirely would pass the `Config` object to `::updateGlobalStorage(const Config& c)` and that update function would read relevant values out of `Config` through the accessors of the class.

.. but only if the LUKS generation needs to be in GS in the first place.

src/modules/partition/Config.cpp
```diff
@@ -356,6 +388,12 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
     }
     setSwapChoice( m_initialSwapChoice );
 
+    m_initialLuksGenerationChoice = luksGenerationChoiceNames().find(
+        CalamaresUtils::getString( configurationMap, "luksGeneration" ), nameFound );
+    setLuksGenerationChoice( m_initialLuksGenerationChoice );
```
adriaandegroot, line 393: .. but you've just set it! In particular, the `if()` in that method is going to just skip everything because the value has not changed. This **might** mean that you should set `m_initialLuksChoice` to `None` (which you would need to add as an enum value, without giving it a name), and then call the setter with the new value. That's if you need the setter to do the work that the setter should (whatever it is).

Anyway, this bit looks a bit weird.

src/modules/partition/Config.cpp
adriaandegroot: .. and looking at the rest of the code in `Config`, I see there are more methods that offer an int and an enum overload. So you're copying existing code, although at this point I no longer remember why there are those overloads anyway. Keep it -- also because there's the *partitionq* branch which is using much more of the Config object.`

src/modules/partition/Config.h
```diff
@@ -63,6 +63,15 @@ class Config : public QObject
 
     using EraseFsTypesSet = QStringList;
 
+    /** @brief Choice of luks disk encryption generation */
+
+    enum LuksGenerationChoice {
```
adriaandegroot, line 68: Might be nice to have `None` or `Invalid` or `Unsupported` in there, just to indicate an uninitialized, invalid, or otherwise bogus value. You don't have to give it a name in the table.

src/modules/partition/Config.h
```diff
@@ -63,6 +63,15 @@ class Config : public QObject
...
+        Luks1,
+        Luks2
```
adriaandegroot, line 70: In C++17 you can leave a trailing `,` here. I think that's something cribbed from Python, where it's a natural thing to do at the end of a list. The effect is visible **later**, in diffs, because if you add another value to the list you don't have to remember to add the trailing `,` on the previous line. I've slowly grown to like it (but this is your code, do your thing; I'm pointing out that C++ syntax is less picky here than it used to be).

src/modules/partition/Config.h
```diff
@@ -63,6 +63,15 @@ class Config : public QObject
...
+    };
+    Q_ENUM( LuksGenerationChoice )
```
adriaandegroot, line 72: Making this a `Q_ENUM` probably makes the `int` overload of `setLuksGeneration` superfluous, though?

src/modules/partition/Config.h
```diff
@@ -112,6 +121,21 @@ class Config : public QObject
      */
     SwapChoice swapChoice() const { return m_swapChoice; }
 
+
+    /** @brief What kind of swap selection is requested **initially**?
```
adriaandegroot, line 125: Please don't copy-paste the API documentation from an unrelated method. Wrong documentation is worse than no documentation. Stylistically, not much of the API documentation uses `@return`, although it might be better if it did. If you fix this bit of docs to reflect what the function actually does, a later PR might go through the API docs of the Config object and polish them up, giving them a more consistent form.

src/modules/partition/Config.h
```diff
@@ -145,11 +169,14 @@ public Q_SLOTS:
     void setInstallChoice( InstallChoice );
     void setSwapChoice( int );  ///< Translates a button ID or so to SwapChoice
     void setSwapChoice( SwapChoice );
+    void setLuksGenerationChoice( int );  ///< Translates a button ID or so to SwapChoice
```
adriaandegroot, line 172: It does not. And as an overload it might not be needed at all; now that I have reached this part of the code, the only reason to have the `int` overload of `setSwapChoice()` is because that is hooked up to the UI, where a combo-box is used which signals changes in selection through an `int`. So this method is only needed if having a UI-visible choice to use LUKS1 or LUKS2 (or others) is a sensible proposition. Even then, there are other ways to do this that don't require an int-taking-slot.

src/modules/partition/partition.conf
```diff
@@ -64,6 +64,16 @@ userSwapChoices:
...
 
+# This setting specifies the luks generation (i.e luks1, luks2) used internally by cryptsetup
+# when creating an encrypted partition
+#
+# This option is set to luks1 by default, as grub doesn't support luks2 + argon2id currently.
+# On the other hand grub does support luks2 with pbkdf2 and could therefore be also set to luks2
+#
+# Choices: luks1, luks2
+#
+luksGeneration: luks1
```
adriaandegroot, line 75: Well said!

src/modules/partition/partition.schema.yaml
```diff
@@ -6,27 +6,33 @@ $id: https://calamares.io/schemas/partition
 additionalProperties: false
 type: object
 properties:
```
adriaandegroot, line 8: Are you using some automated reformatting tool here? Most (all?) the YAML files in the Calamares repo use 4-spaces indents, so the indentation changes come off as a bit gratuitous. **But** if you have a nice formatting tool that follows some kind of JSON-schema standards, I'd be happy to integrate it into `ci/calamaresstyle` and call it a coding standard -- automated formatting is a good thing.

src/modules/partition/partition.schema.yaml
```diff
@@ -6,27 +6,33 @@ $id: https://calamares.io/schemas/partition
...
 
-    userSwapChoices: { type: array, items: { type: string, enum: [ none, reuse, small, suspend, file ] } }
-    # ensureSuspendToDisk: { type: boolean, default: true }  # Legacy
-    # neverCreateSwap: { type: boolean, default: false }  # Legacy
+  userSwapChoices:
+    {
+      type: array,
+      items: { type: string, enum: [none, reuse, small, suspend, file] },
+    }
+  luksGeneration: { type: string, enum: [luks1, luks2] }
```
adriaandegroot, line 18: .. for the purpose of this PR, I'd prefer it if you reverted the indentation changes and just added this one line; then we can PR the indentation changes and tooling separately, if desired.

adriaandegroot (commented):
There's a handful of comments and considerations here. Overall I think this is well-done -- adding the bits for the `Config` object at least. Code-wise this could be landed as-is, but I'd appreciate it if you look at the docs comment in particular, and the design questions.

S0PEX:
Hey, thanks for your feedback. I will try to make the required changes this week.
The yaml formatting got messed up due to CLion code formatting.

AKmatiAK:
Hi, I can donate you S0PEX to finish it because I need this, also I encourage others to donate too plz.

S0PEX:
Hey, sorry for the delay.
I am currently working on my thesis and sadly have no time to fix this.
There is no need to donate it's just stuck due to lack of time.
I will try to work on it after the submission, most likely in October.

jtheoof:
Hello @S0PEX @adriaandegroot I tried to have a crack at this in #2047 based on the good start provided by @S0PEX here.

PR #1968, opened 2022-05-30
[libcalamaresui] sync naming change

demmm:
Branding.h moved the style/color naming, Branding.cpp needs to be in sync
add the changes to the default branding.desc & calamares-sidebar.qml

### DIFF
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -183,7 +183,7 @@ style:
    sidebarBackground:    "#292F34"
    sidebarText:          "#FFFFFF"
    sidebarTextCurrent:    "#292F34"
-   SidebarBackgroundCurrent: "#D35400"
+   sidebarBackgroundCurrent: "#D35400"
 
 ### SLIDESHOW
 #
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -182,8 +182,8 @@ images:
 style:
    sidebarBackground:    "#292F34"
    sidebarText:          "#FFFFFF"
-   sidebarTextSelect:    "#292F34"
-   sidebarTextHighlight: "#D35400"
+   sidebarTextCurrent:    "#292F34"
+   SidebarBackgroundCurrent: "#D35400"
 
 ### SLIDESHOW
 #
--- a/src/calamares/calamares-sidebar.qml
+++ b/src/calamares/calamares-sidebar.qml
@@ -1,7 +1,7 @@
 /* Sample of QML progress tree.
 
    SPDX-FileCopyrightText: 2020 Adriaan de Groot <groot@kde.org>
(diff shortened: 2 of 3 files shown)


src/branding/default/branding.desc
```diff
@@ -182,8 +182,8 @@ images:
 style:
    sidebarBackground:    "#292F34"
    sidebarText:          "#FFFFFF"
-   sidebarTextSelect:    "#292F34"
-   sidebarTextHighlight: "#D35400"
+   sidebarTextCurrent:    "#292F34"
+   SidebarBackgroundCurrent: "#D35400"
```
dalto8: The casing is inconsistent here.

demmm: This is to make the same as for Branding.h https://github.com/calamares/calamares/commit/a7d3630e42b7c001fc6a209a56e3baa7e25f575d#diff-ace2b49f072feb0bbaeea9f8e4197ffb50b3dbfe14a3327226bebef8e726ad43R79

demmm: Sorry, missed the mistake twice.....

src/calamares/calamares-sidebar.qml
```diff
@@ -51,7 +51,7 @@ Rectangle {
                 Text {
                     anchors.verticalCenter: parent.verticalCenter;
                     anchors.horizontalCenter: parent.horizontalCenter;
-                    color: Branding.styleString( index == ViewManager.currentStepIndex ? Branding.SidebarTextCurrent : Branding.SidebarText );
+                    color: Branding.styleString( index == ViewManager.currentStepIndex ? Branding.SidebarTextSelect : Branding.SidebarText );
```
adriaandegroot, line 54: This looks backwards to me. You just introduced the **new** name SidebarTextCurrent.

adriaandegroot:
What is this fixing, exactly? There are multiple names in play here -- they might not be consistent, which is certainly **confusing**, but the change seems to confuse them "the other way" from what they were.

- There's the enum **identifiers**, visible in C++; that's `enum StyleEntry` in `Branding.h`
- There are the enum names visible to Qt, generated through `Q_ENUM`; these names are the same as the identifiers
- There are the keys used in the configuration files

Because of the (kinda derpy) way that the Branding object loads the strings from the *style* key in the config file, the keys need to be the same as the identifiers.

(It might have been part of the plan to break all the `branding.desc` files for 3.3, introducing the new and somewhat-more-consistent names. That would be something that can be extended a bit, checking that the style names that are **used** in the `branding.desc` file are actually valid)

demmm:
With the change in Branding.h, I could not get calamares-seidebar.qml to show correct colors with what is currently set for colors in branding.desc.  So I tried to make Branding.cpp & Branding.h use the same terms (colors now show correctly locally).

But I probably misunderstand what the todo removal/commit in Branding.h accomplished, so I will keep the changes locally only.

PR #1973, opened 2022-06-03
[calamares] Improve "about" and "debug" buttons

adriaandegroot:
This does two things:
- restores the text on the "special" buttons for About and Debug
- renames the object for the About button to "aboutButton" instead of "aboutDialogButton" so the name is consistent with the other button there.

This branch could also be merged back / cherry-picked to 3.2-stable.


adriaandegroot:
The CI failure is not related to the code, but to the CI scripts.

demmm:
Builds, runs, text added for debug & about button:
Screenshot_20220607_144932
Tested on main (3.3) only, but imo ready to merge.

adriaandegroot:
Also cherry-picked to 3.2.x

Issue #1967, opened 2022-05-30
How can we change the look and feel of the new about and debug buttons? (calamares 3.2.59)


killajoe:
#What is the issue and needs to get sorted in the testing?

Before calamares it was this how the debug button looks like on the sidebar, about was on the main window centered..
![before](https://user-images.githubusercontent.com/16797647/170970707-568565c7-3bcf-4364-a080-b458b0f193be.png)
Now both buttons are on the sidebar/menu as icons only:
![cal59](https://user-images.githubusercontent.com/16797647/170970813-5d374e25-2a6e-420e-a13e-485f06b6cf61.png)
The icons are in a dark color badly visible on p.e. darker backgrounds:
![on-grey](https://user-images.githubusercontent.com/16797647/170970898-a3d76180-f9b0-4363-a8d5-a4f72975962d.png)

On EndeavourOS we do used to set the colors for the debug button inside `.../calamares/branding/endeavouros/stylesheet.qss`

https://github.com/endeavouros-team/EndeavourOS-calamares/blob/912214630981ddf76df0b0eb583c4a73dc5d68b5/calamares/branding/endeavouros/stylesheet.qss#L61

This namings does not seem to work anymore and have no effect to the colors e.t.c for the icons/buttons

# To collect:
1. is everyone having the same ? or is this something caused by special setups in use
a. EndeavourOS uses stylesheet.qss and starts calamare with `-style kvantum` flag (not using kvantum doe snot seem to change anything related)
2. Is stylesheet.qss still a valid way to set colors e.t.c for this? 
a. do we simple need new naming? 
b. where are the icons taken from? where is this handled in the code?
3. How we can change the look and feel to fit our needs? 
a. is it possible to use a text instead of the only icon L&F? 

This is indeed not really a test but a call out to collect needed info to setup the buttons to fir the need of usage.

adriaandegroot:
Somewhat-related: branch *fix-color-names* (that will mess up branding even more as the key-names in the branding file change to match the code identifiers) I don't think this will apply to 3.2.59, because breaking everyone's branding file is not a 3.2-stable kind of thing.

I suppose these two are "special buttons" and could get "special colors" to apply to them. (i.e. a key in *style* in `branding.desc`) 

[...]

PR #1976, opened 2022-06-07
[libcalamaresui] Fix color names

adriaandegroot:
This is related to #1968 , but fixes the underlying problem of mismatched names in a different way. It also adds a validation step that checks that the `branding.desc` doesn't name styles (settings for a color) that are not named in code as well.

### DIFF
--- a/src/branding/default/branding.desc
+++ b/src/branding/default/branding.desc
@@ -170,20 +170,20 @@ images:
 
 # Colors for text and background components.
 #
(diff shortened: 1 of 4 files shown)


adriaandegroot:
FIXES #1967 

Well, it makes the names consistent with what @killajoe expected, and makes the example QML sidebar work again.

demmm:
All sidebar colors set as should again, tested for both the widget & QML sidebar, ready to merge.

PR #2007, opened 2022-07-02
Use `${}` everywhere instead of `@@`

adriaandegroot:
The substitution syntax in Calamares is inconsistent, created by some arbitrary coding in many places. Use `${}` (e.g. DictionaryExpander) syntax everywhere, although there's an issue with Python not supporting that directly (so hack it into the bootloader module instead).

Two preliminary commits make the DictionaryExpander nicer to use.

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -393,28 +393,28 @@ class phraseEfi(object):
 
 def get_efi_suffix_generator(name):
     """
-    Handle EFI bootloader Ids with @@<something>@@ for suffix-processing.
+    Handle EFI bootloader Ids with ${<something>} for suffix-processing.
     """
-    if "@@" not in name:
-        raise ValueError("Misplaced call to get_efi_suffix_generator, no @@")
-    parts = name.split("@@")
-    if len(parts) != 3:
-        raise ValueError("EFI Id {!r} is malformed".format(name))
-    if parts[2]:
-        # Supposed to be empty because the string ends with "@@"
-        raise ValueError("EFI Id {!r} is malformed".format(name))
-    if parts[1] not in ("SERIAL", "RANDOM", "PHRASE"):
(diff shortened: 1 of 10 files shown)


src/modules/bootloader/main.py
```diff
@@ -393,39 +393,39 @@ def next(self):
...
-        generator = phraseEfi(parts[0])
+    if generator_name == "SERIAL":
+        generator = serialEfi(prefix)
+    elif generator_name == "RANDOM":
+        generator = randomEfi(prefix)
+    elif generator_name == "PHRASE":
+        generator = phraseEfi(prefix)
     if generator is None:
-        raise ValueError("EFI suffix {!r} is unsupported".format(parts[1]))
+        raise ValueError("EFI suffix {!r} is unsupported".format(generator_name))
```
dalto8, line 417: This part seems strange.  Couldn't this be a single if statement instead of 3?

```
    generator = None
    if generator_name == "SERIAL":
        generator = serialEfi(prefix)
    elif generator_name == "RANDOM":
        generator = randomEfi(prefix)
    elif generator_name == "PHRASE":
        generator = phraseEfi(prefix)
    else:
        raise ValueError("EFI suffix {!r} is unsupported".format(generator_name))
```

dalto8 (approved):

src/modules/bootloader/main.py
kkofler: If the methods can return `None`, then it makes sense to check for `None` always, even after setting `generator`.

dalto8: Hmm...in this case that would mean the object failed to instantiate?

If that is possible, would that be the right error message for that?

EDIT: I should add that it looks good overall.  This was just a comment.

adriaandegroot: The idea was, indeed, that a function like `randomEfi()` could return `None` if it didn't like the prefix after all.

PR #2011, opened 2022-07-03
Improve kpmcore use

adriaandegroot:
This should fix the nightly-debian problem:

- allow an older KPMcore
- do better when using the library, it needs more include-path wrangling
- mark a TODO where a common header assumes KPMcore is available

### DIFF
--- a/.github/workflows/nightly-debian.yml
+++ b/.github/workflows/nightly-debian.yml
@@ -9,9 +9,7 @@ env:
   BUILDDIR: /build
   SRCDIR: ${{ github.workspace }}
   CMAKE_ARGS: |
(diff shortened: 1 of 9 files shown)


abalfoort:
Should CMakeModules/KPMcoreHelper.cmake be fixed also?
It searches specifically for KPMcore 21.12.0 when building which I have to change to 20.12.0 if I want to build Calamares on Debian 11 (current stable).

adriaandegroot:
> Should CMakeModules/KPMcoreHelper.cmake be fixed also? It searches specifically for KPMcore 21.12.0

Um .. it already is, in this branch?

PR #2015, opened 2022-07-05
Add options in services-systemd module to enable systemd sockets and paths

pvshvp-oss:
Currently the `services-systemd` module does not allow enabling of sockets and paths because suffixes are pre-determined. This PR will not change the existing functionality of the module. It will only add support for enabling sockets and paths.

### DIFF
--- a/src/modules/services-systemd/main.py
+++ b/src/modules/services-systemd/main.py
@@ -61,6 +61,10 @@ def systemctl(targets, command, suffix):
 
                 if command == "enable" and suffix == ".service":
                     description = _("Cannot enable systemd service <code>{name!s}</code>.")
+                elif command == "enable" and suffix == ".socket":
+                    description = _("Cannot enable systemd socket <code>{name!s}</code>.")
+                elif command == "enable" and suffix == ".path":
+                    description = _("Cannot enable systemd path <code>{name!s}</code>.")
                 elif command == "enable" and suffix == ".target":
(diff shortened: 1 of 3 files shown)


dalto8:
Since this will probably end up in 3.3, it seems like a great time to make a backwards incompatible change.

I wonder if we should instead just remove the suffixes.  It would make both the conf file and the code simpler and we would no longer need to add each type individually.

pvshvp-oss:
> Since this will probably end up in 3.3, it seems like a great time to make a backwards incompatible change.
> 
> I wonder if we should instead just remove the suffixes. It would make both the conf file and the code simpler and we would no longer need to add each type individually.

I do not mind that. When is 3.3 scheduled to release?

dalto8:
> When is 3.3 scheduled to release?

I don't think there is a defined date but any distro who wanted this sooner than that could easily pull this module in.  Making that change should be totally isolated to this module.

kkofler:
I suppose we could just let the distros write, e.g., `cups.service` or `cups.socket` explicitly, at least when it is ambiguous. (Otherwise, systemd has actually become pretty good at guessing on its own.) So I agree the hardcoded sections in the `.conf` file are not all that helpful.

pvshvp-oss:
I am writing up an alternative PR for removing hardcoded suffixes and changing the schema and conf file to match. Will test it first before sending

pvshvp-oss:
This pull request will be supplanted by https://github.com/calamares/calamares/pull/2016

PR #2016, opened 2022-07-05
Remove suffix-based grouping in services-systemd

pvshvp-oss:
The explicit grouping of *systemd* units by **suffix** in the `services-systemd` module gave rise to the problems of 
- Proliferation of configuration options per suffix
- The need to manually add unit types (like systemd sockets and paths) per suffix

This pull request intends to eliminate *suffix-based* grouping altogether so that in the configuration file, users can just group units by `enable`, `disable`, or `mask`. `mask` is an existing example of such a suffix-less group and the rest of the actions (namely `enable`, and `disable`) follow the conventions used in the implementation of `mask`. The corresponding log messages, schema file, and configuration file (along with the documentation) are also modified.

In addition to the above main goal, there are other minor changes such as renaming the multiple names that units were referred by (some example terms used in main.py are `svcs`, and `targets` in the systemctl function parameter list), to a uniform term: *units*. This is distinct from the earlier *suffix* (also named `target`) that is already obsoleted. And the parameter list of the `systemctl` function is re-ordered to mirror command invocation in a shell: `systemctl <action> <unit>` -> `systemctl(command, units)` to prevent potential errors due to the unintuitive ordering

### DIFF
--- a/src/modules/services-systemd/main.py
+++ b/src/modules/services-systemd/main.py
@@ -7,6 +7,7 @@
 #   SPDX-FileCopyrightText: 2014 Teo Mrnjavac <teo@kde.org>
 #   SPDX-FileCopyrightText: 2017 Alf Gaida <agaida@siduction.org>
 #   SPDX-FileCopyrightText: 2018-2019 Adriaan de Groot <groot@kde.org>
+#   SPDX-FileCopyrightText: 2022 shivanandvp <shivanandvp@rebornos.org>
 #   SPDX-License-Identifier: GPL-3.0-or-later
 #
 #   Calamares is Free Software: see the License-Identifier above.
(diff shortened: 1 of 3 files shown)


src/modules/services-systemd/services-systemd.conf
```diff
@@ -3,20 +3,20 @@
...
-# There are several configuration keys for this module:
-# *services*, *targets*, *timers*, *disable*, *disable-targets* and *mask*.
+# There are three configuration keys for this module:
+# *enable*, *disable*, and *mask*.
 # The value of each key is a list of entries. Each entry has two keys:
-#   - *name* is the (string) name of the service or target that is being
-#     changed. Use quotes. Don't include unit suffix in the name. For
-#     example, it should be "NetworkManager", not "NetworkManager.service"
+#   - *name* is the (string) name of the unit that is being changed.
+#     Use quotes. Please include unit suffix in the name. For example,
```
dalto8, line 18: I don't think we need to require the unit suffix in the name.  Why not leave it up to the distro if they want to include the suffix or not?

src/modules/services-systemd/main.py
```diff
@@ -23,66 +24,59 @@
...
-            name = svc
+    for unit in units:
+        if isinstance(unit, str):
+            name = unit
             mandatory = False
         else:
-            name = svc["name"]
-            mandatory = svc.get("mandatory", False)
+            name = unit["name"]
+            mandatory = unit.get("mandatory", False)
```
dalto8, line 45: Is this still needed or is it legacy code?

dalto8 (commented):
I like this a lot better but couldn't we go one level more generic?  In other words, why limit the commands?

Think about this layout:
```
services:
  - name: "NetworkManager.service"
    command: "enable"
    mandatory: true
  - name: "cups.socket"
    command: "enable"
    mandatory: false
  - name: "pacman-init.service"
    command: "disable"
    mandatory: false
 - name: "bluetooth.service"
   command: "mask"
   mandatory: true
```

In this model, we could use other commands without modification.  For example, we have `mask` but not `unmask`.

It would have the added advantage of distros being able to specify their own order.

Just an idea.  Either way, doing it this way is a big improvement from the current model.

pvshvp-oss:
> I like this a lot better but couldn't we go one level more generic? In other words, why limit the commands?
> 
> Think about this layout:
> 
> ```
> services:
>   - name: "NetworkManager.service"
>     command: "enable"
>     mandatory: true
>   - name: "cups.socket"
>     command: "enable"
>     mandatory: false
>   - name: "pacman-init.service"
>     command: "disable"
>     mandatory: false
>  - name: "bluetooth.service"
>    command: "mask"
>    mandatory: true
> ```
> 
> In this model, we could use other commands without modification. For example, we have `mask` but not `unmask`.
> 
> It would have the added advantage of distros being able to specify their own order.
> 
> Just an idea. Either way, doing it this way is a big improvement from the current model.

Is the list order preserved in whichever library you use for parsing? 

I must say this is beginning to look like a `shellprocess`, with the module configured to read a list like this: 
```yaml
- "systemctl enable NetworkManager.service"
- "-systemctl enable cups.socket"
- "-systemctl mask vboxservice.service"
```

src/modules/services-systemd/services-systemd.conf
pvshvp-oss: > I don't think we need to require the unit suffix in the name. Why not leave it up to the distro if they want to include the suffix or not?

In that case, do we resolve `cups.service` from `cups.socket` if the configuration only specifies `cups`? Send it directly to `systemctl`?

dalto8: > How do we resolve `cups.service` from `cups.socket` if the configuration only specifies `cups`? Send it directly to `systemctl`?

Yes, exactly.  systemd already has a method for resolving those conflicts.  It is just a matter of changing the comments in the conf file.

kkofler:
> Is the list order preserved in whichever library you use for parsing?

Yes, YAML lists are ordered, i.e., the order in the YAML file is preserved, **if** the objects are in the same list. (Of course not if one is under `enable:` and one under `disable:`, which is the point of that "It would have the added advantage of distros being able to specify their own order." comment.)

src/modules/services-systemd/main.py
pvshvp-oss: It handles the two forms of input for specifying units: (1) Just a name, or (2) A name and whether it is mandatory or not. It looks like this code is still necessary

dalto8: It looks like it is checking if `unit` is a string.  If it is, it sets the name to it.

That doesn't look like a valid option at this point.  Am I missing something?

pvshvp-oss: It could either be a string (just the name) , or a map that contains a name and a flag. Reminding that this PR groups the units by action, and the `systemctl` function in `main.py` is still called with a list of unit information, just like how it was earlier. This information either contains just the name (example `NetworkManager.service`) or a map of name and a boolean flag
```
name: "NetworkManager.service"
mandatory: true
```
This behaviour has not changed

pvshvp-oss:
Let me try putting together another PR since I have to rewrite basically everything

PR #2017, opened 2022-07-06
Ungroup systemd units in services-systemd module

pvshvp-oss:
*Based on suggestions by @dalto8 in https://github.com/calamares/calamares/pull/2016*

The explicit grouping of *systemd* units by **suffix** in the `services-systemd` module gave rise to the problems of 
- Proliferation of configuration options per suffix
- The need to manually add unit types (like systemd sockets and paths) per suffix

This pull request intends to eliminate grouping altogether so that in the configuration file, users can just list out units and actions in the order they want handled. An example for a configuration this pull request is meant to handle: 
```yaml
units:
  - name: "NetworkManager"
    action: "enable" 
    mandatory: true

  - name: "cups.socket"
    action: "disable" 
    mandatory: false

  - name: "graphical"
    action: "enable" 

  - name: "pacman-init"
    action: "mask"

  - gdm
  
  - vboxservice.service
  ```
The above example tells the module to perform the below steps in order: 
- *enable* the `NetworkManager` unit (systemd automatically figures out the unit to be *NetworkManager.service*) and fail the installation if the operation fails.
- *disable* the `cups` *socket* unit and proceed with a warning message if the operation fails.
- *enable* the `graphical` unit (systemd automatically figures out the unit to be *graphical.target*) and proceed with a warning message if the operation fails.
- *mask* the `pacman-init` unit (systemd automatically figures out the unit to be *pacman-init.service*) and proceed with a warning message if the operation fails.
- *enable* the `gdm` unit (systemd automatically figures out the unit to be *gdm.service*) and proceed with a warning message if the operation fails.
- *enable* the `vboxservice` service unit and proceed with a warning message if the operation fails.

### DIFF
--- a/src/modules/services-systemd/main.py
+++ b/src/modules/services-systemd/main.py
@@ -7,10 +7,10 @@
 #   SPDX-FileCopyrightText: 2014 Teo Mrnjavac <teo@kde.org>
 #   SPDX-FileCopyrightText: 2017 Alf Gaida <agaida@siduction.org>
 #   SPDX-FileCopyrightText: 2018-2019 Adriaan de Groot <groot@kde.org>
+#   SPDX-FileCopyrightText: 2022 shivanandvp <shivanandvp@rebornos.org>
 #   SPDX-License-Identifier: GPL-3.0-or-later
 #
 #   Calamares is Free Software: see the License-Identifier above.
-#
 
 import libcalamares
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/modules/services-systemd/main.py [hunk 1]
```diff
@@ -23,96 +23,61 @@
...
+
+    for unit in units:
+        if isinstance(unit, str):
+            name = unit
+            action = "enable"
             mandatory = False
         else:
-            name = svc["name"]
-            mandatory = svc.get("mandatory", False)
+            name = unit["name"]
```
dalto8, line 48: What happens here if `unit["name"]` is not populated.  It should probably write an error to the log and move onto the next service.

src/modules/services-systemd/services-systemd.schema.yaml
```diff
@@ -4,22 +4,18 @@
 $schema: https://json-schema.org/schema#
 $id: https://calamares.io/schemas/services-systemd
 definitions:
-    service:
-        $id: 'definitions/service'
+    unit:
+        $id: 'definitions/unit'
         type: object
-        description: a name and a flag for services, targets, and others
+        description: a map containing a unit name, an action, and whether it is mandatory
```
dalto8, line 10: A minor point but I think in yaml it is called a "mapping", not a "map"

dalto8 (commented):
> ```yaml
>   - gdm
>   
>   - vboxservice.service
> ```

I asked about this in the other PR as well but since we are changing the format of the file anyway, why would we want to continue to allow this syntax?  Even more so in the revised conf layout.

It seems like having multiple ways to specify the same thing would only make sense if we were trying to maintain backwards compatibility which we have already broken.

src/modules/services-systemd/services-systemd.schema.yaml
pvshvp-oss: **Implemented as suggested**

src/modules/services-systemd/main.py
pvshvp-oss: **Implemented as suggested**. Although in that case (when the 'name' is missing), it violates the schema format which clearly specifies that the 'name' is required.

pvshvp-oss:
> > ```yaml
> >   - gdm
> >   
> >   - vboxservice.service
> > ```
> 
> I asked about this in the other PR as well but since we are changing the format of the file anyway, why would we want to continue to allow this syntax? Even more so in the revised conf layout.
> 
> It seems like having multiple ways to specify the same thing would only make sense if we were trying to maintain backwards compatibility which we have already broken.

This example is just a short one. A large enough `netinstall` module will usually have dozens of services that will potentially have to be enabled when the services exist. This module (as suggested by the wording in the prior documentation) is mainly used by distributions to enable services *en masse* in a non mandatory manner, by providing a list of services. This PR preserves that functionality for systemd units in addition to meeting all the other suggestions given. Removing this behavior in my opinion would be regressive and unnecessarily verbose for the most common use case for this module.

dalto8 (approved):

src/modules/services-systemd/main.py
```diff
@@ -23,96 +23,64 @@
...
-                    description = _("Cannot mask systemd unit <code>{name!s}</code>.")
-                else:
-                    description = _("Unknown systemd commands <code>{command!s}</code> and <code>{suffix!s}</code> for unit {name!s}.")
-
-                return (title,
-                        description.format(name=name, command=command, suffix=suffix) + " " + diagnostic
-                        )
+                title = _("Cannot modify unit")
+                diagnostic = _("<code>systemctl {_action!s}</code> call in chroot returned error code {_exit_code!s}.").format(_action=action, _exit_code=exit_code)
+                description = _("Cannot {_action!s} systemd unit <code>{_name!s}</code>.").format(_action=action, _name=name)
```
adriaandegroot, line 66: I have some doubts that this will work well in translation, though. Translators will need to know that `{_action!s}` is not-to-be-translated, which is plausible. Also `enable`, `disable`, ... will end up as (latin1) text in the string, which might not make sense at all. In English it stands in for a verb, so it remains grammatical, but any other language is going to be quite pecuiliar.

adriaandegroot:
Merged, with reservations about the translatability of the error messages. (FWIW, since Calamares can use Python 3.6, moving to f-strings is a possibility for substitutions in general with Python modules)

PR #2020, opened 2022-07-17
[partition] Fix encryption being shown when not available

dalto8:
Fixes a bug where the encryption checkbox is shown when manual partitioning is selected.


demmm:
Actually, fixes a bug where the encryption widget was always shown, regardless of manual option, no option selected yet, and so on.

demmm (approved):

adriaandegroot:
CI failure is just CI being derpy / GH actions being impenetrable opaque crap. The change itself looks good; it might be worth introducing a `canEnableEncryption()` method or some such, since the logic is a bit convoluted **and** you might want to re-use that logic elsewhere. **Might** be.

PR #2032, opened 2022-07-29
add HDPI support, fix #1945

### LINKED ISSUE #1945: Calamares needs support for high DPI
On high resolution displays, the Calamares window starts really small, making it difficult to use.
Myself and other users have managed to reproduce this, see the following bug report for an example:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=992162

I don't have a clear solution for this, and not sure what's possible in Qt, perhaps Calamares can adjust the size of UI elements when a certain resolution is detected, or just double its pixels.

codesardine:
@demmm I have tested this on my end and it fixes HDPI issues.

### DIFF
--- a/src/calamares/main.cpp
+++ b/src/calamares/main.cpp
@@ -107,6 +107,7 @@ handle_args( CalamaresApplication& a )
 int
 main( int argc, char* argv[] )
 {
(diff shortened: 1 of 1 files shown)


demmm:
This fixes https://github.com/calamares/calamares/issues/1945

PR #2040, opened 2022-08-10
Add detection of default lightdm-greeter

paolodongilli:
Instead of choosing the first (which order?) .desktop file found in directory /usr/share/xgreeters, I propose to choose the **default** lightdm greeter on the system (which could be `lightdm-gtk-greeter`, `slick-greeter`, etc.) as configured by `update-alternatives`.

### DIFF
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -640,34 +640,37 @@ class DMlightdm(DisplayManager):
 
         # configure lightdm-greeter
         greeter_path = os.path.join(
-            self.root_mount_point, "usr/share/xgreeters"
+            self.root_mount_point, "usr/share/xgreeters/lightdm-greeter.desktop"
+            # lightdm-greeter.desktop is typically a symlink managed
+            # by update-alternatives pointing to
+            # /etc/alternatives/lightdm-greeter which is also a
+            # symlink to a real .desktop file back in
+            # /usr/share/xgreeters/
             )
 
         if (os.path.exists(greeter_path)):
-            # configure first found lightdm-greeter
-            for entry in os.listdir(greeter_path):
-                if entry.endswith('.desktop'):
-                    greeter = entry.split('.')[0]
-                    libcalamares.utils.debug(
-                        "found greeter {!s}".format(greeter)
-                        )
-                    os.system(
-                        "sed -i -e \"s/^.*greeter-session=.*"
-                        "/greeter-session={!s}/\" {!s}".format(
-                            greeter,
-                            lightdm_conf_path
-                            )
-                        )
-                    libcalamares.utils.debug(
-                        "{!s} configured as greeter.".format(greeter)
-                        )
-                    break
-            else:
-                return (
-                    _("Cannot configure LightDM"),
-                    _("No LightDM greeter installed.")
+            # find the default lightdm-greeter
+            entry = os.path.basename(os.path.realpath(greeter_path))
+            if entry.endswith('.desktop'):
+                greeter = entry.split('.')[0]
+                libcalamares.utils.debug(
+                    "found greeter {!s}".format(greeter)
+                )
+                os.system(
+                    "sed -i -e \"s/^.*greeter-session=.*"
+                    "/greeter-session={!s}/\" {!s}".format(
+                        greeter,
+                        lightdm_conf_path
                     )
-
+                )
+                libcalamares.utils.debug(
+                    "{!s} configured as greeter.".format(greeter)
+                )
+        else:
+            return (
+                _("Cannot configure LightDM"),
+                _("No LightDM greeter installed.")
+            )
 
 class DMslim(DisplayManager):
     name = "slim"


dalto8:
`/usr/share/xgreeters/lightdm-greeter.desktop` and `update-alternatives` are not available on all distros.  I don't think Calamares should depend on them.

paolodongilli:
@dalto8 I see your point but the function should take into consideration the case where there are more than one lightdm-greeter in the ISOs and that only of them is the default. Choosing one greeter shouldn't be just picking the first entry found by `os.listdir(greeter_path)`. It seems too simplistic to me. Surely working, no doubt, unless none is found as the function itself already foresees:
```python
else:
                return (
                    _("Cannot configure LightDM"),
                    _("No LightDM greeter installed.")
                    )
```

dalto8:
Maybe can we step back and identify what is the problem we are actually trying to solve here?

Are you creating/maintaining a distro and this is causing an issue?  If so, what is the issue?

paolodongilli:
Sure, thanks. 
We are working on a Debian 11 "bullseye" derivative called FUSS (https://fuss.bz.it). In the ISOs (all based on xfce + lightdm) we foresee 2 possible lightdm-greeters:
- the default `lightdm-gtk-greeter`
- `slick-greeter` (which is our default)

We noticed that on some machines the function `os.listdir(greeter_path)` returned as first file `slick-greeter.desktop`, on others `lightdm-gtk-greeter`. The same Python Documentation states that `os.listdir` returns _ the list in arbitrary order_.
This is why we ask for a deterministic way of choosing one of the different lightdm-greeters installed.

This is the content of the `/usr/share/xgreeters` in debian "bullseye" when the two greeters are installed:
```bash
root@machine-1:~# ls -l /usr/share/xgreeters/
totale 8
lrwxrwxrwx 1 root root  33  9 lug 14.09 lightdm-greeter.desktop -> /etc/alternatives/lightdm-greeter
-rw-r--r-- 1 root root 187  4 nov  2020 lightdm-gtk-greeter.desktop
-rw-r--r-- 1 root root 131 27 dic  2020 slick-greeter.desktop
```
This is what os.listdir returns:
```
root@machine-1:/usr/share/xgreeters# python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.listdir('/usr/share/xgreeters')
['lightdm-gtk-greeter.desktop', 'slick-greeter.desktop', 'lightdm-greeter.desktop']
```

On another machine:

same files

```bash
root@machine-2:~# ls -l /usr/share/xgreeters/
totale 8
lrwxrwxrwx 1 root root  33  9 lug 14.09 lightdm-greeter.desktop -> /etc/alternatives/lightdm-greeter
-rw-r--r-- 1 root root 187  4 nov  2020 lightdm-gtk-greeter.desktop
-rw-r--r-- 1 root root 131 27 dic  2020 slick-greeter.desktop
```
but different order in Python with `os.listdir`:
```
root@machine-2:~# python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.listdir('/usr/share/xgreeters')
['slick-greeter.desktop', 'lightdm-gtk-greeter.desktop', 'lightdm-greeter.desktop']
```

demmm:
It seems any hardcoding of files or paths will be problematic.  This should be adjusted in the bootloader.conf, make the default .desktop adjustable from there.

dalto8:
Hmm....one option would be to add a default greeter in displaymanager.conf.  Since that won't work for all distros, if it is empty or not found, fall back to picking the first one from the list.  I wonder if we should also sort the list so the behaviour is deterministic.

adriaandegroot:
Let's be clear: this is, **right now**, special handling for LightDM, but I can imagine other DMs having multiple possible greeters, and those greeters being managed by the system (package manager, alternatives, whatever). So let's solve this in a somewhat-generic way for LightDM now, keeping in mind that it might be applicable elsewhere.

@paolodongilli thank you for bringing this up. I don't think many OS installs will put multiple greeters (they'll install just whatever default they want, so then this code isn't a problem). But let's fix that.

So here's what we know:
- `/usr/share/xgreeters` is a directory,
- in this directory there are one (zero?) or more `.desktop` files,
- one of these might be the preferred one. The preferred one is called `lightdm-greeter.desktop` (which could be a symlink or whatever).

If we look at the existing code mostly intact, we can see:
- if the directory exists,
- we loop over the files in it,
- and the first `.desktop` file we find is picked (its name is assigned to variable *greeter*),
- the *greeter* is substituted into the LightDM configuration file,
- the loop breaks.

That is some terrible software design right there, with a loop doing two things: searching and then applying the result of the search. First, we can rename *greeter_path*, since it's not a path: it's a directory. Then let's refactor:

```
    greeter = self.find_preferred_greeter(greeter_directory)
    if greeter is not None:
        # Do the sed thing
    else:
        # Return an error
```

The loop is gone! It is hidden inside the `find_preferred_greeter()` function. That function does one thing: searches for a suitable file. We have a better design, and a clear point where we could put customizations. If we implement this function just for LightDM, then it might be designed as follows:

- list the files in the directory
- if one of them is named like the preferred greeter (e.g. if `lightdm-greeter.desktop` is in the list), return the basename of that,
- otherwise sort the list, return the first `.desktop` file in there,
- otherwise return `None`.

An alternative (ha!) would be to look if any of the `.desktop` files is a symlink, and is a symlink to the LightDM-greeter-alternative, and then return that as the preferred one; this is a bit more flexible (allows other names in the directory itself) but is also more Debian-specific because of that symlink-to-alternatives thing. I think just looking for the preferred name is the way to go. That can be documented in the config file as well as a note about LightDM.

adriaandegroot:
@paolodongilli do you foresee working on this, dealing with the design issues? There's a sketch in my previous comment about how it could be approached. If you don't , we should close this PR as "Debian-specific", file an issue, and approach it anew (in a new approach I would still argue that improving the architecture of the module is important)

paolodongilli:
> @paolodongilli do you foresee working on this, dealing with the design issues? There's a sketch in my previous comment about how it could be approached. If you don't , we should close this PR as "Debian-specific", file an issue, and approach it anew (in a new approach I would still argue that improving the architecture of the module is important)

Thank you for your feed-back @adriaandegroot. Yes, I'm going to apply your suggestions to the PR in the coming days.

adriaandegroot:
I'll pick this one up and land it after all.

dalto8:
This is superseded by #2101

PR #2041, opened 2022-08-16
Fix grammar

cake03:


demmm:
Placeholder text is generally viewed like a title/header (as long as it is not a sentence with explanations), so capitalization is the correct way to display in this context.

PR #2045, opened 2022-08-26
[localeq] move to using Drawer for fine tuning options

demmm:
i18n.qml no longer needed
add color setting options to localeq.qml
Offline.qml updated to be inline with keyboardq UI, set index number according to default
America/New York


demmm:
Obligatory image showing the change, here is the locale Drawer activated:

Screenshot_20220826_153821

PR #2046, opened 2022-09-02
[dracut] add option to make the kernel name configurable

demmm:


dalto8:
Should we add a yaml schema since we are adding a conf file?

As a side note, someday we should think about standardizing string quoting in Python.

src/modules/dracut/dracut.conf
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: no
+# SPDX-License-Identifier: CC0-1.0
+#
+# Run dracut(8) with an optional kernel name
+---
+# Dracut defaults to setting initramfs-<kernel-version>.img
+# If you want to specify another filename for the resulting image,
+# set a custom kernel name, including the path
+#
+# kernelName: /boot/initramfs-linux.img
```
adriaandegroot, line 10: This should probably be uncommented, so that there is a specific thing to refer to. And since there is a config file, it should get a schema file (as @dalto8 says). It's eight lines or so.

src/modules/dracut/main.py
```diff
@@ -33,7 +34,12 @@ def run_dracut():
 
     :return:
     """
-    return target_env_call(['dracut', '-f'])
+    kernelName = libcalamares.job.configuration['kernelName']
+
+    if kernelName != 0:
```
adriaandegroot: This is a weird one .. wouldn't it be empty or `None` if unset?

src/modules/dracut/main.py
```diff
@@ -33,7 +34,12 @@ def run_dracut():
...
+        return check_target_env_call(['dracut', '-f', '{}'.format(kernelName)])
```
adriaandegroot: .. but `kernelName` is a string already, isn't it? So it's not necessary to format to string (or you could use `str()` here).

src/modules/dracut/dracut.conf
demmm: Default for dracut is to use `initramfs-<kernel-version>.img`, distributions that have been using this module for the last 8 years should not have to touch a (new) .conf to keep the default behavior.

src/modules/dracut/main.py
demmm: I used it that way for consistency in the file (see line 59), but not clear to me what the standard way should be, from:
https://docs.python.org/3/library/string.html#format-string-syntax

>Changed in version 3.1: The positional argument specifiers can be omitted for [str.format()](https://docs.python.org/3/library/stdtypes.html#str.format), so '{} {}'.format(a, b) is equivalent to '{0} {1}'.format(a, b).

PR #2047, opened 2022-09-07
[partition] Add support for LUKS2

jtheoof:
This commit adds support for LUKS2 behind a new `partition.conf` key:
`luksGeneration`. It started from the good work from @S0PEX originating
in #1964

A bit of context, LUKS2 is the default encryption operating mode since
cryptsetup >= 2.1.0 (See [Arch
wiki](https://wiki.archlinux.org/title/dm-crypt/Device_encryption#Encryption_options_with_dm-crypt).
It is considered more secured and allows additional extensions. It also
comes with Argon2id as the default Password Based Key Derivation
Function (`--pbkdf` option). So it's important to provide this as an
option for Calamares in order to make Linux installs more secure, for
those who wish to encrypt their system.

This commit was tested on a custom Manjaro installer with:

- grub bootloader with the [argon patches](https://aur.archlinux.org/packages/grub-improved-luks2-git).
- [rEFInd](https://wiki.archlinux.org/title/REFInd) bootloader with
  unencrypted `/boot` partition because rEFInd [doesn't support booting
  from an encrypted volume](https://sourceforge.net/p/refind/discussion/general/thread/400418ac/)

**Important consideration for distribution maintainers**:

- You need to have compile flag `WITH_KPMCORE4API` on
- If you are shipping with grub by default please note that you need to
  ship it with the Argon patches. Example on Arch Linux: [grub-improved-luks2-git](https://aur.archlinux.org/packages/grub-improved-luks2-git)
- If `luksGeneration` is not found in partition.conf, it will default to
  luks1
- Please test this on your own distribution as this was only tested on
  Manjaro installer (see above).

### DIFF
--- a/src/modules/partition/Config.cpp
+++ b/src/modules/partition/Config.cpp
@@ -46,6 +46,32 @@ Config::swapChoiceNames()
     return names;
 }
 
+const NamedEnumTable< Config::LuksGeneration >&
+Config::luksGenerationNames()
+{
+    static const NamedEnumTable< LuksGeneration > names { { QStringLiteral( "luks1" ), LuksGeneration::Luks1 },
+                                                          { QStringLiteral( "luks2" ), LuksGeneration::Luks2 } };
+
+    return names;
+}
+
+const QString
+Config::luksGenerationToFSName( Config::LuksGeneration luksGeneration )
+{
+    // Convert luksGenerationChoice from partition.conf into its
+    // corresponding file system type from KPMCore.
+    switch ( luksGeneration )
+    {
+    case Config::LuksGeneration::Luks2:
+        return QStringLiteral( "luks2" );
+    case Config::LuksGeneration::Luks1:
+        return QStringLiteral( "luks" );
+    default:
+        cWarning() << "luksGeneration not supported, defaulting to \"luks\"";
+        return QStringLiteral( "luks" );
+    }
+}
+
 Config::SwapChoice
 pickOne( const Config::SwapChoiceSet& s )
 {
@@ -323,12 +349,23 @@ Config::fillConfigurationFSTypes( const QVariantMap& configurationMap )
         }
     }
 
+    // Set LUKS file system based on luksGeneration provided, defaults to 'luks'.
+    bool nameFound = false;
+    Config::LuksGeneration luksGeneration
+        = luksGenerationNames().find( CalamaresUtils::getString( configurationMap, "luksGeneration" ), nameFound );
+    if ( !nameFound )
+    {
+        cWarning() << "Partition-module setting *luksGeneration* not found or invalid. Defaulting to luks1.";
+        luksGeneration = Config::LuksGeneration::Luks1;
+    }
+    m_luksFileSystemType = Config::luksGenerationToFSName( luksGeneration );
+    gs->insert( "luksFileSystemType", m_luksFileSystemType );
+
     Q_ASSERT( !m_eraseFsTypes.isEmpty() );
     Q_ASSERT( m_eraseFsTypes.contains( fsRealName ) );
     m_eraseFsTypeChoice = fsRealName;
     Q_EMIT eraseModeFilesystemChanged( m_eraseFsTypeChoice );
 }
-
 
 void
 Config::setConfigurationMap( const QVariantMap& configurationMap )
--- a/src/modules/partition/Config.h
+++ b/src/modules/partition/Config.h
@@ -63,6 +63,16 @@ public:
 
     using EraseFsTypesSet = QStringList;
 
+    /** @brief Choice of LUKS disk encryption generation */
+    enum LuksGeneration
+    {
+        Luks1,  // First generation of LUKS
+        Luks2,  // Second generation of LUKS, default since cryptsetup >= 2.1.0
+    };
+    Q_ENUM( LuksGeneration )
+    static const NamedEnumTable< LuksGeneration >& luksGenerationNames();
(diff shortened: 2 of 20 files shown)


src/modules/partition/Config.cpp
```diff
@@ -46,6 +46,32 @@ Config::swapChoiceNames()
     return names;
 }
 
+const NamedEnumTable< Config::LuksGeneration >&
+Config::luksGenerationNames()
+{
+    static const NamedEnumTable< LuksGeneration > names { { QStringLiteral( "luks1" ), LuksGeneration::Luks1 },
+                                                          { QStringLiteral( "luks2" ), LuksGeneration::Luks2 } };
```
adriaandegroot, line 53: You can use `// clang-format: off` if you want to write this in a less far-indented fashion, so it's readable and clang-format doesn't mess it up. You can also write aliases here, e.g. add *luks* as an alias for *luks1*. And if you list it first, it's the "default" for some operations.

src/modules/partition/Config.cpp
```diff
@@ -46,6 +46,32 @@ Config::swapChoiceNames()
...
+
+    return names;
+}
+
+const QString
+Config::luksGenerationToFSName( Config::LuksGeneration luksGeneration )
+{
```
adriaandegroot, line 60: Wouldn't this be simpler as `luksGenerationNames().find(luksGeneration)` ? The named enums are supposed to support lookups both forward and back.

src/modules/partition/Config.h
```diff
@@ -162,6 +181,7 @@ public Q_SLOTS:
     SwapChoiceSet m_swapChoices;
     SwapChoice m_initialSwapChoice = NoSwap;
     SwapChoice m_swapChoice = NoSwap;
+    QString m_luksFileSystemType;
```
adriaandegroot, line 184: I dunno, seems like an odd choice to have an enum, but store the string instead.

src/modules/partition/PartitionViewStep.cpp
```diff
@@ -616,7 +616,12 @@ PartitionViewStep::onLeave()
 
             // If the root partition is encrypted, and there's a separate boot
             // partition which is not encrypted
-            if ( root_p->fileSystem().type() == FileSystem::Luks && boot_p->fileSystem().type() != FileSystem::Luks )
+            if ( ( root_p->fileSystem().type() == FileSystem::Luks && boot_p->fileSystem().type() != FileSystem::Luks )
+#ifdef WITH_KPMCORE4API
```
adriaandegroot: Note that in Calamares 3.3 (which you're merging to), KPMCore4 is a given, and there is no `#define` for it anymore.

jtheoof: Good to know thanks. I'll remove the `#ifdefs`

src/modules/partition/Config.cpp
jtheoof: The other `NamedEnumTable`s of this file are formatted in this way (without formatting off) I opted for the same here for consistency within the file.

src/modules/partition/Config.cpp
jtheoof: Unless I misunderstand, it's not the same thing. The goal of this function is to convert the enum into the File System string set inside the `GlobalStorage` object. I used the `.find` function in line `#355`.

src/modules/partition/Config.h
jtheoof: I see your point. Originally I carried the `enum` all the way down to the `KPMHelpers.createNewEncryptedPartition` but that implied to include `Config.h` in a few places where I found less relevant, so I opted for a less complex storage but we are losing types I agree.

I see 3 options:
1. I rename `luksGeneration` into `luksFsType` with type: `QString` with possible values: `"luks" |  "luks2"` The enum doesn't bring a lot of value IMO.
2. I leave the enum be and carry it inside the `GlobalStorage` (will more complicated I think)
3. I leave the code as is

I would prefer 1 or 3 but could try 2 if you prefer.

src/modules/partition/Config.cpp
jtheoof: This function might be removed depending on the option you prefer in the comment below.

src/modules/partition/PartitionViewStep.cpp
jtheoof: Done

src/modules/partition/Config.h
adriaandegroot: Let's definitely **not** do 2. GlobalStorage is not meant to store enums. It can do whatever QVariant supports, (so with suitable Qt futzing you can "make it fit"), but let's stick to ints, bools and strings there. That way we also don't need to think too hard about the interface from GS to Python. So 2 is out.

If it is most convenient to have a string you can pass in to KPMCore APIs (or other tooling), then by all means keep that.

jtheoof:
Thanks for the review @adriaandegroot. Based on the comments, I'm not sure what remains to be done in order to move forward with this MR. Can you provide some guidance for the next steps?

jtheoof:
Any news on the next steps for this @adriaandegroot ?

adriaandegroot:
Very much a "I need to sit down with at least 20 minutes for Calamares to look at it". Which hasn't been the case due to Akademy and other issues.

adriaandegroot:
I'm going to merge as-is and then hit some things with a hammer (and PR that, and then you can comment there :) )

PR #2051, opened 2022-09-24
fix initcpiocfg filesystems loop

ptr1337:
Fixes current issue, that initcpiocfg does add several filesystems into mkinitcpio.conf

Signed-off-by: Peter Jung <admin@ptr1337.dev>


dalto8 (approved):

dalto8:
Looks good to me.  Also fixes the order issue with `openswap` and `resume`.

PR #2054, opened 2022-10-06
Development scripts

es20490446e:
I have developed a couple of small scripts that makes building and testing calamares the fastest.

Let me know if you are interested in such thing 👍

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -115,11 +115,11 @@ setArguments () {
 
 so () {
 	local tag="-${FUNCNAME[2]}"
-	local commands="${@:1}"
+	local commands="${*:1}"
 	local maxLines=20
 	local error=""
 
-	if ! error="$(eval "${commands[@]}" 2>&1 >"/dev/null")"; then
+	if ! error="$(eval "${commands}" 2>&1 >"/dev/null")"; then
 		if [[ -z "${error}" ]] ; then
 			error="Command failed: ${commands[*]}"
 		else
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,140 @@
+#! /bin/bash
+: '
+For building Calamares using this program, in the application "Terminal" type:
+"/pathToThisFolder/build.sh" "[arguments]"
+
+Where "[arguments]" can either be:
+
+- The build type: "Release" or "Debug"
+- The building flags, for example: -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=/usr
+'
+
+here="$(realpath "$(dirname "${0}")")"
+build="${here}/_BUILD"
+
+
+mainFunction () {
+	enterBuildDir
+	configureBuild
+	compile
+}
+
+
+checkDependencies () {
+	local name=""
+	local location=""
+	local dependency=""
+	local missing=()
+
+	local dependencies=(
+		boost@/usr/bin/b2
+		cmake@/usr/bin/cmake
+		extra-cmake-modules@/usr/share/ECM/cmake/ECMConfig.cmake
+		gcc@/usr/bin/gcc
+		kcoreaddons@/usr/bin/desktoptojson
+		kpmcore@/usr/include/kpmcore/backend/corebackend.h
+		python3@/usr/bin/python3
+		qt@/usr/bin/qmake
+		yaml-cpp@/usr/include/yaml-cpp/binary.h
+	)
+
+	for dependency in "${dependencies[@]}"; do
+		name="$(echo "${dependency}" | cut --delimiter='@' --fields=1)"
+		location="$(echo "${dependency}" | cut --delimiter='@' --fields=2)"
+
+		if [[ ! -f "${location}" ]]; then
+			missing+=("${name}")
+		fi
+	done
+
+	if [[ "${#missing[@]}" -ne 0 ]]; then
+		echo "missing dependencies: ${missing[*]}" >&2
+		exit 1
+	fi
+}
+
+
+compile () {
+	local cores; cores="$(grep -c ^processor /proc/cpuinfo)"
+	local jobs; jobs="$(("${cores}" + "${cores}" / 2))"
+	local load; load="$(optimalLoad "${cores}")"
+
+	so make --jobs="${jobs}" --load="${load}"
+}
+
+
+configureBuild () {
+	# shellcheck disable=SC2068
+	so cmake .. ${arguments[@]}
+}
+
+
+enterBuildDir () {
+	makeBuildDir
+	cd "${build}"
+}
+
+
+makeBuildDir () {
+	if [[ ! -d "${build}" ]]; then
+		mkdir --parents "${build}"
+	fi
+}
+
+
+optimalLoad () {
+	local cores="${1}"
+
+	if [[ "${cores}" -eq 1 ]]; then
+		echo 1
+	else
+		echo "$(("${cores}" * 75 / 100))"
+	fi
+}
+
+
+prepareEnvironment () {
+	set -e
+	renice --priority 10 "${$}" >/dev/null
+	checkDependencies
+	setArguments "${@}"
+}
+
+
+setArguments () {
+	arguments=("${@}")
+
+	if [[ "${#arguments}" -eq 0 ]]; then
+		echo "no build type specified as argument: Release Debug" >&2
+		exit 1
+	elif [[ "${arguments[*]}" == "Release" ]] || [[ "${arguments[*]}" == "Debug" ]]; then
+		arguments=("-DCMAKE_BUILD_TYPE=${arguments[*]}")
+	fi
+}
+
+
+so () {
+	local tag="-${FUNCNAME[2]}"
+	local commands="${@:1}"
+	local maxLines=20
+	local error=""
+
+	if ! error="$(eval "${commands[@]}" 2>&1 >"/dev/null")"; then
+		if [[ -z "${error}" ]] ; then
+			error="Command failed: ${commands[*]}"
+		else
+			if [[ "$(echo "${error}" | wc --lines)" -gt "${maxLines}" ]]; then
+				error="$(echo "${error}" | tail -n "${maxLines}")"
+				error="$(printf "...\n%s", "${error}")"
+			fi
+
+			echo "${tag}${error}" >&2
+		fi
+
+		exit 1
+	fi
+}
+
+
+prepareEnvironment "${@}"
+mainFunction "${@}"
--- a/build.sh
+++ b/build.sh
@@ -27,15 +27,50 @@ checkDependencies () {
 	local missing=()
 
 	local dependencies=(
-		boost@/usr/bin/b2
+		ack@/usr/bin/vendor_perl/ack
+		autoconf@/usr/bin/autoconf
+		automake@/usr/bin/automake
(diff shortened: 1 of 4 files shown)


adriaandegroot:
While I appreciate the effort, this seems like a lot of words for very little improvement: instead of `cmake` followed by `make` (or `ninja`), it's one script, **but** the script makes assumptions about the host system organization, doesn't have a complete dependency list, uses bash (that's not **that** bad, but everything else tries to be `/bin/sh` compatible in the Calamares repo) and is missing SPDX license information.

The `deploycala.py` script already does all that (it is not in **this** repository, though, and it has some specific caveats as well).

If this would fit anywhere, it would go in the `ci/` directory, and come with updates to documentation saying "here is how to build".

es20490446e:
I will have a look.

es20490446e:
What I don't like of `deploycala.py` actually is that installs everything automatically, so rolling back the install once you no longer need it gets messy.

Hence I think that listing the dependencies instead, and allowing the user to manually install them, will be preferred for most people.

Also it has the advantage that you don't really have to take into consideration every single distro out there. As you just check for the files, not for package names.

What do you think?

es20490446e:
It will also help me if you pointed me out which assumptions do you think won't be portable across the systems that can use calamares.

I guess that all systems will store their binaries under `/usr/bin`, won't they?

Maybe not for `/usr/share` or `/usr/include`?

But are such systems using currently calamares, or likely to use it?

Does some of those really need to use `/bin/sh` instead? This is a relevant question, because you will lose arrays and loops over them.

es20490446e:
Finally coding this particular scenario in Bash, instead of Python, looks simpler to me.

As the only thing you are doing is working with files, and calling other commands.

I think people usually prefer Python for system administration just because they don't use functions in Bash, but write everything in a big ball of mud.

Just by breaking the code into the smallest functions, naming variables properly, and throwing meaningful errors, Bash looks like a complete different language (than what I see out there).

For example the function `so` in `build.sh` will make any command failure to tell its arguments, error message, and which functions called it. Without needing any special error handling coding.

dalto8:
I agree that this adds very little value to Calamares.  It is more like a script you would use on your own than something that is generally beneficial to Calamares users in general.

> think people usually prefer Python for system administration just because they don't use functions in Bash, but write everything in a big ball of mud.
> 
> Just by breaking the code into the smallest functions, naming variables properly, and throwing meaningful errors, Bash looks like a complete different language (than what I see out there).

I don't think anyone is proposing that you rewrite this in Python.  That wouldn't really solve the core problem.

es20490446e:
Sometimes small improvements can lead to huge benefits, when they are in activities that you do quite often.

Give it a try, and let me know your decision.

es20490446e:
[deploycala.py](https://github.com/calamares/calamares.github.io/blob/calamares/deploycala.py)

chihero (approved):

build.sh
```diff
@@ -0,0 +1,175 @@
...
+checkDependencies () {
+	local name=""
+	local location=""
+	local dependency=""
+	local missing=()
+
+	local dependencies=(
+		ack@/usr/bin/vendor_perl/ack
+		autoconf@/usr/bin/autoconf
+		automake@/usr/bin/automake
```
adriaandegroot, line 32: None of these are build dependencies of Calamares itself, so why are they here?

build.sh
```diff
@@ -0,0 +1,175 @@
+#! /bin/bash
+: '
+For building Calamares using this program, in the application "Terminal" type:
+"/pathToThisFolder/build.sh" "[arguments]"
```
adriaandegroot, line 4: But this does not just build Calamares (which would be `ninja -C build` for me in my development setup), it does other things as well.

build.sh
```diff
@@ -0,0 +1,175 @@
...
+		automake@/usr/bin/automake
+		binutils@/usr/bin/addr2line
+		bison@/usr/bin/bison
+		boost-dev@/usr/include/boost/accumulators/accumulators.hpp
+		boost-python3-dev@/usr/include/boost/mpi/python.hpp
+		cmake@/usr/bin/cmake
+		extra-cmake-modules@/usr/share/ECM/cmake/ECMConfig.cmake
+		fakeroot@/usr/bin/fakeroot
+		flex@/usr/bin/flex
+		gcc@/usr/bin/gcc
```
adriaandegroot, line 41: **A** C++ compiler is needed, but not necessarily gcc.

dalto8:
Closing for the reasons mentioned above.

This would need quite a bit of work to get to a spot where it would be suitable.  Even if that work was done, this doesn't seem like something that is broadly needed.  Instead, it is more suitable to be a script appropriate for your individual use.

es20490446e:
When you do something frequently it is important to automate it, because even when it is small after a hundred times it starts mattering. Specially when you take into account a broad set of software.

It is sad that developers rarely see how important this is, and most of the work they do in a day is actually just waste. Work that actually doesn't improve code, but it is just recurring silly pity details and maintenance.

kkofler:
There are just too many assumptions in that script to be useful for the intended purpose (making it easier for new developers to get started).

As was already pointed out by others, you assume GCC (as opposed to, e.g., Clang), make (as opposed to, e.g., Ninja), file system locations that are wrong on many distributions (e.g., no sane distribution installs Qt 5 headers to `/usr/include/qt`, but to `/usr/include/qt5`), etc. Your dependency checks are also unnecessary because CMake checks for those dependencies already (or if it does not, it should, so file a bug here to get a check added to the Calamares `CMakeLists.txt`), and it does so in a more portable way than your script. I do not see how it is more helpful to get an error message about a missing dependency from your script than from CMake.

And there is not much that your script does other than checking dependencies. Just running `cmake .` and then either `cmake --build .` or `make` or `ninja` directly is something I can do without any script (and for any CMake project, not just Calamares). You hardcode a few parameters to `cmake`, but what if I want to pass other options? I am better off writing my own 2-line script with the exact CMake parameters I need or just saving a `CMakeUserPresets.json` from the CMake GUI.

es20490446e:
I will have a look.

es20490446e:
> I do not see how it is more helpful to get an error message about a missing dependency from your script than from CMake.

If a dependency is missing `cmake` will complain about the first missing one, not all of them.

In practice that means you will be installing those one by one, or checking those one by one, instead of just getting a list of what is missing.

That's why the script, checking those dependencies each single time you have to build is a silly waste of time.


> Just running `cmake .` and then either `cmake --build .` or `make` or `ninja` directly is something I can do without any script.

You here mean:
```
mkdir --parents build
cd build
cmake --build ..
```

Then you have to run `make` with the right `--jobs` and `--load` arguments, which basically nobody knows what they are.

Otherwise you either get `make` using only one core, or it overloading your CPU and making it unusable for anything else.


> you assume GCC (as opposed to, e.g., Clang), make (as opposed to, e.g., Ninja)

[It doesn't matter](https://signalvnoise.com/archives2/it_just_doesnt_matter.php). Whatever you install will do the same job. It doesn't matter if you need to have installed one or the other.

Taking into account that diversity does nothing.


> What if I want to pass other options? I am better off writing my own 2-line script with the exact CMake parameters I need

How is that easier than just adding those parameters to the existing script?


> file system locations that are wrong on many distributions

That's something that can be easily fixed, if we ever conclude we are using these scripts.

build.sh
es20490446e: Some module failed to build without them.

build.sh
es20490446e: [It doesn't make a difference](https://signalvnoise.com/archives2/it_just_doesnt_matter.php).

build.sh
es20490446e: Checking dependencies is part of building.

kkofler:
Hey, can you please accept that we are **NOT** interested in this contribution? You have got 3 different Calamares developers all telling you that it is not accepted as is and that we generally do not consider it useful and why (and not a single one supporting your change). Now you are just wasting your and our time arguing.

> > I do not see how it is more helpful to get an error message about a missing dependency from your script than from CMake.
> 
> If a dependency is missing `cmake` will complain about the first missing one, not all of them.
> 
> In practice that means you will be installing those one by one, or checking those one by one, instead of just getting a list of what is missing.

I do not see the big issue there. It does not take much longer to install one package at a time than a list of packages. Or in your own words: "It doesn't matter". :-)

> That's why the script, checking those dependencies each single time you have to build is a silly waste of time.

What do you mean with that? If you install the dependencies once, you have them and do not have to do it again the next time you build. If a new dependency was added and is missing on your system, CMake will tell you.

But if you mean that it is a waste to run all the CMake checks, well, your script adds another set of checks on top of that, wasting more time (but both the CMake checks and your script's checks should run in milliseconds at worst on a modern computer).

The problem is, the script is **not** portable and does **not** work on the distributions I or other developers are using. Checking for dependencies portably is complicated and hence best left to CMake, which knows how to do it. You cannot force all Calamares developers to use the same distribution you are using.

> > Just running `cmake .` and then either `cmake --build .` or `make` or `ninja` directly is something I can do without any script.
> 
> You here mean:
> 
> ```
> mkdir --parents build
> cd build
> cmake --build ..
> ```

* `--parents` is useless here, since you are just creating one level of subdirectory in an already existing directory (and besides, everyone uses the short form `-p` for that).
* I always build in tree. If the project refuses to do that, I locally delete the lines that enforce that. I have yet to find a single project that does not build after removing that completely unnecessary check. If you think creating a subdirectory is too complicated, you really want to get that bogus check removed, not wrapper scripts added.

> Then you have to run `make` with the right `--jobs` and `--load` arguments, which basically nobody knows what they are.
> 
> Otherwise you either get `make` using only one core, or it overloading your CPU and making it unusable for anything else.

Sometimes, using only one core is what you want, because it makes the error output less scrambled than in parallel. Also, the "right" `-j` argument depends on the computer and cannot be hardcoded in a script.

> > you assume GCC (as opposed to, e.g., Clang), make (as opposed to, e.g., Ninja)
> 
> [It doesn't matter](https://signalvnoise.com/archives2/it_just_doesnt_matter.php). Whatever you install will do the same job. It doesn't matter if you need to have installed one or the other.
>
> Taking into account that diversity does nothing.

It matters because:

* the developer can have one already installed and not want to install the other just because your script prefers it, when the project builds fine with what they already have,
* Clang produces different warnings than GCC, can be faster or slower to compile things, can produce faster or slower, larger or smaller binaries, or even, only one might work on your system because the other is too old or too new in your distribution or might not be available at all (if the distribution is similarly opinionated as your script),
* Ninja is typically faster than make, but make is installed by default in many places,

so it is a quality of implementation matter to support both alternatives in both cases.

And anyway, "it doesn't matter" is not a KDE design principle, it is more the way GNOME or Apple work.

> > What if I want to pass other options? I am better off writing my own 2-line script with the exact CMake parameters I need
> 
> How is that easier than just adding those parameters to the existing script?

The existing script does not work on my distribution and does things I do not need.

> > file system locations that are wrong on many distributions
> 
> That's something that can be easily fixed, if we ever conclude we are using these scripts.

Not easily. Files can be in different places on different distributions. You end up reinventing the whole CMake dependency checking mechanism for no good reason.

Your script also needs to be updated each time a dependency is added, changed, or removed. Duplicating logic is never a good idea.

So please accept a "no" and quit spamming us.

PR #2056, opened 2022-10-14
[dracut][zfs] adding hostid

demmm:
using ZFS in combination with dracut exposed a bug on system updates hostid from the Live session does not match hostid installed, thus zpool id no longer matches id created by dracut in the kernel img
to work around this, the zfs module now uses zgenhostid to create a hostid, which is later copied, using the dracut module.
if this issue is never encountered using mkinitcpio, it might be best to close this PR otherwise, a better place to copy /etc/hostid need to be found.

### DIFF
--- a/CHANGES-3.3
+++ b/CHANGES-3.3
@@ -29,13 +29,15 @@ This release contains contributions from (alphabetically by first name):
 
 ## Modules ##
  - *dracut* added a configurable kernel name. (thanks Anke)
+ - *dracut* copies hostid file, when ZFS is in use.
  - *initcpiocfg* orders hookds slightly differently. (thanks Peter)
  - *localeq* moved to using Drawer instead of ComboBox in UI. (thanks Anke)
  - *keyboardq* moved to using Drawer instead of ComboBox in UI. (thanks Anke)
  - *netinstall* now has a new *noncheckable* option for groups, which prevent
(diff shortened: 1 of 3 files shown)


dalto8:
Shouldn't we be copying the hostid all the time, not only when using dracut?

demmm:
As in my commit message, if issue occurs in only dracut, then best close this PR, if needed by more (mkinitcpio?), a better place to copy hostid needs to found.

demmm:
I do not see one module that makes sense to add the copy hostid too.  zfs is too early, mount is kind of an odd module to do this, so to me it makes most sense to duplicate the code as used in this PR for dracut, and add it the other modules (as needed?) that create the kernel images/initramfs (initcpio & mkinitfs).
If duplicating code is not an option, then it will be best to close this PR.

dalto8:
I have been thinking about this too.

It seems like the only practical options are either to force it into mount, or add a separate zfshostid module.

I don't think duplicating the code makes sense.  In addition to the obvious problems with duplicate code, because we have no control of which modules people use, and since the hostid isn't actually related to dracut/mkinitcpio forcing it into those modules seems even worse than using mount.

demmm:
>t seems like the only practical options are either to force it into mount, or add a separate zfshostid module.

That is the best option, will work on that

PR #2057, opened 2022-10-15
[dracut] Change image name terminology to match dracut

dalto8:
A recently introduced change to the dracut module added a config option named `kernelName`.  However, what it represented was not the name of the kernel, but the name of the output image.

This PR changes the terminology to match what dracut uses, "initramfs".

### DIFF
--- a/src/modules/dracut/main.py
+++ b/src/modules/dracut/main.py
@@ -42,11 +42,13 @@ def run_dracut():
         try:
             target_env_process_output(['dracut', '-f'])
         except subprocess.CalledProcessError as cpe:
-            return cpe.returncode, cpe.output
+            libcalamares.utils.warning(f"Dracut failed with output: {cpe.output}")
+            return cpe.returncode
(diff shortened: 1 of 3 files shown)


src/modules/dracut/main.py
```diff
@@ -53,4 +53,4 @@ def run():
 
     if return_code != 0:
         return (_("Failed to run dracut on the target"),
-                _("The exit code was {}").format(return_code))
+                _(f"The exit code was {return_code}"))
```
demmm: Not clear why the `f` here, causes invalid python code on my end.

dalto8: That is a python "f string".  As of Calamares 3.3 we are using those.  What version of Python are using?

As a side note, when I was testing that, I noticed an exception handling bug which I am fixing now.

dalto8: Should be working fully now

PR #2058, opened 2022-10-15
[mount] Remove duplicate subvol option

dalto8:
In 3.3, both the mount module and the fstab module are adding the `subvol=` option.

Remove the copy from mount.


ptr1337:
> In 3.3, both the mount module and the fstab module are adding the `subvol=` option.
> 
> Remove the copy from mount.

Does fix the current issue, tested it with the latest commit including this patch on a arch based distro

PR #2059, opened 2022-10-16
Hostid

demmm:


src/modules/zfshostid/zfshostid.schema.yaml
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: 2022 Anke Boersma <demm@kaosx.us>
+# SPDX-License-Identifier: GPL-3.0-or-later
+---
+$schema: https://json-schema.org/schema#
+$id: https://calamares.io/schemas/zfshostid
+additionalProperties: false
+type: object
+
+required:
+    - zfs
```
dalto8: I think this means that the item "zfs" is required.  Not the module zfs.  I am not sure that you need a schema for a module which has no config.

I think module requirements are handled by setting `requiredModules` in module descriptor.

src/modules/zfshostid/main.py
```diff
@@ -0,0 +1,45 @@
...
+def run():
+
+    zfs = libcalamares.globalstorage.value("zfsDatasets")
+    root_mount_point = libcalamares.globalstorage.value("rootMountPoint")
+
+    if zfs:
+        hostid_source = '/etc/hostid'
+        hostid_destination = '{!s}/etc/hostid'.format(root_mount_point)
+
+        if os.path.exists(hostid_source):
```
dalto8: I am not sure this "if" is needed.  It causes the module to fail without giving an error/warning.

If you remove it, the file being missing should cause an exception which will then write a warning per your exception handler.

dalto8:
I made a couple of small comments on the code itself but I thinks this look good.  I like this approach.

PR #2065, opened 2022-10-29
fixed obvious mistakes in the French translation

vjestin:


demmm:
Translations are done in Transifex, not by PR:
Can you correct there?

adriaandegroot:
Yes, now that you point at them they are obvious typo's. As demmm points out, the translations are done in Transifex, so joining there is the most effective way to improve the French translation. We **can** take translation fixes this way, but it is a huge amount of extra work.

vjestin:
No problem. I should have read the translation instructions first. I will redo the work in Transifex.

PR #2066, opened 2022-10-30
[zfs] make the zfs module require zfshostid

demmm:
with https://github.com/calamares/calamares/commit/1b96832bf7e60a849119975adcb5fa7959e05e3e a unique hostid is created, used during zpool creation, when that hostid file is not copied to the new install, the mismatch in hostid call will create boot issues, thus the zfshostid module is now required

### DIFF
--- a/src/modules/zfs/CMakeLists.txt
+++ b/src/modules/zfs/CMakeLists.txt
@@ -9,4 +9,5 @@ calamares_add_plugin(zfs
     SOURCES
         ZfsJob.cpp
     SHARED_LIB
+    REQUIRES zfshostid
 )


adriaandegroot:
From Telegram it seems like you would like to have a discussion about this one. It is technically straightforward and looks obviously correct (and it takes just one test of editing `settings.conf` to confirm).

dalto8 (approved):

dalto8:
Since we are now properly creating a hostid, it seems copying it to the host is mandatory so this makes sense to me.

demmm:
Sorry, not properly tested, can't be used as is.  I did check the created module.desc in the build directory, which is fine:
```
# AUTO-GENERATED metadata file
# Syntax is YAML 1.2
---
type: "job"
name: "zfs"
interface: "qtplugin"
load: "libcalamares_job_zfs.so"
requiredModules:
 - zfshostid
 ```

Removing zfshostid from the settings.conf worked as should too, message:
`ERROR: STARTUP: failed modules are ("zfs@zfs")`

But I did not check further up yesterday in the log, issue is now that zfshostid is expected to run before zfs (which is not posible, needs to run after mounting):
`ERROR: Module "zfs" requires "zfshostid" before it in sequence.`

Closing this PR, trying to find a working solution.

PR #2069, opened 2022-11-05
[grubcfg] Fix bugs and add option to force defaults to update existing values

dalto8:
This PR does a few things
* Fix a bug that was causing existing values in `GRUB_CMDLINE_LINUX_DEFAULT` to not be preserved.
* Add a new config option `always_use_defaults`
* General code cleanup on the sections I modified

The new option forces the options set to overwrite any existing settings.  This basically means that any value in the grub config can now be set using the module if desired.

This change should be backwards compatible with existing config files


src/modules/grubcfg/grubcfg.conf
```diff
@@ -46,3 +46,6 @@ defaults:
     GRUB_DISABLE_SUBMENU: true
     GRUB_TERMINAL_OUTPUT: "console"
     GRUB_DISABLE_RECOVERY: true
+
+# Set to true to force defaults to be used even when not overwriting
+always_use_defaults: false
```
adriaandegroot: Needs a newline at end (check your editor configuration)

src/modules/grubcfg/main.py
```diff
@@ -85,6 +86,30 @@ def get_zfs_root():
     return None
 
 
+def update_existing_config(default_grub, grub_config_items):
+    """
+    Updates the existing grub configuration file with any items present in @p grub_config_items
```
adriaandegroot, line 91: The code just prints things, there is no writing to the file that I can see (or `fileinput` is **really* fancy and I don't know about it -- is it always available?).

src/modules/grubcfg/main.py
```diff
@@ -190,62 +215,45 @@ def modify_grub_default(partitions, root_mount_point, distributor):
...
 
+    always_use_defaults = libcalamares.job.configuration.get("always_use_defaults", False)
+
+    # If applicable add the items from defaults to the dict containing the grub config to wirte/modify
+    if always_use_defaults or overwrite or not os.path.exists(default_grub):
         if "defaults" in libcalamares.job.configuration:
-            for key, value in libcalamares.job.configuration[
-                    "defaults"].items():
+            for key, value in libcalamares.job.configuration["defaults"].items():
                 if value.__class__.__name__ == "bool":
```
adriaandegroot: `isinstance(value, bool)` ? (Not your code, I know)

adriaandegroot (changes requested):

adriaandegroot:
It would be nice if we also had tests demonstrating the parts of the code now factored-out (and a test would convince me that update-in-place function does the right thing).

src/modules/grubcfg/grubcfg.conf
dalto8: Fixed

src/modules/grubcfg/main.py
dalto8: Yeah, I guess it is fancy by that logic. :)

It takes the output and uses it to update the file in place.

src/modules/grubcfg/main.py
dalto8: fixed

JustTNE:
What is this PR blocked by at this moment?

dalto8:
It is good to go from my perspective.  We have been using on a recent ISO so it has had a lot of testing.

ptr1337:
Also merged at CachyOS since 2 Months. all good

PR #2070, opened 2022-11-05
Replace luks-is-a-string by an enum

adriaandegroot:
This tightens up merge request #2047, by replacing a QString -- used to pass the string "luks1" or "luks2" around until it finally reaches the KPMCore layer -- by an enum value. While here, remove some code that was written in 2015 and never used.


jtheoof:
Looking good. Some commits could be squashed together from my personal preference but otherwise thanks for looking after my foolish cpp attempt.

dalto8 (approved):

PR #2079, opened 2022-12-10
Don't install default settings.conf to CMAKE_BINARY_DIR

### LINKED ISSUE #2075: settings.conf installed in strange place
There is a file settings.conf which is installed into the build directory

 https://github.com/calamares/calamares/commit/8f72d2541f241ef5a1f4bc945bdc908e7dbe9fdd

https://github.com/calamares/calamares/blob/calamares/CMakeLists.txt#L589 

e.g. when I build locally it gets installed into `/home/jr/src/extras/calamares/github/calamares/build/settings.conf` instead of /usr/share/calamares or similar

What is this madness?

dalto8:
Installing a default `settings.conf` to `CMAKE_BINARY_DIR` causes packaging challenges.

I am proposing removing this because I don't have visibility into the situation that would require this.  If there is a need for this behavior, I think it should not be the default and require an option to enable it.

This fixes https://github.com/calamares/calamares/issues/2075.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -585,9 +585,6 @@ install(
 ### Miscellaneous installs
 #
 #
-if( settings.conf IS_NEWER_THAN ${CMAKE_BINARY_DIR}/settings.conf )
-    install(FILES settings.conf DESTINATION ${CMAKE_BINARY_DIR})
-endif()
 
 if(INSTALL_POLKIT)
     install(FILES com.github.calamares.calamares.policy DESTINATION "${POLKITQT-1_POLICY_FILES_INSTALL_DIR}")


kkofler:
-1. We should revert commit 8f72d2541f241ef5a1f4bc945bdc908e7dbe9fdd instead.

dalto8:
I don't think it should be reverted.  From my perspective removing the ability to install the default config was a good change overall.

If a distro wants to install a custom config, they can do that with a patch when they patch the config files.

kkofler:
To elaborate on this: Installing this file (and only this file) to CMAKE_BINARY_DIR (a behavior introduced by commit 8f72d2541f241ef5a1f4bc945bdc908e7dbe9fdd) indeed does not make sense, but we need the INSTALL_CONFIG option back (which was removed by commit 8f72d2541f241ef5a1f4bc945bdc908e7dbe9fdd) that allows installing this file to the proper place and together with the other required settings files.

kkofler:
I do not see why we should patch the build system with a downstream revert of commit 8f72d2541f241ef5a1f4bc945bdc908e7dbe9fdd when this can easily be handled with a CMake setting as it has been all this time. It just means more work for distribution packagers for no good reason.

dalto8:
Because there is a good reason.  Calamares doesn't need and shouldn't have the ability to install the default config.  It makes no sense to do so.

I would argue that patching the default module configs is the wrong thing for a distro to do in the first place and is explicitly the opposite of what Calamares recommends.

However, if a distro wishes to do that, they need only add a handful of additional lines to their patch(es).

kkofler:
> Because there is a good reason. Calamares doesn't need and shouldn't have the ability to install the default config. It makes no sense to do so.

Which is why the CMake option has been off by default for years.

> I would argue that patching the default module configs is the wrong thing for a distro to do in the first place and is explicitly the opposite of what Calamares recommends.

And I have never agreed with that recommendation. I think patching the configs to adjust them to the distribution's needs makes a lot more sense than shipping configuration files separately from the application package. We do not do that for any other application. Why does the installer need to be special there? Distribution-specific settings for Calamares belong into the distribution package of Calamares, not in some random other package.

> However, if a distro wishes to do that, they need only add a handful of additional lines to their patch(es).

And I do not see why we should need to maintain those additional changes (which touch at least 3 CMake input files – commit 8f72d2541f241ef5a1f4bc945bdc908e7dbe9fdd has removed a few lines from 3 files: one `CMakeLists.txt` and two `*.cmake`) instead of just setting an option that was already there upstream and has been removed without consulting us for a reason that we do not agree with.

PR #2081, opened 2022-12-18
Add "kms" hook to initcpiocfg

ptr1337:
The "kms" hook got added with commit¹ to the default hooks array. Follow the archlinux defaults and add it also.

1. https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/commit/b99eb1c0d5704ed0a0d220d0abcbbbeec10a3b76

Signed-off-by: Peter Jung <admin@ptr1337.dev>

### DIFF
--- a/src/modules/initcpiocfg/main.py
+++ b/src/modules/initcpiocfg/main.py
@@ -146,8 +146,8 @@ def find_initcpio_features(partitions, root_mount_point):
     hooks = [
         "base",
         "udev",
-        "kms",
         "autodetect",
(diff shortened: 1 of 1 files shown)


dalto8:
Is it safe to put that hook before autodetect?

ptr1337:
> Is it safe to put that hook before autodetect?

It worked so far. But I moved it.  Thanks for pointing

dalto8 (approved):

PR #2083, opened 2022-12-22
Add refind, bootloader selection and general improvements

dalto8:
This PR represents several changes and improvements to the existing bootloader module
* Adds support for refind
* Modifies the kernel option building for systemd-boot and refind to be more in-line with the way options are built for grub
* Adds the ability to optionally select a bootloader dynamically via a global storage variable
* Changes the systemd-boot to call `kernel-install`, prior to this, it was installing in a kernel-install compatible fashion manually.  This has a number of implications:
	* It is much more distro agnostic as `kernel-install` will automatically do the right thing for the specific distro
	* It moves a few configuration items to systemd such as kernel names and entry names, this makes it further distro agnostic
	* It allows for the full suite of customization offered by `kernel-install`  in addition to what Calamares offers
	* It is more "future proof" as Calamares no longer needs to mirror changes in systemd(because `kernel-install` is part of systemd)
* Adds support for customization the `loader.conf` for systemd-boot
* Adds functionality for using flexibile kernel matching sematics for systemd-boot
* Adds support for configuring kernel options for systemd-boot and refind(this is already supported for grub)
* Fix zfs support for dracut

These changes have been tested extensively over the last 30 days.

### DIFF
--- a/src/modules/bootloader/bootloader.conf
+++ b/src/modules/bootloader/bootloader.conf
@@ -9,25 +9,31 @@
 # should specifically set *efiBootloaderId* to "debian" because that is
 # hard-coded in `grubx64.efi`.
 ---
+# A variable from global storage which overrides the value of efiBootLoader
+#efiBootLoaderVar: "packagechooser_bootloader"
+
 # Define which bootloader you want to use for EFI installations
 # Possible options are 'grub', 'sb-shim' and 'systemd-boot'.
 efiBootLoader: "grub"
 
-# systemd-boot configuration files settings, set kernel search path, kernel name
-# and amount of time before default selection boots
+# systemd-boot configuration files settings
+
+# kernelSearchPath is the path relative to the root of the install to search for kernels
+# A kernel is identified by finding files which match regular expression, kernelPattern
 kernelSearchPath: "/usr/lib/modules"
-kernelName: "vmlinuz"
-timeout: "10"
+kernelPattern: "^vmlinuz.*"
 
-# additionalInitrdFiles is a comma seperated list of file names
-additionalInitrdFiles:
-    - "/boot/amd-ucode"
-    - "/boot/intel-ucode"
+# loaderEntries is an array of options to add to loader.conf for systemd-boot
+# please note that the "default" option is added programmatically
+loaderEntries:
+  - "timeout 5"
+  - "console-mode keep"
 
-# Optionally set the menu entry name to use in systemd-boot.
-# If not specified here, these settings will be taken from branding.desc.
-#
-# bootloaderEntryName: "Generic GNU/Linux"
+# systemd-boot and refind support custom kernel params
+kernelParams: [ "quiet" ]
+
+# A list of kernel names that refind should accept as kernels
+#refindKernelList: [ "linux","linux-lts","linux-zen","linux-hardened" ]
 
 # GRUB 2 binary names and boot directory
 # Some distributions (e.g. Fedora) use grub2-* (resp. /boot/grub2/) names.
--- a/src/modules/bootloader/bootloader.conf
+++ b/src/modules/bootloader/bootloader.conf
@@ -13,7 +13,7 @@
 #efiBootLoaderVar: "packagechooser_bootloader"
 
 # Define which bootloader you want to use for EFI installations
-# Possible options are 'grub', 'sb-shim' and 'systemd-boot'.
+# Possible options are 'grub', 'sb-shim', 'refind` and 'systemd-boot'.
 efiBootLoader: "grub"
 
 # systemd-boot configuration files settings
--- a/src/modules/bootloader/bootloader.schema.yaml
+++ b/src/modules/bootloader/bootloader.schema.yaml
@@ -6,17 +6,13 @@ $id: https://calamares.io/schemas/bootloader
 additionalProperties: false
 type: object
 properties:
+    efiBootLoaderVar: { type: string }
     efiBootLoader: { type: string }
     kernelSearchPath:  { type: string }
     kernelName:  { type: string }
-    timeout:  { type: string }  # Inserted verbatim
-    additionalInitrdFiles:
-        type: array
-        items:
-            type: string
-    bootloaderEntryName:  { type: string }
-    kernelLine:  { type: string }
-    fallbackKernelLine:  { type: string }
+    kernelParams: { type: array, items: { type: string } }
+    loaderEntries: { type: array, items: { type: string } }
+    refindKernelList: { type: array, items: { type: string } }
 
     # Programs
     grubInstall:  { type: string }
@@ -27,12 +23,3 @@ properties:
 
     efiBootloaderId:  { type: string }
     installEFIFallback: { type: boolean }
-
-required:
-    - efiBootLoader
-    - kernelSearchPath
-    - kernelName
-    - grubInstall
-    - grubMkconfig
-    - grubCfg
-    - grubProbe
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -20,7 +20,9 @@
 #   Calamares is Free Software: see the License-Identifier above.
 #
 
+import fileinput
 import os
+import re
 import shutil
 import subprocess
 
@@ -28,8 +30,8 @@ import libcalamares
 
 from libcalamares.utils import check_target_env_call
 
-
 import gettext
+
 _ = gettext.translation("calamares-python",
                         localedir=libcalamares.utils.gettext_path(),
                         languages=libcalamares.utils.gettext_languages(),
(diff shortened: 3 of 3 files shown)


demmm:
Probably I am alone, but this is in my opinion complete the opposite of what should happen with the bootloader module, refind is already split in a separate module, bootloader should be split more, so I won't approve this PR. 
https://github.com/calamares/calamares-extensions/tree/calamares/modules/refind

dalto8:
I originally went in with thinking of splitting them up.  However, it would have created too much duplicated code and configuration to do it that way.

The kernel option building, the bootloader selection, etc, etc.

Really the grubcfg module should be moved here as well but I thought that might be too radical of a change for people to accept.

demmm:
Just a small extra note regarding modules. When Calamares originally started, it was a clear agreement from the start to keep all modules doing one thing only (just the partition module was kept out of this agreement).

kkofler:
The `bootloader` module does one thing only: setting up the bootloader. I agree that `grubcfg` should be part of it. And it does not make sense to duplicate the code for legacy BIOS fallback between different modules for different UEFI bootloaders. As long as we need that fallback, a unified module is the best approach to avoid copypasta.

demmm:
No, bootloader is not one thing, it runs grub, systemd-boot, now probably rEfind, secure boot and so on. Modules like dracut vs initcpio or openrc vs vs systemd are also there for 1 goal, but never crammed into 1 module.

kkofler:
I am not sure whether a unified design would not be better for those modules too. (Not unified with `bootloader`, of course, but unified between the different modules doing the same thing on different distros.) See also the `packages` module's design.

But our disagreement on this subject is known and unlikely to change, so I am not sure it is worth continuing this discussion here.

ghost (approved):

dalto8:
I thought I would just add that the decision to split or not split the bootloader module isn't really relevant to this PR in my opinion.  The bootloader modules is already combined, this just adds one additional option.

If it is decided to split the bootloader module later, this doesn't add any significant amount of effort to that.

Additionally, there are many other changes fixes in addition to refind support being added in this PR.

pvshvp-oss:
@demmm said: 
> Probably I am alone, but this is in my opinion complete the opposite of what should happen with the bootloader module, refind is already split in a separate module, bootloader should be split more, so I won't approve this PR. 
> https://github.com/calamares/calamares-extensions/tree/calamares/modules/refind

If Calamares' design allowed for dynamically loading modules, then splitting up everything (separate modules for each bootloader) would have made a bit more sense. But if the idea of having choices pre-made and written down appeals to you, it tells more about how you like your distribution to be installed, than anything about users' preferences.

It is definitely a significant improvement, in my opinion, to add the ability to choose bootloaders. If we try looking at counter-examples to the "split up everything" idea, the `packages` module, for example, is **not** split up into separate modules among `pacman`, `deb`, `rpm`, etc. and a package manager can be selected. I would argue this is not very different from choosing a bootloader through a variable in the globalstorage. And as @dalto8 already pointed out, the current `bootloader` module is not split up into multiple modules either. And this effort does not change that fact. The bootloader choice can just shift from the configuration file to the global storage.

ptr1337:
Tested this PR with different configurations and all bootloaders on CachyOS.
Everything is working fine.

dalto8:
Unless someone is opposed, I am going to merge this on Wednesday.

demmm:
Not clear what you are insinuating, but that `someone is opposed` has been well established in this PR.

dalto8:
>  that `someone is opposed` has been well established in this PR.

Is your opposition only because you don't think the bootloader should be combined?  If so, that isn't a change being introduced by this PR.  Calamares already has a combined bootloader.  If a decision is made later to separate it, this PR doesn't make that materially more difficult.

Having a half-combined bootloader module where some bootloaders are combined into one and others are separate makes no sense at all.  Until such time that a decision is made to separate it, any additional bootloaders should be added to this module.

demmm:
As noted in my very first comment, I am opposed to adding rEfind (module already was approved years ago to be added as a separate module, so the comment of rEfind shouldn't be a module, should have been made then).  No need to go further into that discussion, just don't say now there is no opposition known, I have no say in what doesn't get merged, so just merge and state you are merging despite knowing of some who are against.

kkofler:
I think we are not going to get unanimous consensus on this patch. So you should not claim that there is no opposition, but we can hold some kind of vote (though it is not clear who would be entitled to vote). I vote +1, merge this now. I think it is clearly an improvement over the status quo, splitting the bootloader module is a separate discussion as you point out, and I personally think the merged module makes most sense anyway (but I have been in disagreement with @demmm over that issue for years).

adriaandegroot:
There's a couple of relevant discussion points here:

## Available Modules 

- the *bootloader* module in the main Calamares repo is a bag-of-arbitrary-things module that happens to install "the" bootloader. It does not help that it has a fair bit of code for specialized situations (e.g. the random-suffix generation for EFI, that's my fault) and is weirdly structured (e.g. `vfat_correct_case`).
- the *refind* module in the Calamares-extensions repo does one thing (installing refind), carries some support code  and duplicates some code from the other module (e.g. `get_uuid()`).

## Hypotheticals

- *If* there was a way to *runtime* toggle modules, then it would be easier to work with small (single-purpose) modules: you could toggle one of a set on or off at runtime, and then include all of the chooseable-set. Use packagechooser to pick, set a globalstorage value (e.g. `module-enabled.refind:true`) .. you would probably need to extend packagechooser to allow setting multiple values to do this effectively, but it would give you a UI for picking modules at runtime "for cheap".
- *If* there was a good way to share code between Python modules, then we could dispense with "sharing implies do all the things in `main.py`". For that matter, just having a module consist of more-than-one-`.py`-file would already be an improvement. Then we could apply some software engineering to the Python modules, rather than bag-of-bits which they are now.

## Bogus Comparisons

A distro comes with a bootloader, and needs to install one (if there isn't one already or there is one that it does not understand), but doesn't derive its identity from the-bootloader-that-starts-it. Debian booted with GRUB is just as much Debian booted with rEFInd. The comparison with package managers is bogus. Even if a distro supports multiple package managers (you might claim that KDE Discover, GNOME Software Center + whatever CLI tool a distro uses means that "everybody" supports multiple package managers) there is a fair bit of identity tied up with the package manager. Debian + rpm isn't Debian, simply because Debian doesn't produce rpms to use in such a context. A package manager also does not need to deal with hardware variations (BIOS, EFI) or what-if-something-is-already-installed.

## Wishlist

I'd be **happier** with this PR if it did one thing -- lots-of-improvements, for instance, or adding rEFInd. That's basically "split the PR up" kind of work. I would be **much happier** if there was some serious cleanup done on the module. For instance, putting systemd-boot configurtion options inside a *systemd-boot* object, rather than relying on documentation in the configuration file to make clear that "these options only make sense if you pick that thing over there".

All that said, I got no skin in this game.

demmm:
>If there was a way to runtime toggle modules

This is pretty much the main reason why the refind module was added.  KaOS includes a GUI for bootloader options since 2021, one for UEFI systems & one for BIOS (through the packagechooser module, used twice for this).  And the bootloader options have been split for KaOS in separate modules ever since (so there is a refind, grub & systemd-boot module).
Reason that there has been no PR for these modules, is there has been no indication that any other distro was willing to accept the splitting of the bootloader module.

kkofler:
> _If_ there was a good way to share code between Python modules, then we could dispense with "sharing implies do all the things in `main.py`". For that matter, just having a module consist of more-than-one-`.py`-file would already be an improvement. Then we could apply some software engineering to the Python modules, rather than bag-of-bits which they are now.

But maybe we should just port `bootloader` from Python to C++ to begin with? Then it can without any changes to Calamares be nicely structured, with one file (+ one header file) per class and one class per bootloader, with, e.g., an abstract superclass for GRUB and then subclasses for BIOS GRUB, GRUB-EFI (which could also multiple-inherit another abstract superclass for EFI bootloaders), shim+GRUB-EFI (which could possibly be a subclass of GRUB-EFI).

> A package manager also does not need to deal with hardware variations (BIOS, EFI) or what-if-something-is-already-installed.

Hardware variations are actually the main reason why the `bootloader` module is unified. Because, guess what, all those fancy bootloaders we support, well, they only support (U)EFI! So both `grub` and the old `bootloader` module (which supported only BIOS GRUB and systemd-boot for EFI) ended up having an identical copy of the same BIOS GRUB installation code.

Now, there are also other BIOS bootloaders, such as LILO or syslinux, but not only has nobody had interest in adding support for Calamares so far, but doing so would only worsen the problem, because you would then end up with `n`×`m` bootloader modules handling all combinations.

dalto8:
From reviewing the above comments, it seems that more people are in favor of this that against so I will merge this.

PR #2086, opened 2023-01-13
Update license shields

wiz64:
Will change :
image
To:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 [![Current issue](https://img.shields.io/badge/issue-in_progress-FE9B48)](https://github.com/calamares/calamares/labels/hacking%3A%20in-progress)
 [![GitHub release](https://img.shields.io/github/release/calamares/calamares.svg)](https://github.com/calamares/calamares/releases)
 [![GitHub Build Status](https://img.shields.io/github/actions/workflow/status/calamares/calamares/push.yml)](https://github.com/calamares/calamares/actions?query=workflow%3Aci)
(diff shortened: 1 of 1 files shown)


dalto8:
Is that the right thing to do since different licenses applies to individual files in Calamares?

wiz64:
> ng to do since different licenses applies to individual file

took reference from :
https://en.wikipedia.org/wiki/Calamares_(software)

wiz64:
> Is that the right thing to do since different licenses applies to individual files in Calamares?

took reference from : https://en.wikipedia.org/wiki/Calamares_(software)
(and also the previous button points to GPL)

demmm:
IMO, either use this PR, or remove the license icon entirely.  As stated, icon has always linked to GPL, so might as well use it.  Having it as unspecified makes no sense at all.

wiz64:
> IMO, either use this PR, or remove the license icon entirely. As stated, icon has always linked to GPL, so might as well use it. Having it as unspecified makes no sense at all.

or we can put it to "Multiple", but GPL makes sense

demmm:
If you want to put multiple, then also adjust the link, make it go to the license directory https://github.com/calamares/calamares/tree/calamares/LICENSES

wiz64:
> If you want to put multiple, then also adjust the link, make it go to the license directory https://github.com/calamares/calamares/tree/calamares/LICENSES

Updated to Multiple
image
I would opt for GPL ... Multiple is also okay

Issue #2090, opened 2023-02-06
greetd: incorrect user and group in Debian


Sunderland93:
> Hi! Thank you for helping improve Calamares. If you are seeing a problem in installing a specific distribution, you should **probably** report the problem in the distribution's bug tracker, first. That helps filter out issues with packaging, mis-configuration, etc. that Calamares has no control over. If you are a distribution packager or maintainer, this page is for you.

**Describe the bug**
`greetd` is now in Debian main archive, but it uses different user and group unlike upstream. Upstream uses `greeter` as user and `greetd` as a group, but Debian's packages uses `_greetd` as user and `_greetd` as a group

**To Reproduce**
1. Install 

**Expected behavior**
Calamares probably should check if it is running on Debian/Ubuntu, and set user name `_greetd` and group for greetd in
https://github.com/calamares/calamares/blob/d8a03bb34b7f05f737eab26c84f0db01de90399a/src/modules/displaymanager/main.py#L763

**Screenshots and Logs**
If applicable, add screenshots to help explain your problem. Calamares has an installation log (usually `~/.cache/calamares/session.log`), please check it for confidential information and attach it if possible.

**Additional context**

demmm:
Since it is not an upstream default, but a decision by a distribution, not really a Calamares bug, more a feature request to make to greetd user & group configurable.  This is best done by those who use a distribution that created this need, thus a PR is very welcomed (so, add an option in displaymanager.conf to set the greetd user & group and use that .conf variable in the main.py).

adriaandegroot:
There might be more configuration-thingies that are distro-specific or should be customized for a distro. The PR adds a general mechanism to do so (frankly I'm not entirely happy with "greetd creates users on the target system" but it is what it is -- or it needs a nicer API in libcalamares). The schema should prevent people from shooting themselves in the foot too hard.

demmm:
With https://github.com/calamares/calamares/pull/2091 merged, can this now be closed?

Sunderland93:
Yes

PR #2091, opened 2023-02-06
Allow overriding greetd user/group

adriaandegroot:
See #2090 

This is a general mechanism: any DM-field internally may be overwritten. The schema constrains this to be sensible, but the code right now allows anything (e.g. `sddm: basic_setup: 1` will replace the setup function with an integer, which will explode at runtime).


Sunderland93:
Just tested it in my Ubuntu Sway Remix distro. Works great!

PR #2093, opened 2023-02-11
SecureBoot and OOM killer mitigation measures

phoepsilonix:
This patch aims to enable secure boot support by using mokutil, sbsigntools and the shim-signed package from AUR. It also aims to reduce the number of cases where the installer is forced to terminate by the oom killer under high load. If a swap partition exists, it enables swap.

### DIFF
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -492,6 +492,11 @@ def run():
 
         is_first = False
 
+    # swap
+    swap_enabled=subprocess.call(["sh", "-c", 'F=$(swapon --show) && [[ "$F" == "" ]]'])
+    if ( swap_enabled == 0 ):
+        subprocess.call(["sh", "-c", "SWAP=$(lsblk -l -f -n -p | awk '{if ($2==\"swap\") print $1}') && ( echo $SWAP && sudo swapon $SWAP || ( sudo mkswap $SWAP & sudo swapon $SWAP))"])
+
     repair_root_permissions(root_mount_point)
     try:
         unpackop = UnpackOperation(unpack)


src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
         efi_bootloader_id = efi_label(efi_directory)
         efi_target, efi_grub_file, efi_boot_file = get_grub_efi_parameters()
+        grub_modules = get_modules()
+        install_path = libcalamares.globalstorage.value("rootMountPoint")
 
         if is_zfs:
             check_target_env_call(["sh", "-c", "ZPOOL_VDEV_NAME_PATH=1 " + libcalamares.job.configuration["grubInstall"]
                                    + " --target=" + efi_target + " --efi-directory=" + efi_directory
+                                   + " --modules=\"" + grub_modules + "\" --sbat /usr/share/grub/sbat.csv"
+                                   + " --disable-shim-lock"
```
dalto8: Did you test these changes with zfs to ensure they are working?

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
                                    + " --bootloader-id=" + efi_bootloader_id + " --force"])
         else:
-            check_target_env_call([libcalamares.job.configuration["grubInstall"],
+            check_target_env_call(["sh", "-c", " ".join([libcalamares.job.configuration["grubInstall"],
```
dalto8: I am not sure there is a need to use `sh -c` here.  That just wraps it another layer of shell.

I think this should be avoided if possible.

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+                                   "--sbat /usr/share/grub/sbat.csv",
                                    "--bootloader-id=" + efi_bootloader_id,
-                                   "--force"])
+                                   "--force"])])
+        MOK_path = "/etc/secureboot/"
+        if os.path.exists(install_path + MOK_path + "MOK.key"):
+            grub_path = efi_directory + "/EFI/" + efi_bootloader_id + "/" + efi_grub_file
+            check_target_env_call(["sh", "-c", "/usr/bin/find /boot/ -maxdepth 1 -name 'vmlinuz-*' | /usr/bin/xargs -I{} /bin/sh -c 'if ! /usr/bin/sbverify --cert /etc/secureboot/MOK.crt {} >/dev/null 2>&1 ;then sudo /usr/bin/sbsign --key " + MOK_path + "MOK.key --cert " + MOK_path + "MOK.crt --output {} {}; fi'"])
+            check_target_env_call(["sh", "-c", "/usr/bin/sudo /usr/bin/sbsign --key " + MOK_path + "MOK.key --cert " + MOK_path + "MOK.crt --output " + grub_path + " " + grub_path])
+            check_target_env_call(["sh", "-c", "/usr/bin/sudo /usr/bin/mokutil --import " + MOK_path + "MOK.cer --root-pw"])
```
dalto8: A couple of things.

First, we should be able to avoid the `sh -c` and all this shell pipelining.  Do the work in Python, not in the shell.

Next, you don't need to use `sudo`, the application runs as root.  Also, there is no guarantee `sudo ` is installed.

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+            install_efi_directory = install_path + efi_directory
+            shutil.copy2(os.path.join(install_path, "/usr/share/shim-signed/shimx64.efi") , os.path.join(install_efi_directory, "EFI", efi_bootloader_id))
+            shutil.copy2(os.path.join(install_path, "/usr/share/shim-signed/mmx64.efi") , os.path.join(install_efi_directory, "EFI", efi_bootloader_id))
+            shutil.copy2(os.path.join(install_path, "/usr/share/shim-signed/shimx64.efi") , os.path.join(install_efi_directory, "EFI", efi_bootloader_id, "bootx64.efi"))
```
dalto8: Have you tested this?  I think `os.path.join` will discard `install_path` in this situation and so those files are being copied from the host instead of the target.

That may be problematic if the host and target are different.

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+            efi_drive = subprocess.check_output([
+            libcalamares.job.configuration["grubProbe"],
+            "-t", "drive", install_efi_directory]).decode("ascii")
+            efi_disk = subprocess.check_output([
+            libcalamares.job.configuration["grubProbe"],
+            "-t", "disk", install_efi_directory]).decode("ascii")
```
dalto8: Don't use `subprocess`, use one of the built in calamares functions as describe in the modules readme.

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+            efi_drive_partition = efi_drive.replace("(","").replace(")","").split('/')[3]
```
dalto8: You can use `strip()` instead of multiple invocations of `replace()` here.  Also since you are discarding some of the string, you can use `strip()` on the result of the split instead of the entire string.

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+            start = None
+            while c < len(efi_drive_partition):
+                if efi_drive_partition[c].isdigit() and start is None:
+                    start = c
+                if not efi_drive_partition[c].isdigit() and start is not None:
+                    efi_partition_number = efi_drive_partition[start:c]
+                    break
+                c += 1
+            if efi_partition_number is None:
+                raise ValueError("No partition number found for %s" % install_efi_directory)
```
dalto8: Don't raise an exception unless you are going to handle it somewhere.

In this case, I think it would make more sense to use `libcalamares.util.error()`

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+            efi_partition_number = None
+            c = 0
+            start = None
+            while c < len(efi_drive_partition):
+                if efi_drive_partition[c].isdigit() and start is None:
+                    start = c
+                if not efi_drive_partition[c].isdigit() and start is not None:
+                    efi_partition_number = efi_drive_partition[start:c]
+                    break
+                c += 1
```
dalto8: Use `re.search` or something similar here.

src/modules/bootloader/main.py
```diff
@@ -788,6 +842,110 @@ def prepare_bootloader(fw_type):
...
+	gcry_tiger
+	gcry_twofish
+	gcry_whirlpool
+	luks
+	lvm
+	mdraid09
+	mdraid1x
+	raid5rec
+	raid6rec
+    """
```
dalto8: Use lists instead of space separated strings.

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+                "-L", efi_bootloader_id + "Shim",
+                "-d", efi_disk,
+                "-p", efi_partition_number,
+                "-l", os.path.join(install_efi_directory, "EFI", efi_bootloader_id, install_efi_bin)])
+
+            efi_boot_next()
+
+            check_target_env_call([libcalamares.job.configuration["grubMkconfig"],
+                           "-o", os.path.join(efi_directory, "EFI",
+                                              efi_bootloader_id, "grub.cfg")])
```
dalto8: `grub-mkconfig` is already being run after this function ends.  Is this needed in addition to that?

src/modules/unpackfs/main.py
```diff
@@ -492,6 +492,11 @@ def run():
 
         is_first = False
 
+    # swap
+    swap_enabled=subprocess.call(["sh", "-c", 'F=$(swapon --show) && [[ "$F" == "" ]]'])
+    if ( swap_enabled == 0 ):
+        subprocess.call(["sh", "-c", "SWAP=$(lsblk -l -f -n -p | awk '{if ($2==\"swap\") print $1}') && ( echo $SWAP && sudo swapon $SWAP || ( sudo mkswap $SWAP & sudo swapon $SWAP))"])
+
```
dalto8, line 499: I don't think this is the right place to enable swap.

src/modules/bootloader/main.py
```diff
@@ -605,17 +605,71 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+
+            install_efi_bin = "shimx64.efi"
+            subprocess.call([
+                libcalamares.job.configuration["efiBootMgr"],
+                "-c",
+                "-w",
+                "-L", efi_bootloader_id + "Shim",
+                "-d", efi_disk,
+                "-p", efi_partition_number,
+                "-l", os.path.join(install_efi_directory, "EFI", efi_bootloader_id, install_efi_bin)])
```
dalto8: As above, don't use `subprocess`

dalto8 (changes requested):
Thanks for submitting this.  I made some specific comments but I also have a few general comments.

The code formatting here needs to be fixed.  If you aren't familiar with how to format Python, apply a style/linter to the new code.

I don't think we should modify the grub bootloader install here.  Not all distros will want this and we can't assume all this will be available on every distro.  I think if we want to add support for this it needs to be done as additional boot option.

You aren't doing any exception handling.  There are many places here where Python can throw unhandled exceptions.  You should address all of those.

kkofler:
There are several issues with this pull request:
* It mixes 2 completely unrelated changes that should be submitted in 2 separate pull requests.
* The `bootloader.py` change is very Arch-specific. Calamares is not only for Arch and Arch-based distributions! There is already an `sb-shim` option, you probably want to add a `sb-shim-mokutil` variant for the Arch setup (but I would argue that a distribution that requires a MOK to be used does in fact **not** support "Secure Boot" and that generating MOKs is not the installer's job, it is called a machine-owner key, not an installer key). In any case, you **must not** add this to the existing `grub` or `sb-shim` options, which work fine on distributions that either do not attempt to support "Secure Boot" at all (`grub` option) or that ship properly signed `shim` and `grub` binaries (`sb-shim` option). "Properly signed" means in practice that the `shim` needs to be signed by Microsoft and needs to carry the distribution's public key with which `grub` is signed (so every distribution needs to send a `shim` with its own distribution key to Microsoft for signing, blame the "Secure Boot" spec for that silliness).
* The `unpackfs.py` change does not belong into `unpackfs.py` at all (as already pointed out by @dalto8). If anything, it should be a separate module. But detecting and reusing existing swap partitions is really something that should be done (if at all – there can be reasons not to do that to begin with) in the live image itself, not in Calamares.
* There are several code quality issues with the Python code, see @dalto8's review.

phoepsilonix:
Understood. First, prioritize only the swap portion for OOM killer protection.

PR #2094, opened 2023-02-12
 I added the activation processing of the SWAP partition here because the mounting of the target device should be complete at the point of the unpackfs module

phoepsilonix:
…lso be added to automatically mount the swap partition if it exists when installing the device, including if manual partitioning is performed. This is because, especially in kernel-6.1, in processes with a large memory usage such as copying a large amount of files or generating initramfs, the OOM Killer will force the process to stop not only the installer but also the desktop environment, causing the installation to be interrupted. If there is no SWAP partition, it may also be possible to create a swapfile on the target device being installed. I think such processing should be added somewhere, but for now, I added the activation processing of the SWAP partition here because the mounting of the target device should be complete at the point of the unpackfs module

### DIFF
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -492,6 +492,11 @@ def run():
 
         is_first = False
 
+    # swap
+    swap_enabled=subprocess.call(["sh", "-c", 'F=$(swapon --show) && [[ "$F" == "" ]]'])
+    if ( swap_enabled == 0 ):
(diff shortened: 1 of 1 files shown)


dalto8 (changes requested):
As mentioned in the last PR, working with swap doesn't belong in unpackfs.  Many installations don't even use unpackfs so this is will never get called.  Even if they did, this is not the right location to activate swap.

This probably should happen in the mount module(as an option in the config) or in it's own module.

This also has all the same code quality issues as noted in detail in the prior PR.  I described them in detail over there but a quick summary would be:
* Use calamares process calls instead of `subprocess.call`.  These are described in modules readme.
* Replace `sh -c` and shell pipelining with Python logic
* Don't use `sudo`
* Ensure you have proper exception handling anywhere an exception can be thrown

PR #2096, opened 2023-02-13

phoepsilonix:
If there is no swap at all when the partition of the installation destination device is mounted, if a Swap partition exists, enable swap. target_env_call cannot be used because it is before the files are copied to the installation destination. host_env_process_output cannot be used because the return value cannot be obtained. To avoid reinventing the wheel as much as possible, use subprocess.check_output.

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -307,6 +307,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
                                         mount_option) != 0:
                 libcalamares.utils.warning("Cannot mount {}".format(device))
 
+def enable_swap_partition(swap_device):
+    try:
+        subprocess.check_call(["swapon", swap_device])
+    except subprocess.CalledProcessError:
+        raise Exception(_("Failed to swapon " + swap_device))
+
 
 def run():
     """
@@ -320,6 +326,12 @@ def run():
         libcalamares.utils.warning("partitions is empty, {!s}".format(partitions))
         return (_("Configuration Error"),
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
+
+    # swap
+    swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap']
+    swap_device="".join(map(str,swap_partitions))
+    if ( swap_device != "" ):
+        enable_swap_partition(swap_device)
 
     root_mount_point = tempfile.mkdtemp(prefix="calamares-root-")
 


src/modules/mount/main.py
```diff
@@ -149,6 +149,21 @@ def get_btrfs_subvolumes(partitions):
 
     return btrfs_subvolumes
 
+def enable_swap_partition():
+    swap_is_exist=subprocess.check_output(["swapon","-s"], universal_newlines=True)
+    if ( swap_is_exist == "" ):
+        all_partitions=subprocess.check_output(["lsblk", "-l", "-f", "-n", "-p"], universal_newlines=True).split("\n")
```
dalto8: The list of partitions is already available in `partitions` if you read above a little in the code.  That should be used instead of a call to `lsblk`.

Using this same data structure, you should also ensure the partition is "claimed".  Otherwise we could overwrite data in a swap partition the user had not selected to use.  There are examples of using this in the fstab module if you want a reference.

src/modules/mount/main.py
```diff
@@ -149,6 +149,21 @@ def get_btrfs_subvolumes(partitions):
...
+        for p in all_partitions:
+            q = p.split()
+            if ( len(q) > 1 and q[1] == "swap" ):
+                swap = q[0]
+                try:
+                    subprocess.check_call(["swapon", swap])
+                except Exception as e:
```
dalto8: This is too broad.  You should apply proper exception handling for the explicit exceptions that could be thrown.

src/modules/mount/main.py
```diff
@@ -149,6 +149,21 @@ def get_btrfs_subvolumes(partitions):
...
+                    try:
+                        subprocess.check_call(["mkswap", swap])
```
dalto8: We shouldn't blindly call `mkswap` here.  `mkswap` should have already been run by the partition module.

src/modules/mount/main.py
```diff
@@ -149,6 +149,21 @@ def get_btrfs_subvolumes(partitions):
 
     return btrfs_subvolumes
 
+def enable_swap_partition():
+    swap_is_exist=subprocess.check_output(["swapon","-s"], universal_newlines=True)
```
dalto8: Don't use `subprocess`, instead use the built-in Calamares functions as described in the modules readme.

dalto8 (changes requested):
This looks much better than the previous PR.  Just a few things we need to touch up here.

PR #2097, opened 2023-02-13

phoepsilonix:
If there is no swap at all when the partition of the installation destination device is mounted, if a Swap partition exists, enable swap.

### DIFF
--- a/src/modules/displaymanager/displaymanager.conf
+++ b/src/modules/displaymanager/displaymanager.conf
@@ -59,3 +59,11 @@ basicSetup: false
 # *displaymanagers* list (as the only one).
 #
 sysconfigSetup: false
+
+# Some DMs have specific settings. These can be customized here.
+#
+# greetd has configurable user and group; the user and group is created if it
+# does not exist, and the user is set as default-session user.
+greetd:
+  greeter_user: "tom_bombadil"
+  greeter_group: "wheel"
--- a/src/modules/displaymanager/displaymanager.schema.yaml
+++ b/src/modules/displaymanager/displaymanager.schema.yaml
@@ -20,3 +20,10 @@ properties:
         required: [ executable, desktopFile ]
     basicSetup: { type: boolean, default: false }
     sysconfigSetup: { type: boolean, default: false }
+    greetd:
+        type: object
+        properties:
+            greeter_user: { type: string }
+            greeter_group: { type: string }
+        additionalProperties: false
+
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -983,6 +983,11 @@ def run():
     # Do the actual configuration and collect messages
     dm_setup_message = []
     for dm in dm_impl:
+        dm_specific_configuration = libcalamares.job.configuration.get(dm.name, None)
+        if dm_specific_configuration and isinstance(dm_specific_configuration, dict):
+            for k, v in dm_specific_configuration.items():
+                if hasattr(dm, k):
+                    setattr(dm, k, v)
         dm_message = None
         if enable_basic_setup:
             dm_message = dm.basic_setup()


dalto8:
I will put comments on this in a few minutes.  However, please stop closing the PRs and creating new ones.  You can continue to push changes to your PR.  You don't need to create new ones each time you make changes.

src/modules/mount/main.py
```diff
@@ -307,6 +307,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
                                         mount_option) != 0:
                 libcalamares.utils.warning("Cannot mount {}".format(device))
 
+def enable_swap_partition(swap_device):
+    try:
+        subprocess.check_call(["swapon", swap_device])
```
dalto8: Don't use subprocess, instead use the built-in Calamares functions as described in the modules readme.

src/modules/mount/main.py
```diff
@@ -307,6 +307,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
...
+    except subprocess.CalledProcessError:
+        raise Exception(_("Failed to swapon " + swap_device))
```
dalto8: This exception isn't being handled.

Either create a specific exception class and handle or call `libcalamares.utils.warning()` instead of raising an exception here.  The latter would be simpler.

src/modules/mount/main.py
```diff
@@ -321,6 +327,12 @@ def run():
         return (_("Configuration Error"),
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
 
+    # swap
+    swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap']
```
dalto8: You need to ensure the swap partition is "claimed".  We shouldn't enable an unclaimed partition.

kkofler:
In particular, you can amend your commit and force-push to your branch, and the pull request will automatically pick up the amended commit.

phoepsilonix:
I understand. Could you please teach me how to make changes, as I don't know how to do it?

src/modules/mount/main.py
```diff
@@ -321,6 +327,12 @@ def run():
...
+    swap_device="".join(map(str,swap_partitions))
```
dalto8: I don't think the call to `map()` is needed but more importantly this will concat all the devices together in a string with no separator.  I don't think that is what you want here.

src/modules/mount/main.py
```diff
@@ -321,6 +327,12 @@ def run():
...
+    if ( swap_device != "" ):
+        enable_swap_partition(swap_device)
```
dalto8: Why not simply pass `swap_partitions` to this function instead and have it enable them?

src/modules/mount/main.py
```diff
@@ -307,6 +307,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
                                         mount_option) != 0:
                 libcalamares.utils.warning("Cannot mount {}".format(device))
 
+def enable_swap_partition(swap_device):
```
dalto8: You probably need to handle the condition where the swap partition is encrypted.

dalto8 (changes requested):
Getting closer now.

dalto8:
> I understand. Could you please teach me how to make changes, as I don't know how to do it?

Whenever you update your branch, the PR is automatically updated.

src/modules/mount/main.py
phoepsilonix: For example?

dalto8: `libcalamares.utils.host_env_process_output(["swapon", swap_device])`

It throws `subprocess.CalledProcessError` if there is an error.

src/modules/mount/main.py
phoepsilonix: Why do you ask?
It seems that in some cases the swap partition may not be recognized even though I manually configure the partition and expect it to be recognized as claimed false.

dalto8: The "claimed" flag should be set if we are using it with Calamares.

We can't use other swap partitions because we don't know what is in them.  For example, it might hold resume data for another distro.

src/modules/mount/main.py
```diff
@@ -321,6 +327,11 @@ def run():
         return (_("Configuration Error"),
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
 
+    # swap
+    swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap' and p['claimed'] == False ]
+    if ( swap_partitions != [] ):
```
dalto8: I don't think this `if` is needed.  If it is empty, the for loop will be skipped.

src/modules/mount/main.py
```diff
@@ -307,6 +307,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
                                         mount_option) != 0:
                 libcalamares.utils.warning("Cannot mount {}".format(device))
 
+def enable_swap_partition(swap_partitions):
```
dalto8: This should probably be renamed to be plural now that it supports more than one.

src/modules/mount/main.py
```diff
@@ -321,6 +327,11 @@ def run():
         return (_("Configuration Error"),
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
 
+    # swap
+    swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap' and p['claimed'] == False ]
```
dalto8: Shouldn't `claimed` be `True` here?

dalto8 (changes requested):

src/modules/mount/main.py
phoepsilonix: "As far as I can see from checking the log output, I believe False is the correct answer."

src/modules/mount/main.py
phoepsilonix: I agree.

src/modules/mount/main.py
phoepsilonix: It seems that in case of manual configuration it is False, in case of automatic configuration it will beTrue as well.

Issue #1911, opened 2022-03-24
FS Selection Active for "Replace Partition" option


TechXero:
**Is your feature request related to a problem? Please describe.**

"Why when selecting the replace partition in cala does not allow me to change filesystem type and swap with it greyed out stuck at default choice?"

**Describe the solution you'd like**

When user selects the "Replace Partition" option file system selection should remain active allowing user to select a different one with swap if he chooses to...

philmmanjaro:
As I see it the quick settings for file system and swap are only linked to erasing disk only. To specify more file systems you have to modify the [config file](https://github.com/calamares/calamares/blob/calamares/src/modules/partition/partition.conf#L159)

TechXero:
I have it like that 

```
efiSystemPartition:     "/boot/efi"

userSwapChoices:
    - none      # Create no swap, use no swap
    - small     # Up to 4GB
    - suspend   # At least main memory size
    - file      # To swap file instead of partition

drawNestedPartitions:   false

alwaysShowPartitionLabels: true

initialPartitioningChoice: none

initialSwapChoice: none

defaultFileSystemType:  "btrfs"

availableFileSystemTypes:  ["btrfs","xfs","ext4"]
```

Still won't show

philmmanjaro:
It will only show for erase disk. Not for replace partition! That would be a new feature.

TechXero:
Ok so on the list ? LoL

philmmanjaro:
Which list?

TechXero:
Matter of speech... Feature for the future maybe ?

adriaandegroot:
See branch *issue-1911* for initial work.

adriaandegroot:
Branch *issue-1911* was merged in december, so this one is done.

PR #2100, opened 2023-02-17
[partition] Fix issues with replace partition

dalto8:
This PR fixes some issues introduced by https://github.com/calamares/calamares/commit/61ce22338a124060652cdfde65e137214b826490:
* The partition to replace was being deselected when the filesystem was changed
* The encrypt box was being shown for zfs
* It was not using the selected filesystem

It is also removes `m_lastSelectedActionIndex` which was never being set to anything except `-1`


ptr1337:
Just tested it. It does fixes https://github.com/calamares/calamares/commit/61ce22338a124060652cdfde65e137214b826490.
The FS selection is now working correctly when using the replace option.

ptr1337:
If merged, https://github.com/calamares/calamares/issues/1911 can be closed

PR #2101, opened 2023-02-20
Enable preferred-greeters with lightdm

adriaandegroot:
This builds on @paolodongilli PR and makes things configurable, while improving the code that is looking for the greeter a little bit. Since we have support for per-DM configuration now, it's relatively simple to add this to the configuration file.

### DIFF
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -644,8 +644,22 @@ class DMlightdm(DisplayManager):
         None to indicate nothing-was-found.
         """
         greeters_dir = "usr/share/xgreeters"
-        greeter_path = os.path.join(self.root_mount_point, greeters_dir, "lightdm-greeter.desktop")
-        return greeter_path
+        greeters_target_path = os.path.join(self.root_mount_point, greeters_dir)
+        preferred_names = ("lightdm-greeter.desktop", )
+        available_names = os.listdir(greeters_target_path)
+        available_names.sort()
+        desktop_names = [n for n in preferred_names if n in available_names] # Preferred ones
+        if desktop_names:
+            return desktop_names[0]
+        desktop_names = [n for n in available_names if n.endswith(".desktop")] # .. otherwise any .desktop
+        if desktop_names:
+            return desktop_names[0]
+        desktop_names = [n for n in available_names if not n.startswith(".")] # .. otherwise any non-dot-file
+        if desktop_names:
+            return desktop_names[0]
+        if available_names: # Anything?
+            return available_names[0]
+        return None
 
 
     def greeter_setup(self):
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -659,8 +659,6 @@ class DMlightdm(DisplayManager):
         desktop_names = [n for n in available_names if not n.startswith(".")] # .. otherwise any non-dot-file
         if desktop_names:
             return desktop_names[0]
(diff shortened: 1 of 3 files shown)


adriaandegroot:
This PR takes over #2040

src/modules/displaymanager/main.py
```diff
@@ -544,6 +544,9 @@ class DMlightdm(DisplayManager):
     name = "lightdm"
     executable = "lightdm"
 
+    # Can be overridden in the .conf file
+    preferred_greeters = ["lightdm-greeter.desktop"]
```
dalto8: Since `lightdm-greeter` is debian specific, maybe `lightdm-gtk-greeter` is a more reasonable default?

src/modules/displaymanager/main.py
```diff
@@ -633,41 +636,57 @@ def desktop_environment_setup(self, default_desktop_environment):
...
+        if desktop_names:
+            return desktop_names[0]
+        desktop_names = [n for n in available_names if n.endswith(".desktop")] # .. otherwise any .desktop
+        if desktop_names:
+            return desktop_names[0]
+        desktop_names = [n for n in available_names if not n.startswith(".")] # .. otherwise any non-dot-file
+        if desktop_names:
+            return desktop_names[0]
+        if available_names: # Anything?
+            return available_names[0]
```
dalto8: Are these conditions valid?  Should we just be throwing an exception or returning an empty array here?

src/modules/displaymanager/main.py
```diff
@@ -633,41 +636,57 @@ def desktop_environment_setup(self, default_desktop_environment):
...
+            if greeter.endswith('.desktop'):
+                greeter = greeter[:-8] # Remove ".desktop" from end
+
+            libcalamares.utils.debug("found greeter {!s}".format(greeter))
+            os.system(
+                "sed -i -e \"s/^.*greeter-session=.*"
+                "/greeter-session={!s}/\" {!s}".format(
+                    greeter,
+                    lightdm_conf_path
+                )
```
dalto8, line 679: This should probably be done in python instead of using `sed` but at the very least it should be calling a libcalamares function instead of `os.system`.

dalto8 (changes requested):

src/modules/displaymanager/main.py
adriaandegroot: I've picked "empty list" as default so it doesn't have **any** preferred one, which falls back to the older behavior (except now it's deterministic by alphabet).

src/modules/displaymanager/main.py
adriaandegroot: Good point, the last one doesn't make any sense (the one on L660 does, if none of the preferred ones is matched).

src/modules/displaymanager/main.py
adriaandegroot: We don't really have a sed-like thing, do we (maybe we can link to a libsed or some suitable line-by-line replacer?). This module and the tracking module (! in C++) are the primary users of sed, so that would be a good improvement (also, eventually, towards dropping root).

But I'm not going to fix it in this PR, that's a larger issue.

src/modules/displaymanager/main.py
dalto8: Isn't the one at line 659 also invalid?  That looks like it accepts any file that doesn't start with `.`

dalto8 (approved):

PR #2102, opened 2023-02-21

phoepsilonix:
If there is no swap at all when the partition of the installation destination device is mounted,
if a Swap partition exists, enable swap.


src/modules/mount/main.py
```diff
@@ -321,6 +327,10 @@ def run():
         return (_("Configuration Error"),
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
 
+    # swap
+    swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap' ]
```
dalto8: Isn't this the same as before?

We can't use swap partitions the user hasn't chosen to use as part of this install.  It isn't safe to overwrite the data blindly.

dalto8 (changes requested):

src/modules/mount/main.py
phoepsilonix: Same content.
I requested it again as it was not taken into the branch.
Whether the user has configured the partition manually or automatically, the swap partition will not be mounted (swapon) if it does not exist, right?
If the swap does not exist, it will not be processed, so there is no problem.

phoepsilonix: It is probably missing due to MERGE.
It has been corrected and uploaded again.
Now, how about this?
```
    swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap' and p['claimed'] == False ]
    if ( swap_partitions != [] ):
        enable_swap_partition(swap_partitions)
```

dalto8: It looks like the code isn't quite committed correctly.

You don't need that `if` you added with the most recent commit but you didn't add the other.

src/modules/mount/main.py
```diff
@@ -329,7 +329,8 @@ def run():
 
     # swap
     swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap' ]
-    enable_swap_partitions(swap_partitions)
+    if ( swap_partitions != [] ):
+        enable_swap_partition(swap_partitions)
```
phoepsilonix, line 333: Now, how about this?

src/modules/mount/main.py
```diff
@@ -307,6 +307,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
                                         mount_option) != 0:
                 libcalamares.utils.warning("Cannot mount {}".format(device))
 
+def enable_swap_partitions(swap_partitions):
+    try:
+        for swap_device in swap_partitions:
+            libcalamares.utils.host_env_process_output(["swapon", swap_device])
+    except subprocess.CalledProcessError:
+        libcalamares.utils.warning("Failed to swapon " + swap_device)
```
adriaandegroot: We generally use `.format()`, or f-strings in the 3.3 series; this is *ok*, but could be nicer.

adriaandegroot:
> If there is no swap at all when the partition of the installation destination device is mounted, if a Swap partition exists, enable swap.

Why would you want to do that, though? Generally the intention is to reboot into the installed system afterwards, at which point the "new" system will take over swap, however that is configured. This just enables the live-CD-installer environment to swap onto the newly-created swap partion -- which does not seem very useful?

dalto8:
> Why would you want to do that, though? Generally the intention is to reboot into the installed system afterwards, at which point the "new" system will take over swap, however that is configured. This just enables the live-CD-installer environment to swap onto the newly-created swap partion -- which does not seem very useful?

The context has been lost across the many times the PR has been recreated, here is the original rationale:

> It also aims to reduce the number of cases where the installer is forced to terminate by the oom killer under high load. If a swap partition exists, it enables swap.

src/modules/mount/main.py
dalto8: That change does nothing.

`enable_swap_partitions` does nothing when `swap_partitions` is empty so checking if it is empty doesn't matter.

The line that needs to be changed is this one:
```
swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap' ]
```

It needs to have the claimed check added back to it.

phoepsilonix (commented):
Added automatic enabling of swap partition before installation starts.
Mainly to suppress installer stoppages caused by oom killer.

src/modules/mount/main.py
dalto8: In case you aren't famliair with f strings, it would look like this:
```
libcalamares.utils.warning(f"Failed to swapon {swap_device}")
```

phoepsilonix (commented):
Added automatic enabling of swap partition before installation starts.
Mainly to suppress installer stoppages caused by oom killer.

phoepsilonix (commented):
Corrected by.

The line that needs to be changed is this one:
```
swap_partitions = [p['device'] for p in partitions if p['fs'] == 'linuxswap' ]
```
It needs to have the claimed check added back to it.

src/modules/mount/main.py
```diff
@@ -307,6 +307,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
                                         mount_option) != 0:
                 libcalamares.utils.warning("Cannot mount {}".format(device))
 
+def enable_swap_partitions(swap_partitions):
+    try:
+        for swap_device in swap_partitions:
+            libcalamares.utils.host_env_process_output(["swapon", swap_device])
+    except subprocess.CalledProcessError:
+        libcalamares.utils.warning(f"Failed to swapon " + swap_device)
```
dalto8: You can't just put an `f` in front of it.  You need to change it the way I showed you above.

src/modules/mount/main.py
dalto8: This is still missing the check for claimed.

dalto8 (changes requested):

src/modules/mount/main.py
phoepsilonix: OK

src/modules/mount/main.py
phoepsilonix: The presence or absence of CLAIMED should not be a concern.
This is because whether a swap partition is present or not, whether it is configured automatically or manually, is a case-by-case basis.

phoepsilonix: Whether in automatic or manual configuration, if a swap partition exists, it can be enabled to deter the occurrence of OOM killers on low-spec machines.

dalto8: You have to do *something* to ensure that you are not enabling a swap partition that hasn't been identified for use in the installer. 

You can't just enable any swap partition because it may have data that shouldn't be overwritten such as resume data for another distro.

phoepsilonix: Of course, if we were going to get serious, swapfile, or even zram, would be a consideration, as well as adding another command to exclude the oom killer.
```
pidof -xw Xwayland calamares calamares_polkit | xargs -n1 sudo choom -n -1000 -p
```
But for now, just enabling the swap partition will reduce the likelihood of the installer being forced to quit in the middle of the install on many low-spec machines.

dalto8: Again, the issue is that you can't enable a swap partition because it exists.  That can cause breakage.

You can only enable a swap partition if it has been selected to be used with the install.

phoepsilonix (commented):
```
swap_partitions = [p['device'] for p in partitions if ( p['fs'] == 'linuxswap' and p.get("claimed", None) )]
```
The test needs time to be tested as it is reflected in the package.

src/modules/mount/main.py
phoepsilonix:
```
swap_partitions = [p['device'] for p in partitions if ( p['fs'] == 'linuxswap' and p.get("claimed", None) )]
```

dalto8:
Have you tested this in it's current form?

src/modules/mount/main.py
kkofler: And if you are going to check for `claimed`, you need to check that it is `true`, as you suggested here, not that it is `false`, as I have seen in several previous iterations of your patch.

dalto8:
Uhh...I don't know what you just did but it is completely broken now.  It likes the file is partially merged with a different version of itself.

You are making this harder than it needs to be from a git perspective.  You should probably stop doing merges and force pushes.

phoepsilonix:
Sure enough, things were getting a little crazy, so I adjusted the merge.
Please review the final file changes.
If possible, it would be nice if claimed would be true for manual partitioning configurations, but I don't know how to do that, so it is incomplete, but perhaps the patch will automatically enable SWAP partitions for automatic configurations.

src/modules/mount/main.py
```diff
@@ -307,6 +307,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
                                         mount_option) != 0:
                 libcalamares.utils.warning("Cannot mount {}".format(device))
 
+def enable_swap_partition(swap_device):
+    try:
+        for d in swap_device:
+        	subprocess.check_call(["swapon", d])
+    except subprocess.CalledProcessError:
```
dalto8: It looks like we reverted back to subprocess at some point in the many merges.  This should use `host_env_process_output()` instead.

src/modules/mount/main.py
```diff
@@ -321,6 +327,10 @@ def run():
         return (_("Configuration Error"),
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
 
+    swap_partitions = [p['device'] for p in partitions if ( p['fs'] == 'linuxswap' and p['claimed'] == True ) ]
```
dalto8: This should be:
```
swap_partitions = [p['device'] for p in partitions if ( p['fs'] == 'linuxswap' and p['claimed'] ) ]
```

There is no point in comparing it to `True`.

dalto8 (changes requested):

src/modules/mount/main.py
phoepsilonix: OK?

src/modules/mount/main.py
phoepsilonix: OK

phoepsilonix:
Are there any other issues?

dalto8:
It looks like you resolved the most important issues.

Have you tested it the way the code is now?

phoepsilonix:
Yes, I did. Done.

PR #2103, opened 2023-02-22
[partition] Add support for zfs encryption

dalto8:
This PR adds support for zfs encryption.

### DIFF
--- a/src/modules/partition/gui/EncryptWidget.cpp
+++ b/src/modules/partition/gui/EncryptWidget.cpp
@@ -16,6 +16,9 @@
 #include "Branding.h"
 #include "utils/CalamaresUtilsGui.h"
 #include "utils/Retranslator.h"
+#include <kpmcore/fs/filesystem.h>
+
+constexpr uint ZFS_MIN_LENGTH = 8;
 
 /** @brief Does this system support whole-disk encryption?
  *
@@ -150,10 +153,16 @@ EncryptWidget::updateState()
         QString p1 = m_ui->m_passphraseLineEdit->text();
         QString p2 = m_ui->m_confirmLineEdit->text();
 
+
         if ( p1.isEmpty() && p2.isEmpty() )
         {
             applyPixmap( m_ui->m_iconLabel, CalamaresUtils::StatusWarning );
             m_ui->m_iconLabel->setToolTip( tr( "Please enter the same passphrase in both boxes." ) );
+        }
+        else if ( FileSystem::typeForName( m_filesystem ) == FileSystem::Zfs && p1.length() < ZFS_MIN_LENGTH )
+        {
+            applyPixmap( m_ui->m_iconLabel, CalamaresUtils::StatusError );
+            m_ui->m_iconLabel->setToolTip( tr( "Password must be a minimum of %1 characters" ).arg( ZFS_MIN_LENGTH ) );
         }
         else if ( p1 == p2 )
         {
@@ -199,5 +208,12 @@ EncryptWidget::onCheckBoxStateChanged( int checked )
     m_ui->m_confirmLineEdit->clear();
     m_ui->m_iconLabel->clear();
 
+    updateState();
+}
+
+void
+EncryptWidget::setFilesystem( const QString& fs )
+{
+    m_filesystem = fs;
     updateState();
 }
--- a/src/modules/partition/gui/EncryptWidget.cpp
+++ b/src/modules/partition/gui/EncryptWidget.cpp
@@ -16,7 +16,6 @@
 #include "Branding.h"
 #include "utils/CalamaresUtilsGui.h"
 #include "utils/Retranslator.h"
-#include <kpmcore/fs/filesystem.h>
 
 constexpr int ZFS_MIN_LENGTH = 8;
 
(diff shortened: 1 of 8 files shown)


src/modules/partition/gui/EncryptWidget.cpp
```diff
@@ -143,18 +146,24 @@ applyPixmap( QLabel* label, CalamaresUtils::ImageType pixmap )
...
 void
-EncryptWidget::updateState()
+EncryptWidget::updateState( const bool notify )
 {
     if ( m_ui->m_passphraseLineEdit->isVisible() )
     {
         QString p1 = m_ui->m_passphraseLineEdit->text();
         QString p2 = m_ui->m_confirmLineEdit->text();
 
+
```
adriaandegroot: Stray blank?

ptr1337:
Reviewed and tested with CachyOS.
Everything works. Actually Im not sure, if we can add a restriction, that if grub or refind is selected that the encryption box won't show.
But this can be also solved from the maintainer config.

Everything good here :) Thank you for the PR

src/modules/partition/gui/EncryptWidget.cpp
```diff
@@ -143,18 +146,24 @@ applyPixmap( QLabel* label, CalamaresUtils::ImageType pixmap )
 }
 
 void
-EncryptWidget::updateState()
+EncryptWidget::updateState( const bool notify )
```
adriaandegroot, line 149: `const` value parameters always seem a bit sketchy to me: they're copies of something and their constness doesn't really matter (unless you're concerned that `notify` might be changed inside the function and the logic gets confused).

dalto8:
> Everything works. Actually Im not sure, if we can add a restriction, that if grub or refind is selected that the encryption box won't show.

That may not work if the bootloader is being selected dynamically.  

How about a config option for the partition module that determines if zfs encryption should be enabled?

src/modules/partition/gui/EncryptWidget.cpp
```diff
@@ -143,18 +146,24 @@ applyPixmap( QLabel* label, CalamaresUtils::ImageType pixmap )
...
+
         if ( p1.isEmpty() && p2.isEmpty() )
         {
             applyPixmap( m_ui->m_iconLabel, CalamaresUtils::StatusWarning );
             m_ui->m_iconLabel->setToolTip( tr( "Please enter the same passphrase in both boxes." ) );
         }
+        else if ( FileSystem::typeForName( m_filesystem ) == FileSystem::Zfs && p1.length() < ZFS_MIN_LENGTH )
+        {
+            applyPixmap( m_ui->m_iconLabel, CalamaresUtils::StatusError );
+            m_ui->m_iconLabel->setToolTip( tr( "Password must be a minimum of %1 characters" ).arg( ZFS_MIN_LENGTH ) );
```
adriaandegroot, line 164: is that a ZFS-specific requirement that always needs to be enforced? Otherwise, why not plug in existing password-checking mechanisms? (Er .. probably because they, klike libpwquality, are in the *user* module and not here).

src/modules/partition/gui/EncryptWidget.cpp
```diff
@@ -201,3 +213,13 @@ EncryptWidget::onCheckBoxStateChanged( int checked )
...
     updateState();
 }
+
+void
+EncryptWidget::setFilesystem( const QString& fs )
+{
+    m_filesystem = fs;
+    if ( m_state != Encryption::Disabled )
+    {
+        updateState( false );
```
adriaandegroot, line 222: I don't get why you would want to avoid an update-notification here?

src/modules/partition/gui/EncryptWidget.h
```diff
@@ -38,6 +38,12 @@ class EncryptWidget : public QWidget
     Encryption state() const;
     void setText( const QString& text );
 
+    /**
+     * @brief setFilesystem sets the filesystem name used for password validation
+     * @param fs A QString containing the name of the filesystem
+     */
+    void setFilesystem( const QString& fs );
```
adriaandegroot: This should be `setFilesystemName` then, since it's .. well, maybe `...Label` except the Label is something else again (?).  It's definitely not the Filesystem though, then I'd expect to use an enum from KPMCore.

adriaandegroot (changes requested):

ptr1337:
> > Everything works. Actually Im not sure, if we can add a restriction, that if grub or refind is selected that the encryption box won't show.
> 
> That may not work if the bootloader is being selected dynamically.
> 
> How about a config option for the partition module that determines if zfs encryption should be enabled?

That would be a idea. Or a note, that grub is not supported (not sure about refind with the extra driver).

src/modules/partition/gui/EncryptWidget.cpp
dalto8: Yes, it is a minimum for zfs specifically.

src/modules/partition/gui/EncryptWidget.h
dalto8: I am converting it to an enum later.  It is probably better to change it here so it expects an enum.

src/modules/partition/gui/EncryptWidget.cpp
dalto8: Changes to the encryptwidget trigger a call to `ChoicePage::onActionChanged`.  That in turn calls `EncryptWidget::setFilesystem` which was then triggering a signal storm.

src/modules/partition/gui/EncryptWidget.cpp
dalto8: That was my general reasoning, nothing should be changing it inside the function.

dalto8:
I added a config option `allowZfsEncryption` that makes it possible to disable zfs encryption if needed when support isn't available in the distro/bootloader.

PR #2106, opened 2023-02-28
greetd: Add custom CSS file path to gtkgreet

Sunderland93:
By default `gtkgreet` uses Adwaita GTK theme. This option adds the ability to apply a custom style to `gtkgreet`. If the `css` file is not present in this directory, `gtkgreet` will simply issue a warning and start with the Adwaita style. I think this is not the ideal solution, so I'll be glad to know if there is a better approach.

### DIFF
--- a/src/modules/displaymanager/displaymanager.conf
+++ b/src/modules/displaymanager/displaymanager.conf
@@ -65,11 +65,14 @@ sysconfigSetup: false
 # greetd has configurable user and group; the user and group is created if it
 # does not exist, and the user is set as default-session user.
 #
+# Some greeters for greetd (e.g gtkgreet or regreet) have support for a user's GTK CSS style to change appearance.
+#
 # lightdm has a list of greeters to look for, preferring them in order if
 # they are installed (if not, picks the alphabetically first greeter that is installed).
 #
 greetd:
   greeter_user: "tom_bombadil"
   greeter_group: "wheel"
+  greeter_css_location: "/etc/greetd/style.css"
 lightdm:
   preferred_greeters: ["lightdm-greeter.desktop", "slick-greeter.desktop"]
--- a/src/modules/displaymanager/displaymanager.schema.yaml
+++ b/src/modules/displaymanager/displaymanager.schema.yaml
@@ -25,6 +25,7 @@ properties:
         properties:
             greeter_user: { type: string }
             greeter_group: { type: string }
+            greeter_css_location: { type: string }
         additionalProperties: false
     lightdm:
         type: object
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -782,6 +782,7 @@ class DMgreetd(DisplayManager):
     executable = "greetd"
     greeter_user = "greeter"
     greeter_group = "greetd"
+    greeter_css_location = None
     config_data = {}
 
     def os_path(self, path):
@@ -849,6 +850,8 @@ class DMgreetd(DisplayManager):
         de_command = default_desktop_environment.executable
         if os.path.exists(self.os_path("usr/bin/gtkgreet")) and os.path.exists(self.os_path("usr/bin/cage")):
             self.config_data['default_session']['command'] = "cage -d -s -- gtkgreet"
+            if self.greeter_css_location:
+                self.config_data['default_session']['command'] += f" -s {self.greeter_css_location}"
         elif os.path.exists(self.os_path("usr/bin/tuigreet")):
             tuigreet_base_cmd = "tuigreet --remember --time --issue --asterisks --cmd "
             self.config_data['default_session']['command'] = tuigreet_base_cmd + de_command


Sunderland93:
See https://git.sr.ht/~kennylevinsen/gtkgreet/commit/90424c865f3b8509ec684a987c76599e697fbb83

dalto8:
What about adding a key to the conf file called `greeter_css_location` which contains the path to the CSS?

If it is omitted, don't use `-c`.  If it is present, use `-c` with the path from the config.

Sunderland93:
@dalto8 ok, good idea. I'll try

Sunderland93:
@dalto8 can you review changes? I'm new to the codebase of Calamares

src/modules/displaymanager/main.py
```diff
@@ -782,6 +782,7 @@ class DMgreetd(DisplayManager):
     executable = "greetd"
     greeter_user = "greeter"
     greeter_group = "greetd"
+    greeter_css_location = "etc/greetd/style.css"
```
dalto8: Is this path and filename an official default for gtreetd?  If not, this should probably be set to `None` instead of being hard set to a path.

If it is an official path, it probably needs a leading `/` based on how you are using it.

src/modules/displaymanager/displaymanager.conf
```diff
@@ -71,5 +71,6 @@ sysconfigSetup: false
 greetd:
   greeter_user: "tom_bombadil"
   greeter_group: "wheel"
+  greeter_css_location: "/etc/greetd/style.css"
```
dalto8, line 76: There should be some comments above that describe what this is and how to use it.

src/modules/displaymanager/main.py
```diff
@@ -847,7 +848,9 @@ def set_autologin(self, username, do_autologin, default_desktop_environment):
         self.config_load()
 
         de_command = default_desktop_environment.executable
-        if os.path.exists(self.os_path("usr/bin/gtkgreet")) and os.path.exists(self.os_path("usr/bin/cage")):
+        if os.path.exists(self.os_path("usr/bin/gtkgreet")) and os.path.exists(self.os_path("usr/bin/cage")) and os.path.exist(self.os_path(self.greeter_css_location)):
+            self.config_data['default_session']['command'] = "cage -d -s -- gtkgreet -s" + self.greeter_css_location
+        elif os.path.exists(self.os_path("usr/bin/gtkgreet")) and os.path.exists(self.os_path("usr/bin/cage")):
```
dalto8: This could be refactored to avoid the logic duplication.

dalto8 (changes requested):

src/modules/displaymanager/main.py
Sunderland93: There is no strict recommendation on where to place this file, however, everything related to greetd is located in /etс/greetd, it is customary to place custom configuration files and so on there.

dalto8: In that case, set it to `None`.  I don't think we should create an arbitrary default.

If they don't set it in the config file, we can not use it.

src/modules/displaymanager/main.py
Sunderland93: Please check new logic

src/modules/displaymanager/main.py
```diff
@@ -849,6 +850,8 @@ def set_autologin(self, username, do_autologin, default_desktop_environment):
         de_command = default_desktop_environment.executable
         if os.path.exists(self.os_path("usr/bin/gtkgreet")) and os.path.exists(self.os_path("usr/bin/cage")):
             self.config_data['default_session']['command'] = "cage -d -s -- gtkgreet"
+            if self.greeter_css_location is not None:
+               self.config_data['default_session']['command'] + " -s " + self.greeter_css_location
```
dalto8: I have a couple of questions on this line:
* Shouldn't you be assigning it to something?
* Isn't it `-c`?

kkofler:
Why should Calamares set this? Should this not be already set on the live image?

dalto8:
> Why should Calamares set this? Should this not be already set on the live image?

Not everyone installs by unpacking an image.  Also, not everyone uses the same DM on the live image as on the installed system.

src/modules/displaymanager/main.py
Sunderland93: @dalto8 
1) What do you mean?
2) No. According to `gtkgreet --help:`

`-s, --style              CSS style to use`

src/modules/displaymanager/main.py
```diff
@@ -848,7 +849,10 @@ def set_autologin(self, username, do_autologin, default_desktop_environment):
 
         de_command = default_desktop_environment.executable
         if os.path.exists(self.os_path("usr/bin/gtkgreet")) and os.path.exists(self.os_path("usr/bin/cage")):
-            self.config_data['default_session']['command'] = "cage -d -s -- gtkgreet"
+            if self.greeter_css_location is not None:
+                self.config_data['default_session']['command'] = "cage -d -s -- gtkgreet -s " + self.greeter_css_location
+            else:
+               self.config_data['default_session']['command'] = "cage -d -s -- gtkgreet"
```
dalto8: We went back to duplication again.

How about something like this:
```
self.config_data['default_session']['command'] = "cage -d -s -- gtkgreet -s"
if self.greeter_css_location:
    self.config_data['default_session']['command'] += " -s {self.greeter_css_location}"
```

src/modules/displaymanager/main.py
```diff
@@ -849,6 +850,8 @@ def set_autologin(self, username, do_autologin, default_desktop_environment):
         de_command = default_desktop_environment.executable
         if os.path.exists(self.os_path("usr/bin/gtkgreet")) and os.path.exists(self.os_path("usr/bin/cage")):
             self.config_data['default_session']['command'] = "cage -d -s -- gtkgreet"
+            if self.greeter_css_location:
+                self.config_data['default_session']['command'] += " -s {self.greeter_css_location}"
```
dalto8: Oops, there is a typo I here.  I missed the `f`.  It should be:
```
                self.config_data['default_session']['command'] += f" -s {self.greeter_css_location}"
```

Sunderland93: Done

dalto8 (approved):

kkofler:
> Not everyone installs by unpacking an image.

Then they install from what? Packages? Then the package should ship the desired default configuration.

> Also, not everyone uses the same DM on the live image as on the installed system.

That does not preclude shipping on the live image a working configuration for the DM that will be installed.

I really do not see why this kind of setup belongs in Calamares. And where do we draw the line? There are all sorts of distro-specific configuration tweaks that could theoretically be done in Calamares. But I do not see the advantage of adding a Calamares configuration setting where that can be tweaked over just tweaking the original configuration file to begin with.

dalto8:
> Then the package should ship the desired default configuration.

In my experience, this is rarely the case.  Especially since Calamares is often used by derivative distros that don't control the packaging.



> And where do we draw the line? There are all sorts of distro-specific configuration tweaks that could theoretically be done in Calamares.

A fair question I suppose but we already have lots of similar configuration options already for other things.

Also, since it is optional, it seems like there is little harm in adding it.

Sunderland93:
>That does not preclude shipping on the live image a working configuration for the DM that will be installed.

User can put the configuration file for `greetd` in LiveCD, yes. But after installing, Calamares will override it, based on options defined in `displaymanager.conf`. So if in LiveCD `config.toml` has a `cage -d -s -- gtkgreet -s /etc/greetd/style.css`, after installing it will changed to `cage -d -s -- gtkgreet`, since there is no option to define path to `style.css` file.

kkofler:
Then it is already doing too much and the logic should be removed, not added to.

Calamares should only modify those settings that it needs to modify (e.g., the name of the user to autologin if autologin was chosen) and leave all others untouched. The code for other display managers such as sddm already works that way. The greetd code changes way too much stuff.

Sunderland93:
>The code for other display managers such as sddm already works that way. The greetd code changes way too much stuff.

`greetd` has several greeters with different settings. `gtkgreet` is official (created by the same developer as a greetd, "The flagship graphical, GTK based greeter" according to the web site). What's wrong with adding a simple extra option?

kkofler:
Choosing what greeter to use is something that should be already preconfigured on the image. Calamares should only change the settings it needs to change (basically only the autologin settings, i.e., yes/no and, if yes, the username and the session type (desktop environment) to use) and keep the others as is. The method is called `set_autologin`, not `set_greeter`.

Compare, e.g., with the `set_autologin` method for SDDM:
https://github.com/calamares/calamares/blob/1c20d6d5080a62a7aea4e519d235a195d655a917/src/modules/displaymanager/main.py#L738

Sunderland93:
Calamares will override any pre-defined options, because it fully rewrite `config.toml`

https://github.com/calamares/calamares/blob/1c20d6d5080a62a7aea4e519d235a195d655a917/src/modules/displaymanager/main.py#L817

That's why there is no option to define `config.toml` in LiveCD and keep it after system will installed.

dalto8:
> Choosing what greeter to use is something that should be already preconfigured on the image. Calamares should only change the settings it needs to change

I continue to have the same two objections to this above:
1. The image configuration and the installed system configuration aren't always related as in the case of a net install.  Calamares is used in many different use cases.
2. There are other modules which allow configuration similar to this.  Even this module allows choosing a greeter for other DMs. There are many places where Calamares supports adding additional configuration to make it easier for the distros using it.

The idea that Calamares should only do the absolute minimum number of things it can do to create an installed system is not consistent with what is currently in Calamares.  My opinion is that there is no reason to refuse contributions which make Calamares easier to use.

kkofler:
> Calamares will override any pre-defined options, because it fully rewrite `config.toml`

And that is exactly what needs to be fixed. It is a bug. As I said, the code for the other display managers deliberately does not work that way.

kkofler:
> 2\. Even this module allows choosing a greeter for other DMs.

For **one** other display manager, lightdm, where this was recently added, and IMHO should not have been. Instead, there too, Calamares should preserve, not overwrite, the setting that was already there. There is no reason to touch it at all.

Sunderland93:
Tested today in my Ubuntu Sway Remix distro. Works as expected!

src/modules/displaymanager/displaymanager.conf
```diff
@@ -65,11 +65,15 @@ sysconfigSetup: false
 # greetd has configurable user and group; the user and group is created if it
 # does not exist, and the user is set as default-session user.
 #
+# some greeters for greetd (e.g gtkgreet or regreet) supports for user's GTK CSS style for changing appearance.
```
adriaandegroot:
```suggestion
# Some greeters for greetd (e.g gtkgreet or regreet) have support for a user's GTK CSS style to change appearance.
```

src/modules/displaymanager/displaymanager.schema.yaml
```diff
@@ -25,6 +25,7 @@ properties:
         properties:
             greeter_user: { type: string }
             greeter_group: { type: string }
+            greeter_css_location: {type: string}
```
adriaandegroot: Please stick to the formatting of the other schema lines (e.g. more spaces -- Calamares is a very spacey codebase)

adriaandegroot:
There's a combinatorial explosion going on *somewhere*, though I don't know exactly where:
- there's a display manager (well, one gets installed, but there's at least a half-dozen of them)
- a display manager may have one greeter, or configurable greeters
- (possibly) a greeter may be used by more than one DM

Ideally, it would be the packaging of the greeter that does-whatever-is-needed to use that greeter in the DM it's being installed for -- that means packaging smarts. Lacking those smarts, you need some other mechanism. If a distro picks a specific DM + greeter for installation, then a specific config file ought to work -- in that case the DM module should **not** edit the config file beyond the minimum. I suspect that the problem comes up when the greeter is selectable during the installation, or the config file comes from packaging and doesn't match what the distro actually wants (in which case, why is the distro using that package and not re-packaging?).

Returning to @Sunderland93 's original problem: if the issue is that the DM module is overwriting the specific config-key for the greeter (because of earlier feature requests to make it detect which greeter is installed), how about a config key `dont_touch_the_greeter_command`? Then you could pre-configure the config file with the way the distro wants it, and avoid the (mis)feature where Calamares messes with the command.

Sunderland93:
>how about a config key dont_touch_the_greeter_command?

This would be great as it would allow any `greetd` greeter's to be used with any compositor to run them (not just `cage`, but `sway` for example), leaving Calamares to just enable autologin functionality. However, I don't know how to implement something like this properly.

kkofler:
> how about a config key `dont_touch_the_greeter_command`?

IMHO, the code to touch the greeter command should just be removed. It is an incompatible change, but there have been others like this.

dalto8:
The idea that Calamares shouldn't support setting configuration seems to ignore two points.

The first is that Calamares already supports this in many others modules.  It isn't only this module that allows configuration.

Perhaps the larger reason is that how distros are using Calamares is evolving.  There are a growing number of distros that are not installing an image, but instead using Calamares as a tool to give the user an increasing amount of control over the installation.  This drives changes in not only the selection of packages that are installed but the config of certain elements of the install as well.  This may not have been the original intention of Calamares but if you look at the changes in the code base over the last few years, there are many enhancements to support this use case.

Sunderland93:
So, can this be merged?

PR #2107, opened 2023-03-02
greetd: Add ReGreet greeter

Sunderland93:
https://github.com/rharish101/ReGreet

### DIFF
--- a/src/modules/displaymanager/main.py
+++ b/src/modules/displaymanager/main.py
@@ -854,6 +854,8 @@ class DMgreetd(DisplayManager):
             self.config_data['default_session']['command'] = tuigreet_base_cmd + de_command
         elif os.path.exists(self.os_path("usr/bin/ddlm")):
             self.config_data['default_session']['command'] = "ddlm --target " + de_command
+        elif os.path.exists(self.os_path("usr/bin/regreet")) and os.path.exists(self.os_path("usr/bin/cage")):
+            self.config_data['default_session']['command'] = "cage -d -s -- regreet"
         else:
             self.config_data['default_session']['command'] = "agreety --cmd " + de_command
 


adriaandegroot:
I think most of the discussion in the other greetd-related PR applies here, too: this is turning the DM module into a greetd-configuration-thing, which doesn't seem like the right place to be editing greetd's config file.

adriaandegroot:
.. ok, I see there's an `else:` block that forces *agreetty*, and that the desktop-environment command needs to be passed to the greeter, so maybe it does need more configuration from the Calamares side than expected, since substituting in the desktop-environment command needs you to know where it goes.

Maybe doing variable expansion (e.g. write `${CALAMARES_DESKTOP_ENVIRONMENT}` in the config file and let Calamares do substitution on that, rather that re-writing the desktop command entirely) would help?

kkofler:
> I think most of the discussion in the other greetd-related PR applies here, too: this is turning the DM module into a greetd-configuration-thing, which doesn't seem like the right place to be editing greetd's config file.

Indeed.

Sunderland93:
@adriaandegroot I think that due to the ever-increasing number of greeters for `greetd`, only the user and group setup for `greetd` should be left in Calamares, as well as the autologin and startup command in `/etc/greetd/environments`. The choice of the required greeter as well as the compositor for its launch (`cage`, `sway`, etc) is preferably left as the user configured it by default in LiveCD

Sunderland93:
>Maybe doing variable expansion (e.g. write ${CALAMARES_DESKTOP_ENVIRONMENT} in the config file and let Calamares do substitution on that, rather that re-writing the desktop command entirely) would help?

You mean option in displaymanager.conf like "`CALAMARES_DESKTOP_ENVIRONMENT: "cage -s -d -- gtkgreet"`"?

adriaandegroot:
@Sunderland93 yes, like that. We could add a key to `displaymanager.conf`, maybe call it *vars* which define arbitrary text substitutions to do on the config-file. Then you need to deliver the *greetd* configuration file with those `${variables}` in it, which would get substituted.

adriaandegroot:
.. but it would take more code to do that, for sure.

adriaandegroot:
Closing this one as "needs some re-architecting to deal with the increasing amount of stuff that ends up being greetd-specific".

PR #2115, opened 2023-03-22
cmake: Change Release build from -O4 to -O3

ptr1337:


adriaandegroot:
Although "this flag has been like this since 2015" isn't an amazingly good reason to keep it set that way, and `-O4` is documented to be the same as `-O3` (for clang; gcc does not mention it at all), what is the reason you want to change it?

ptr1337:
Yes, -O4 would be the same as -O3 at all.
Since -O4 was deprecated years ago and I don't think there should be such values in it.

If you don't see any sense in it, just close the PR at all

kkofler:
For GCC, `-O4` is also the same as `-O3`.

(For the record: Unless you are talking about the TIGCC cross-compiler for TI calculators, where I reduced the inlining parameters for `-O3` and kept `-O4` and higher as is, but since TIGCC cannot compile Calamares anyway, and it would not really make sense on the target to begin with, that detail does not matter here.)

PR #2128, opened 2023-04-19
Default to luks2 when creating encrypted partitions

### LINKED ISSUE #2127: Deprecate luks1 and use luks2 with argon2id
Context: https://twitter.com/hipsterelectron/status/1648167820495142915, https://mjg59.dreamwidth.org/66429.html

I'm proposing that calamares defaults to luks2 and effectively deprecate luks1.

So switch https://github.com/calamares/calamares/blob/4f2ab856a320c840ec3fc492faf1f3de32769398/src/modules/partition/gui/CreatePartitionDialog.cpp#L260 to luks2

Even though this PR isn't ready: https://invent.kde.org/system/kpmcore/-/merge_requests/43, I think we can still switch the defaults to be luks2 and hope that `cryptsetup` was compiled with correct defaults and it picks argon2id.

aaomidi:
Partially resolves: https://github.com/calamares/calamares/issues/2127

### DIFF
--- a/src/modules/partition/gui/CreatePartitionDialog.cpp
+++ b/src/modules/partition/gui/CreatePartitionDialog.cpp
@@ -257,7 +257,7 @@ CreatePartitionDialog::getNewlyCreatedPartition()
                                                              fsLabel,
                                                              first,
                                                              last,
-                                                             Config::luksGenerationNames().find(luksFsType, Config::LuksGeneration::Luks1),
+                                                             Config::luksGenerationNames().find(luksFsType, Config::LuksGeneration::Luks2),
                                                              luksPassphrase,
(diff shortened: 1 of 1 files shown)


ptr1337:
Hi,

I don't  think we should default to luks2 right now.
The big problem here is that some bootloaders are not supporting it right now out of the box (e.G grub).

aaomidi:
I think I'll keep the discussion in the #2127, just so the thread doesn't get lost!

dalto8:
I agree that changing to luks2 as a global default isn't a viable option without broadly available grub support.

However, we already have support for distros to set luks2 as the default if the distro supports luks2 so I don't think there is anything to be done here right now.

PR #2136, opened 2023-05-08
fix non ascii keyboard

### LINKED ISSUE #2135: non-ascii-layouts not write to default/keyboard
Hello

**Bug**
When install system with non-ascii keyboards (ru, ua, gr and other) write incorrect configuration in `/etc/default/keyboard` and after reboot we get os without english keyboard on login screen

current after installation
```
XKBMODEL="pc105"
XKBLAYOUT="ru"
XKBVARIANT=""
XKBOPTIONS=""
```

**Expected behavior**
correct `/etc/default/keyboard`
```
XKBMODEL="pc105"
XKBLAYOUT="us,ru"
XKBVARIANT=","
XKBOPTIONS="grp:caps_toggle,grp_led:scroll"
```

In git i see https://github.com/calamares/calamares/blob/calamares/src/modules/keyboard/SetKeyboardLayoutJob.cpp#L269
you have correct code for `xorg.conf.d` file, but need add same for `/etc/default/keyboard` around https://github.com/calamares/calamares/blob/calamares/src/modules/keyboard/SetKeyboardLayoutJob.cpp#L321

ivan1986:
fix #2135
checked

### DIFF
--- a/src/modules/keyboard/SetKeyboardLayoutJob.cpp
+++ b/src/modules/keyboard/SetKeyboardLayoutJob.cpp
@@ -175,6 +175,13 @@ SetKeyboardLayoutJob::findLegacyKeymap() const
     return ::findLegacyKeymap( m_layout, m_model, m_variant );
 }
 
+QStringList
+SetKeyboardLayoutJob::removeEmpty(QStringList& list) const
+{
+    list.removeAll(QString());
+    return list;
+}
+
 
 bool
 SetKeyboardLayoutJob::writeVConsoleData( const QString& vconsoleConfPath, const QString& convertedKeymapPath ) const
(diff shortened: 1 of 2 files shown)


dalto8:
Can we modularize that logic instead of duplicating it in a multiple functions?

ivan1986:
format of xorg.conf.d and keybord files is different, i think write universal function give much more code, it's just 5 if's but if you find better solution, it's just one case how to fix this issue

src/modules/keyboard/SetKeyboardLayoutJob.cpp
```diff
@@ -324,9 +324,36 @@ SetKeyboardLayoutJob::writeDefaultKeyboardData( const QString& defaultKeyboardPa
               "# Consult the keyboard(5) manual page.\n\n";
 
     stream << "XKBMODEL=\"" << m_model << "\"\n";
-    stream << "XKBLAYOUT=\"" << m_layout << "\"\n";
-    stream << "XKBVARIANT=\"" << m_variant << "\"\n";
-    stream << "XKBOPTIONS=\"\"\n\n";
+    if ( m_additionalLayoutInfo.additionalLayout.isEmpty() )
```
adriaandegroot: Perhaps something along these lines:
```
auto removeEmpties = [](QStringList && l) { l.removeAll(QString()); return l; };
const QStringList layouts(removeEmpties({m_additionalLayoutInfo.additionalLayout, m_layout});
const QStringList variants(removeEmpties({m_additionalLayoutInfo.additionalVariant, m_variant});
```
Er, no, that's yucky. Well, you could bind-by-reference `m_additionalLayoutInfo` and pass in two strings, returning s1,s2 or only s2 depending. And the options string is
```
const QString options = m_additionalLayoutInfo.isEmpty() ? QString() : M-additionalLayoutInfo.groupSwitcher();
```
Then you can keep one set of layout-output lines, outputting only the calculated strings.

kkofler:
> ```c++
> auto removeEmpties = [](QStringList && l) { l.removeAll(QString()); return l; };
> ```

Is it even allowed to call a mutating method like `removeAll` on an rvalue reference?

ivan1986:
checked on endeverOS latest with replace `libcalamares_viewmodule_keyboard.so`

ivan1986:
add some more for keyboard in #2164

ivan1986:
but why is it enabled only for a certain list of layouts, shouldn't it be for everyone except the english layout to add the english layout, which can be done with add english always and remove duplicates

adriaandegroot:
I'm going to merge this, then fix a handful of nitpicks (missing `const`, unnecessary repeated calls to `RemoveEmpty()`, needless method when it could be a free function.

PR #2140, opened 2023-05-20
Initcipiocfg fixes for non x86 cpus and add armInstall option to the partition.conf

Rippanda12:
ARM cpus don't have a vendor tag so the initcpicfg module fails this fix just makes it so if there is a vendor tag to check

armInstall leaves 16M at the start of the disk in automatic partitioning theres usually uboot there


dalto8:
Please separate the initcpio changes and the partition changes into separate PRs.

PR #2142, opened 2023-05-21
partitioning: Add armInstall

Rippanda12:
armInstall leaves 16M at the start of the disk in automatic partitioning theres usually uboot there


demmm:
For clarity about (future) commits/PRs, please use the recommended commit style, see last paragraph:
https://github.com/calamares/calamares/wiki/Develop-Code
and include the module name at the the start of the commit message.

dalto8:
Instead of putting this in global storage, we could store it as a private member of Config such as m_isArmInstall.

You can look at all the other values here if you need an example.

PR #2144, opened 2023-05-22
[packages] Call dnf-3 binary, not dnf

evan-goode:
DNF is being replaced by a new package manager, DNF 5, in Fedora 39+. The `dnf` binary will soon symlink to DNF 5 instead of DNF 5. The old DNF 4 binary will still be (and always has been) available as /usr/bin/dnf-3.

Until Calamares adds support for DNF 5, it should explicitly call the old dnf-3 binary.

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -282,12 +282,12 @@ class PMDnf(PackageManager):
     backend = "dnf"
 
     def install(self, pkgs, from_local=False):
-        check_target_env_call(["dnf", "-y", "install"] + pkgs)
+        check_target_env_call(["dnf-3", "-y", "install"] + pkgs)
 
     def remove(self, pkgs):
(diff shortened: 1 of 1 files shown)


dalto8:
Are we sure that all distros which use `dnf` provide `dnf-3`?

There are several non-RH/Fedora-based distros that also use `dnf`.

evan-goode:
Hmm, good question, they should unless they are really straying from our package spec. `dnf` requires `python3-dnf` which provides `/usr/bin/dnf-3`: https://github.com/rpm-software-management/dnf/blob/1ef42f78b51714edaf3f53433841afef1738886a/dnf.spec#LL349C13-L349C13. The way it's implemented, dnf is just a symlink to dnf-3.

I checked OpenSUSE and Mageia and they have dnf-3.

PR #2146, opened 2023-05-23
Create key file for LUKS2 devices

abalfoort:
This change will handle LUKS2 encrypted devices.

Five installation scenarios were used for testing:
[luks2_testing.tar.gz](https://github.com/calamares/calamares/files/11543424/luks2_testing.tar.gz)

Note:
Not tested with an encrypted boot because Grub2 cannot handle that yet.

To safely convert LUKS1 devices to LUKS2 devices I created a separate module that is run after the partition module: [partition-luks2](https://github.com/abalfoort/calamares-settings-solydxk/tree/master/calamares-modules/partition-luks2)

If this is something that should be implemented in the partition module, please let me know.

### DIFF
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -164,26 +164,20 @@ setupLuks( const LuksDevice& d )
         cDebug() << "Setup LUKS1 for " << d.device;
     }
 
+    // Check the number of slots used
     // Output of LUKS1 and LUKS2 differ
     auto search_pattern = luks2_hash.isEmpty() ? QStringLiteral( R"(\d+:\s*enabled)" ) : QStringLiteral( R"(\d+:\s*luks2)" );
     QRegularExpression slots_re( search_pattern, QRegularExpression::CaseInsensitiveOption );
     slots_count = luks_dump.getOutput().count( slots_re );
 
-    // This should be a clean install: remove all existing keys, except the first slot
-    // Remove slots but leave the first
-    for ( int i = 1; i < slots_count; i++ )
+    if ( ( luks2_hash.isEmpty() && slots_count == 8 ) ||
+         ( !luks2_hash.isEmpty() && slots_count == 32 ))
     {
-        cDebug() << Logger::SubEntry << "Remove surplus key slot: " << QString::number( i ) << " from " << d.device;
-        CalamaresUtils::System::instance()->targetEnvCommand(
-            { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksKillSlot" ), d.device, QString::number( i ) },
-            QString(),
-            d.passphrase,
-            std::chrono::seconds( 5 ) );
+        // No key slots left: return gracefully
+        return true;
     }
 
-    // For LUKS1 you cannot change the key in slot 1, or you'll end up in Grub command.
-    // This doesn't seem the case for LUKS2.
-    // It has not been tested with a LUKS2 encrypted boot because Grub2 does not yet support this.
+    // Add the key to the keyfile
     QStringList args = { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksAddKey" ) };
     if ( !luks2_hash.isEmpty() )
     {
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -138,7 +138,6 @@ static bool
 setupLuks( const LuksDevice& d, const QString& luks2Hash )
 {
     // Get luksDump for this device
-    int slots_count = 0;
     auto luks_dump = CalamaresUtils::System::instance()->targetEnvCommand(
         { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksDump" ), d.device },
         QString(),
@@ -153,31 +152,42 @@ setupLuks( const LuksDevice& d, const QString& luks2Hash )
 
     // Check LUKS version
     int luks_version = 0;
-    QRegularExpression version_re( QStringLiteral( R"(version:\s*([0-9]))" ), QRegularExpression::CaseInsensitiveOption );
+    QRegularExpression version_re( QStringLiteral( R"(version:\s*([0-9]))" ),
+                                   QRegularExpression::CaseInsensitiveOption );
     QRegularExpressionMatch match = version_re.match( luks_dump.getOutput() );
     if ( ! match.hasMatch() )
     {
         cWarning() << "Could not get LUKS version on device: " << d.device;
         return false;
     }
-    luks_version = match.captured(1).toInt();
+    bool ok;
+    luks_version = match.captured(1).toInt(&ok);
+    if( ! ok )
+    {
+        cWarning() << "Could not get LUKS version on device: " << d.device;
+        return false;
+    }
     cDebug() << "LUKS" << luks_version << " found on device: " << d.device;
 
-    // Check the number of slots used
-    // Output of LUKS1 and LUKS2 differ
-    auto search_pattern = luks_version == 1 ? QStringLiteral( R"(\d+:\s*enabled)" ) : QStringLiteral( R"(\d+:\s*luks2)" );
-    QRegularExpression slots_re( search_pattern, QRegularExpression::CaseInsensitiveOption );
-    slots_count = luks_dump.getOutput().count( slots_re );
-    if ( luks_version == 1 && slots_count == 8 )
+    // Check the number of slots used for LUKS1 devices
+    if ( luks_version == 1 )
     {
-        cWarning() << "No key slots left on LUKS1 device: " << d.device;
-        return false;
+        QRegularExpression slots_re( QStringLiteral( R"(\d+:\s*enabled)" ),
+                                     QRegularExpression::CaseInsensitiveOption );
+        if ( luks_dump.getOutput().count( slots_re ) == 8 )
+        {
+            cWarning() << "No key slots left on LUKS1 device: " << d.device;
+            return false;
+        }
     }
 
     // Add the key to the keyfile
-    QStringList args_luks1 = { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksAddKey" ), d.device, keyfile };
-    QStringList args_luks2 = { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksAddKey" ), "--pbkdf", luks2Hash, d.device, keyfile };
-    QStringList args = luks_version == 1 ? args_luks1 : args_luks2;
+    QStringList args = { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksAddKey" ), d.device, keyfile };
+    if ( luks_version == 2 && luks2Hash != QString() )
+    {
+        args.insert(2, "--pbkdf");
+        args.insert(3, luks2Hash);
+    }
     auto r = CalamaresUtils::System::instance()->targetEnvCommand(
         args,
         QString(),
@@ -263,12 +273,6 @@ LuksBootKeyFileJob::exec()
             "LuksBootKeyFile", tr( "No partitions are defined." ), Calamares::JobResult::InvalidConfiguration );
     }
 
-    if ( m_luks2Hash.isEmpty() )
-    {
-        return Calamares::JobResult::internalError(
-            "LuksBootKeyFile", tr( "No luks2Hash is set." ), Calamares::JobResult::InvalidConfiguration );
-    }
-
     cDebug() << "There are" << s.devices.count() << "LUKS partitions";
     if ( s.devices.count() < 1 )
     {
@@ -336,7 +340,7 @@ void
 LuksBootKeyFileJob::setConfigurationMap( const QVariantMap& configurationMap )
 {
     m_luks2Hash = CalamaresUtils::getString(
-        configurationMap, QStringLiteral( "luks2Hash" ), QStringLiteral( "pbkdf2" ) );
+        configurationMap, QStringLiteral( "luks2Hash" ), QString() );
 }
 
 CALAMARES_PLUGIN_FACTORY_DEFINITION( LuksBootKeyFileJobFactory, registerPlugin< LuksBootKeyFileJob >(); )
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -137,32 +137,64 @@ generateTargetKeyfile()
 static bool
 setupLuks( const LuksDevice& d )
 {
-    // Sometimes this error is thrown: "All key slots full"
-    // luksAddKey will fail. So, remove the first slot to make room
+    // Get luksDump for this device
+    int slots_count = 0;
     auto luks_dump = CalamaresUtils::System::instance()->targetEnvCommand(
-        { "cryptsetup", "luksDump", d.device }, QString(), QString(), std::chrono::seconds( 5 ) );
-    if ( luks_dump.getExitCode() == 0 )
+        { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksDump" ), d.device },
+        QString(),
+        QString(),
+        std::chrono::seconds( 5 ) );
+    if ( luks_dump.getExitCode() != 0 )
     {
-        QRegularExpression re( QStringLiteral( R"(\d+:\s*enabled)" ), QRegularExpression::CaseInsensitiveOption );
-        int count = luks_dump.getOutput().count( re );
-        cDebug() << "Luks Dump slot count: " << count;
-        if ( count >= 7 )
-        {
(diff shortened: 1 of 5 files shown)


dalto8:
I need to spend some additional time reviewing this in more detail but there are a couple of things I noticed in a brief review.

People use Calamares to install into existing luks partitions so it isn't safe to remove all the keys from a luks partition.  Even if it was safe, we shouldn't be doing it in the luksbootkeyfile module.

Why have you started creating a keyfile when `/boot` is unencrypted?  I don't see a reason to have a keyfile in that scenario.


> To safely convert LUKS1 devices to LUKS2 devices I created a separate module that is run after the partition module: [partition-luks2](https://github.com/abalfoort/calamares-settings-solydxk/tree/master/calamares-modules/partition-luks2)

The partition module supports luks2 since a recent commit.

abalfoort:
> People use Calamares to install into existing luks partitions so it isn't safe to remove all the keys from a luks partition. Even if it was safe, we shouldn't be doing it in the luksbootkeyfile module.

Normally, I would agree. However, if you user decides not to format the root partition luksbootkeyfile will add a new key to the key file. Eventually Calamares will crash (I had that) when there are no more free slots.

> Why have you started creating a keyfile when `/boot` is unencrypted? I don't see a reason to have a keyfile in that scenario.

You are right, I'll have to look at that. Thanks.

> The partition module supports luks2 since a recent commit.

True, but in some scenarios the user would end up in a unbootable system. E.g. if the user decides a simple install and erase the system and create a fully encrypted system with LUKS2, the root partition is going to be LUKS2 encrypted. Grub2 doesn't like that, yet.

The module I created converts only those devices from LUKS1 to LUKS2 if it can safely be converted. So, in the above example the root partition is still LUKS1 while the /home partition is being converted in LUKS2. A compromise as long as Grub2 cannot handle LUKS2 partitions.

dalto8:
> if you user decides not to format the root partition luksbootkeyfile will add a new key to the key file. Eventually Calamares will crash (I had that) when there are no more free slots.

In that case, we should check for that and simply not add a keyfile if there are no slots remaining.

Failing to create a keyfile shouldn't stop the system from booting.  It may result in a user having to enter the password more than once if they have multiple luks volumes.  That seems like a much lower risk than stripping off all the existing keyslots.

kkofler:
Yes, IMHO, stripping existing keys from an existing LUKS group is a no-go.

abalfoort:
Checking for the number of key slots used is simple in the case of LUKS1. For LUKS2 it is not easy to determine:

> [The number of keyslots is limited only by the provided header area size.](https://gitlab.com/cryptsetup/cryptsetup/blob/master/docs/on-disk-format-luks2.pdf)

LUKS2 doesn't seem to have a hard limit on the number of slots available, like LUKS1 has. So, I can check for LUKS1 slots, but I don't know how to calculate the amount of slots available from the LUKS2 header information.

Or I can assume there are 32 LUKS2 slots. That should be more than enough for this case:
```
    if ( ( luks2_hash.isEmpty() && slots_count == 8 ) ||
         ( !luks2_hash.isEmpty() && slots_count == 32 ))
    {
        // No key slots left: return gracefully
        return true;
    }
```

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -137,32 +137,58 @@ generateTargetKeyfile()
...
+        cWarning() << "Could not get LUKS information on " << d.device << ':' << luks_dump.getOutput() << "(exit code"
+                   << luks_dump.getExitCode() << ')';
+        return false;
     }
 
-    // Adding the key can take some times, measured around 15 seconds with
-    // a HDD (spinning rust) and a slow-ish computer. Give it a minute.
+    // Check for LUKS2
+    QString luks2_hash = QString();
+    QRegularExpression pbkdf_re( QStringLiteral( R"(pbkdf:\s*(.*))" ), QRegularExpression::CaseInsensitiveOption );
```
dalto8: If we want to know if it luks2 we should check the `Version` or export it as JSON and check `type`

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -137,32 +137,58 @@ generateTargetKeyfile()
...
+    // Output of LUKS1 and LUKS2 differ
+    auto search_pattern = luks2_hash.isEmpty() ? QStringLiteral( R"(\d+:\s*enabled)" ) : QStringLiteral( R"(\d+:\s*luks2)" );
+    QRegularExpression slots_re( search_pattern, QRegularExpression::CaseInsensitiveOption );
+    slots_count = luks_dump.getOutput().count( slots_re );
+
+    if ( ( luks2_hash.isEmpty() && slots_count == 8 ) ||
+         ( !luks2_hash.isEmpty() && slots_count == 32 ))
+    {
+        // No key slots left: return gracefully
+        return true;
```
dalto8: This should return `false`

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -137,32 +137,58 @@ generateTargetKeyfile()
...
+    }
+
+    // Check the number of slots used
+    // Output of LUKS1 and LUKS2 differ
+    auto search_pattern = luks2_hash.isEmpty() ? QStringLiteral( R"(\d+:\s*enabled)" ) : QStringLiteral( R"(\d+:\s*luks2)" );
+    QRegularExpression slots_re( search_pattern, QRegularExpression::CaseInsensitiveOption );
+    slots_count = luks_dump.getOutput().count( slots_re );
+
+    if ( ( luks2_hash.isEmpty() && slots_count == 8 ) ||
+         ( !luks2_hash.isEmpty() && slots_count == 32 ))
```
dalto8: For luks2, the number of slots doesn't matter but the total area does.  You can add up the area each key is using and then compare it to the total to determine if there is sufficient room.

Alternatively, we could just not check for luks2 since the chance of overflowing a luks2 volume with keys is fairly low.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -137,32 +137,58 @@ generateTargetKeyfile()
...
+    {
+        // No key slots left: return gracefully
+        return true;
+    }
+
+    // Add the key to the keyfile
+    QStringList args = { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksAddKey" ) };
+    if ( !luks2_hash.isEmpty() )
+    {
+        args << "--pbkdf" << luks2_hash;
```
dalto8: I am not sure that grabbing the algorithm from another key is the right thing here.  I would say that we should either use the default or add a config option for the user to specify it.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -137,32 +137,58 @@ generateTargetKeyfile()
...
     }
 
-    // Adding the key can take some times, measured around 15 seconds with
-    // a HDD (spinning rust) and a slow-ish computer. Give it a minute.
+    // Check for LUKS2
+    QString luks2_hash = QString();
+    QRegularExpression pbkdf_re( QStringLiteral( R"(pbkdf:\s*(.*))" ), QRegularExpression::CaseInsensitiveOption );
+    QRegularExpressionMatch match = pbkdf_re.match( luks_dump.getOutput() );
+    if ( match.hasMatch() ) {
+        luks2_hash = match.captured(1);
```
dalto8: Why is `1` used here?

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -297,8 +323,8 @@ LuksBootKeyFileJob::exec()
 
         if ( !setupLuks( d ) )
             return Calamares::JobResult::error(
-                tr( "Encrypted rootfs setup error" ),
-                tr( "Could not configure LUKS key file on partition %1." ).arg( d.device ) );
+                tr( "Encryption setup error" ),
+                tr( "Could not configure LUKS on partition %1." ).arg( d.device ) );
```
dalto8: I think the prior error was more descriptive.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -297,8 +323,8 @@ LuksBootKeyFileJob::exec()
 
         if ( !setupLuks( d ) )
             return Calamares::JobResult::error(
-                tr( "Encrypted rootfs setup error" ),
```
dalto8, line 300: I know you didn't write this part, but I wonder if this should not throw an error?  It seems like in most cases we would not want a failure here to cause the install to fail.

dalto8 (changes requested):

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
abalfoort: How does one export the output of `cryptSetup luksDump` to json and isn't that a lot of overhead in this use case?

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
abalfoort: Agreed, but then we should not return `Calamares::JobResult::error` in exec() as you commented in your last remark below.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
abalfoort: I prefer the latter. I already felt I was overreaching here. Just a check for LUKS1 should be enough.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
abalfoort: The LUKS2 hash (pbkdf, argon2i, argon2id) should be configured in the partition module. It is needed when creating the encrypted partitions. Here, we add a key to the keyfile and for LUKS2 that means we need to provide the hash key used when creating the encrypted partition. At least, that's what I understood and that's why I checked for the used hash of that partition. Am I wrong?

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
abalfoort: I save only the value of pgkdf: the first set of parenthesis has index 0, which holds the total string of "pbkdf: value" and the second pair of parenthesis has index 1, which holds only "value".

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
abalfoort: I agree, whether or not the keyfile creation is successful or not, installation should continue to prevent an unbootable system.

[Edit]
We probably should add something like:
`JobResult::warning( const QString& message, const QString& details )`
to [Job.cpp](https://github.com/calamares/calamares/blob/calamares/src/libcalamares/Job.cpp)

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
abalfoort: I changed the error message to distinct between two identical error messages. So, it would be easier to see which code threw the error.

[Edit]
This would be solved if we replace the below error message with a warning.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
dalto8: There is an option to `cryptsetup` for JSON output.  However, you don't need to do that, I think it is fine to check `Version` instead.  I was just pointing out that there were different options for how to check it.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
dalto8: You could look at the code in the partition module to verify but IIRC you can't select it there.  It is using the default.

Either way, I think it is better to either use the default here or offer a config option.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
dalto8: The prior message noted that the failure was related to the key file and the rootfs.  Now it is either overly generic or misleading depending on how you read it.

src/modules/luksbootkeyfile/luksbootkeyfile.conf
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: no
+# SPDX-License-Identifier: CC0-1.0
+#
+# Luksbootkeyfile configuration. A key file is created for the
+# LUKS encrypted devices.
+---
+# Set Password-Based Key Derivation Function (PBKDF) algorithm
+# for LUKS keyslot. The PBKDF can be: pbkdf2, argon2i or argon2id.
+# Default: pbkdf2
+luks2Hash: pbkdf2
```
dalto8: See above for comment on the default here.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -295,13 +321,22 @@ LuksBootKeyFileJob::exec()
...
     }
 
     return Calamares::JobResult::ok();
 }
 
+void
+LuksBootKeyFileJob::setConfigurationMap( const QVariantMap& configurationMap )
+{
+    m_luks2Hash = CalamaresUtils::getString(
+        configurationMap, QStringLiteral( "luks2Hash" ), QStringLiteral( "pbkdf2" ) );
```
dalto8: We don't need a default at all here, do we?  Can't we allow this to be unset and run the command without a specific setting in that case?

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -135,34 +135,54 @@ generateTargetKeyfile()
...
+    // Check LUKS version
+    int luks_version = 0;
+    QRegularExpression version_re( QStringLiteral( R"(version:\s*([0-9]))" ), QRegularExpression::CaseInsensitiveOption );
+    QRegularExpressionMatch match = version_re.match( luks_dump.getOutput() );
+    if ( ! match.hasMatch() )
+    {
+        cWarning() << "Could not get LUKS version on device: " << d.device;
+        return false;
+    }
+    luks_version = match.captured(1).toInt();
```
dalto8: We should check to ensure this conversion worked.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -135,34 +135,54 @@ generateTargetKeyfile()
...
+    // Output of LUKS1 and LUKS2 differ
+    auto search_pattern = luks_version == 1 ? QStringLiteral( R"(\d+:\s*enabled)" ) : QStringLiteral( R"(\d+:\s*luks2)" );
+    QRegularExpression slots_re( search_pattern, QRegularExpression::CaseInsensitiveOption );
+    slots_count = luks_dump.getOutput().count( slots_re );
+    if ( luks_version == 1 && slots_count == 8 )
+    {
+        cWarning() << "No key slots left on LUKS1 device: " << d.device;
+        return false;
+    }
+
```
dalto8: We only need to do this if version is `1` I think.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -135,34 +135,54 @@ generateTargetKeyfile()
...
+    // Add the key to the keyfile
+    QStringList args_luks1 = { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksAddKey" ), d.device, keyfile };
+    QStringList args_luks2 = { QStringLiteral( "cryptsetup" ), QStringLiteral( "luksAddKey" ), "--pbkdf", luks2Hash, d.device, keyfile };
+    QStringList args = luks_version == 1 ? args_luks1 : args_luks2;
```
dalto8: Can we avoid using 3 lists here?

The only difference is the addition of a single option.  Couldn't we use a simply conditional instead?

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -243,6 +263,12 @@ LuksBootKeyFileJob::exec()
             "LuksBootKeyFile", tr( "No partitions are defined." ), Calamares::JobResult::InvalidConfiguration );
     }
 
+    if ( m_luks2Hash.isEmpty() )
+    {
+        return Calamares::JobResult::internalError(
+            "LuksBootKeyFile", tr( "No luks2Hash is set." ), Calamares::JobResult::InvalidConfiguration );
+    }
+
```
dalto8: See my comment below but I don't think this should result in a failure.

dalto8 (changes requested):

dalto8 (approved):

PR #2159, opened 2023-06-18
packages: Added remove_deps and try_remove_deps operations for pacman class

D3vil0p3r:
Added `remove_deps` and `try_remove_deps` operations in `packages` module for Pacman class in order to have an operation to remove a package and all its dependencies. Useful for OS Calamares configuration that would like to allow the user to set a "No Desktop" environment by removing the current DE like, for example, `gnome` group package.

Currently this action could be by writing a `pacman -Rcns` command on `shellprocess` module, but having this kind of operation in `packages` module could provide immediate action for the user.

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -114,16 +114,6 @@ class PackageManager(metaclass=abc.ABCMeta):
             list of package names
         """
         pass
-    
-    @abc.abstractmethod
-    def remove_deps(self, pkgs):
-        """
-        Removes packages and their dependencies.
-
-        @param pkgs: list[str]
-            list of package names
-        """
-        pass
 
     @abc.abstractmethod
     def update_db(self):
@@ -167,7 +157,7 @@ class PackageManager(metaclass=abc.ABCMeta):
             self.remove([packagedata["package"]])
             self.run(packagedata["post-script"])
             
-    def remove_deps_package(self, packagedata):
+    def remove_with_deps_package(self, packagedata):
         """
         Remove a package and its dependencies from a single entry in the remove list.
         This can be either a single package name, or an object
@@ -177,10 +167,10 @@ class PackageManager(metaclass=abc.ABCMeta):
         @param packagedata: str|dict
         """
         if isinstance(packagedata, str):
-            self.remove_deps([packagedata])
+            self.remove_with_deps([packagedata])
         else:
             self.run(packagedata["pre-script"])
-            self.remove_deps([packagedata["package"]])
+            self.remove_with_deps([packagedata["package"]])
             self.run(packagedata["post-script"])
 
     def operation_install(self, package_list, from_local=False):
@@ -246,7 +236,7 @@ class PackageManager(metaclass=abc.ABCMeta):
             for package in package_list:
                 self.remove_package(package)
     
-    def operation_remove_deps(self, package_list):
+    def operation_remove_with_deps(self, package_list):
         """
         Removes the list of packages named in @p package_list and their dependencies .
         These can be strings -- plain package names -- or
@@ -261,10 +251,10 @@ class PackageManager(metaclass=abc.ABCMeta):
         NOTE: exceptions should be raised to indicate failure
         """
         if all([isinstance(x, str) for x in package_list]):
-            self.remove_deps(package_list)
+            self.remove_with_deps(package_list)
         else:
             for package in package_list:
-                self.remove_deps_package(package)
+                self.remove_with_deps_package(package)
     
     def operation_try_remove(self, package_list):
         """
@@ -280,7 +270,7 @@ class PackageManager(metaclass=abc.ABCMeta):
             except subprocess.CalledProcessError:
                 libcalamares.utils.warning("Could not remove package %s" % package)
                 
-    def operation_try_remove_deps(self, package_list):
+    def operation_try_remove_with_deps(self, package_list):
         """
         Same relation as try_install has to install, except it removes
         packages and their dependencies instead. Packages are removed one-by-one.
@@ -290,7 +280,7 @@ class PackageManager(metaclass=abc.ABCMeta):
         """
         for package in package_list:
             try:
-                self.remove_deps_package(package)
+                self.remove_with_deps_package(package)
             except subprocess.CalledProcessError:
                 libcalamares.utils.warning("Could not remove package %s" % package)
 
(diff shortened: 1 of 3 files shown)


src/modules/packages/packages.schema.yaml
```diff
@@ -44,8 +44,10 @@ properties:
                 #       need their own little schema.
                 install: { type: array }
                 remove: { type: array }
+                remove_deps: { type: array }
```
dalto8: `remove_deps` makes it sounds like it removes the dependencies only.  I think this name should be clearer.

src/modules/packages/packages.conf
```diff
@@ -106,7 +106,7 @@ pacman:
 #       useful if you have a static package archive on the install media.
 #       The *pacman* package manager is the only one to specially support
 #       this operation (all others treat this the same as *install*).
-# - *remove*, *try_remove*: will call the package manager to
+# - *remove*, *remove_deps*, *try_remove*, *try_remove_deps*: will call the package manager to
```
dalto8: This needs a better explanation of the differences.

src/modules/packages/main.py
```diff
@@ -464,6 +524,10 @@ def install(self, pkgs, from_local=False):
     def remove(self, pkgs):
         self.reset_progress()
         self.run_pacman(["pacman", "-Rs", "--noconfirm"] + pkgs, True)
+        
+    def remove_deps(self, pkgs):
+        self.reset_progress()
+        self.run_pacman(["pacman", "-Rcns", "--noconfirm"] + pkgs, True)
```
dalto8: This should probably be `"-Rc"`.  We don't use `n` in the other options so we probably shouldn't do it here.  Combining `-Rs` and -`Rc` is not a good idea in an installer.  It removes packages recursively in both directions and nearly always takes more packages away than it should.  You would commonly have situations where you are removing packages the user selected to install in prior steps.  `-Rc` is sufficient to ensure your package gets removed.

src/modules/packages/main.py
```diff
@@ -114,6 +114,16 @@ def remove(self, pkgs):
             list of package names
         """
         pass
+    
+    @abc.abstractmethod
+    def remove_deps(self, pkgs):
```
dalto8: This abstract method is only implemented by a single package manager class.  What happens when someone who isn't using that package manager tries to use that functionality?

dalto8 (changes requested):
In addition to the specific comments below, I have a few broader concerns with this:
* I don't see an obvious use case for this in an installer
* It seems overly specific to pacman
* If we need this functionality, why not add an option to the pacman settings in the conf file that let's the user specify the options to send to pacman?  That should serve the same purpose in a more configurable and less obtrusive way.

src/modules/packages/packages.conf
D3vil0p3r: Fixed

src/modules/packages/packages.schema.yaml
D3vil0p3r: True. Atm I have these proposals in mind:
* try_remove_with_deps
* try_remove_w_deps
* try_remove_all

I chose "try_remove_with_deps".

Ofc same for "remove_*".

src/modules/packages/main.py
D3vil0p3r: True. I will remove it from here since what we need is already defined in Pacman class.
UPDATE: fixed.

D3vil0p3r:
The current use case in my mind is the one where, for example, I have an Arch-based Live Environment with GNOME and Calamares as installer. On Calamares I would like to provide several environments to the users to be chosen, for example No Desktop, GNOME, XFCE.

In the 1st and 3rd cases, I need that the `gnome` package group will be removed because no needed.

Currently I can do this by shellprocess by explicitely calling `pacman -Rc` before the install of these environments, or contextualprocess module, but it means to add the explicit removal command for all environment choices different from GNOME.

By implementing this operation on packages module could be more comfortable and immediate in my opinion because I need only to add the name of the package group in the conf file.

About your last proposal, do you mean to implement a new setting in packages.conf under `pacman:` for enabling the chance to remove either just the package or the package and its deps? If so, Im afraid to not be able to manage cases where I need to remove both just a package and a package with all its dependency. Is this the case of your last bullet?

dalto8:
> About your last proposal, do you mean to implement a new setting in packages.conf under `pacman:` for enabling the chance to remove either just the package or the package and its deps? 

Sort of, I was thinking more like you could specify the exact options you wanted like `-Rus`, `-Rc` or `-Rs` or whatever else you needed.

> If so, Im afraid to not be able to manage cases where I need to remove both just a package and a package with all its dependency. Is this the case of your last bullet?

In this case, you could use more than one instance of the packages module with different options.

This implementation would be simpler to maintain and easier to understand.

src/modules/packages/main.py
D3vil0p3r: Fixed

D3vil0p3r:
> > If so, Im afraid to not be able to manage cases where I need to remove both just a package and a package with all its dependency. Is this the case of your last bullet?
> 
> In this case, you could use more than one instance of the packages module with different options.
> 
I thought all day about this, but I think that if we have also the case of remove with dependencies in `packages` module, could make it more "complete" about the generic removal cases, then it does not need to be modified. I think that using an additional package module with different options could be uncomfortable if I dress the user side.

dalto8:
> I thought all day about this, but I think that if we have also the case of remove with dependencies in `packages` module, could make it more "complete" about the generic removal cases, then it does not need to be modified. I think that using an additional package module with different options could be uncomfortable if I dress the user side.

The problem is that it doesn't make any sense for most of the package managers.  The question of "Which makes more sense for Calamares?" might have a different answer in this case.

Adding a single option to the conf file meets the same need but only impacts pacman seems a lot cleaner to me.

D3vil0p3r:
If we add both? And furthermore, how implement your proposal? Which files must be edited?

dalto8:
> If we add both?

It would make no sense to do both.  They both achieve the same goal

> And furthermore, how implement your proposal? Which files must be edited?

You would add an option to the conf file in the existing file.

Then you would change the below line so it reads the setting instead of hard-coding `-Rs`.

```
 self.run_pacman(["pacman", "-Rs", "--noconfirm"] + pkgs, True)
```

dalto8:
Since this has been dormant for quite some time I am going to close it for now.

PR #2161, opened 2023-06-23
partitioning: Show/hide "Boot partition not encrypted" warning

abalfoort:
This change will make showing or hiding the "Boot partition not encrypted" warning configurable.

Default is the current behavior: show the warning when a non-encrypted boot partition is found while encrypting the root partition.

### DIFF
--- a/src/modules/partition/PartitionViewStep.cpp
+++ b/src/modules/partition/PartitionViewStep.cpp
@@ -615,27 +615,32 @@ PartitionViewStep::onLeave()
 
             // If the root partition is encrypted, and there's a separate boot
             // partition which is not encrypted
-            if ( ( root_p->fileSystem().type() == FileSystem::Luks && boot_p->fileSystem().type() != FileSystem::Luks )
-                 || ( root_p->fileSystem().type() == FileSystem::Luks2
-                      && boot_p->fileSystem().type() != FileSystem::Luks2 ) )
+            if ( m_config->showNotEncryptedBootMessage() )
             {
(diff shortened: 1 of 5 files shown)


src/modules/partition/PartitionViewStep.cpp
```diff
@@ -615,27 +615,32 @@ PartitionViewStep::onLeave()
...
-                                  "You may continue if you wish, but filesystem "
-                                  "unlocking will happen later during system startup."
-                                  "<br/>To encrypt the boot partition, go back and "
-                                  "recreate it, selecting <strong>Encrypt</strong> "
-                                  "in the partition creation window." );
-
-                QMessageBox mb( QMessageBox::Warning, message, description, QMessageBox::Ok, m_manualPartitionPage );
-                Calamares::fixButtonLabels( &mb );
-                mb.exec();
+                if ( ( root_p->fileSystem().type() == FileSystem::Luks
```
dalto8: We shouldn't have an `if` block with nothing but another `if` block.  I would create a function for checking root_p and boot_p and then use a single `if` that should be readable.

dalto8 (changes requested):

src/modules/partition/PartitionViewStep.cpp
abalfoort: You're right. This was the lazy solution. I used shouldWarnForGPTOnBIOS as an example.

dalto8 (approved):

PR #2163, opened 2023-07-06
Fixing partutils.cpp

1337Skid:
This is a little PR to fix the error during compilation of calamares...


dalto8:
Thanks! Can you apply the Calamares coding style to it using `clang-format`(Or by re-adding the spaces around the parens you removed)?

1337Skid:
Oh yeah I forgot about it, will edit it right now.

PR #2164, opened 2023-07-08
add keyboard layout switch selector

ivan1986:
on keyboard select page select box for switch layout toggle method
extract grp: options

### DIFF
--- a/src/modules/keyboard/Config.cpp
+++ b/src/modules/keyboard/Config.cpp
@@ -196,7 +196,7 @@ Config::Config( QObject* parent )
              [ & ]( int index )
              {
                  m_selectedVariant = m_keyboardVariantsModel->key( index );
-                 Config::xkbChanged();
+                 xkbChanged();
                  emit prettyStatusChanged();
              } );
     connect( m_KeyboardGroupSwitcherModel,
@@ -204,7 +204,7 @@ Config::Config( QObject* parent )
              [ & ]( int index )
              {
                  m_selectedGroup = m_KeyboardGroupSwitcherModel->key( index );
-                 Config::xkbChanged();
+                 xkbChanged();
                  emit prettyStatusChanged();
              } );
 
--- a/src/modules/keyboard/Config.cpp
+++ b/src/modules/keyboard/Config.cpp
@@ -160,6 +160,7 @@ Config::Config( QObject* parent )
     , m_keyboardModelsModel( new KeyboardModelsModel( this ) )
     , m_keyboardLayoutsModel( new KeyboardLayoutModel( this ) )
     , m_keyboardVariantsModel( new KeyboardVariantsModel( this ) )
+    , m_keyboardGroupsModel( new KeyboardGroupsModel( this ) )
 {
     m_setxkbmapTimer.setSingleShot( true );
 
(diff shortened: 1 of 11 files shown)


ivan1986:
checked on endeverOS latest with replace `libcalamares_viewmodule_keyboard.so`

adriaandegroot:
I can't exactly tell what this is supposed to add / fix / improve, really. Can you provide a screenshot from your branch, and describe what the change achieves?

My guess based on the code is "if there are multiple layouts in use (e.g. ru,us or ar,us both of which make sense if you pick a different language to install in) then having a drop-down to configure the keyboard-layout switcher for X11 (to pick e.g. right-alt as the switcher or some other key) is a nice-to-have so that users do not need to change that post-installation."

@marcan you **just** added Wayland / locale1 support for keyboard layouts -- does that support a way of changing the layout switcher as well? Then we will need something extra for the Wayland case, as well. Or maybe we will need a way to suppress the drop-down.

ivan1986:
yes - it all select box for keyboard switch

before
before
after
after

adriaandegroot:
Can we move the switch drop-down next to *typ klaviaturi* to save some vertical space? There are rarely long names in the keyboard-type drop-down.

src/modules/keyboard/Config.h
```diff
@@ -58,6 +59,9 @@ class Config : public QObject
      * (dvorak).
      */
     KeyboardVariantsModel* keyboardVariants() const;
+    /* A group describes a toggle groups of change layouts
+     */
+    KeyboardGroupsModel* keyboardGroups() const;
```
adriaandegroot: Let's call it a *GroupSwitcher* then, so `KeyboardGroupSwitcherModel` and `keyboardGroupSwitchers()`

src/modules/keyboard/KeyboardLayoutModel.h
```diff
@@ -154,6 +154,20 @@ class KeyboardVariantsModel : public XKBListModel
     void setVariants( QMap< QString, QString > variants );
 };
 
+/** @brief A list of variants (xkb id and human-readable)
```
adriaandegroot: Group switching key-combinations, not variants. The comment does not match the code.

src/modules/keyboard/KeyboardPage.ui
```diff
@@ -117,6 +117,30 @@ SPDX-License-Identifier: GPL-3.0-or-later
      </item>
     </layout>
    </item>
+   <item>
```
adriaandegroot, line 120: I think you can fit this next to the keyboard-type drop-down, instead, to save some vertical space.

adriaandegroot:
Looking good though! My comments are mostly about some copy-paste-o's in the code, not about functionality.

ivan1986:
updated

ivan1986:
screen

src/modules/keyboard/Config.cpp
```diff
@@ -190,25 +191,40 @@ Config::Config( QObject* parent )
                  emit prettyStatusChanged();
              } );
 
-    connect( m_keyboardVariantsModel, &KeyboardVariantsModel::currentIndexChanged, this, &Config::xkbChanged );
+    connect( m_keyboardVariantsModel,
+             &KeyboardVariantsModel::currentIndexChanged,
+             [ & ]( int index )
+             {
+                 m_selectedVariant = m_keyboardVariantsModel->key( index );
+                 Config::xkbChanged();
```
adriaandegroot: Since you are now using a lambda with implicit capture of `this`, you can just call member functions, like `xkbChanged()`. No need for the weird syntax here.

The reason it said `&Config::xkbChanged` in the first place is that that was a member-function-pointer, which has specific syntax.

src/modules/keyboard/Config.cpp
```diff
@@ -190,25 +191,40 @@ Config::Config( QObject* parent )
...
+                 emit prettyStatusChanged();
+             } );
+    connect( m_KeyboardGroupSwitcherModel,
+             &KeyboardGroupsSwitchersModel::currentIndexChanged,
+             [ & ]( int index )
+             {
+                 m_selectedGroup = m_KeyboardGroupSwitcherModel->key( index );
+                 Config::xkbChanged();
```
adriaandegroot: Same here

src/modules/keyboard/keyboardwidget/keyboardglobal.cpp
```diff
@@ -199,6 +199,48 @@ parseKeyboardLayouts( const char* filepath )
...
+    {
+        QByteArray line = fh.readLine();
+
+        // check if we start a new section
+        if ( line.startsWith( '!' ) )
+        {
+            break;
+        }
+
+        // here we are in the model section, otherwise we would continue or break
```
adriaandegroot: Well, in the option section :smiley_cat:

src/modules/keyboard/keyboardwidget/keyboardglobal.cpp
```diff
@@ -199,6 +199,48 @@ parseKeyboardLayouts( const char* filepath )
...
+    QFile fh( filepath );
+    fh.open( QIODevice::ReadOnly );
+
+    if ( !fh.isOpen() )
+    {
+        cDebug() << "X11 Keyboard model definitions not found!";
+        return models;
+    }
+
+    bool modelsFound = findSection( fh, "! option" );
```
adriaandegroot:
```suggestion
    bool optionSectionFound = findSection( fh, "! option" );
```

src/modules/keyboard/keyboardwidget/keyboardglobal.cpp
```diff
@@ -199,6 +199,48 @@ parseKeyboardLayouts( const char* filepath )
...
+    // read the file until the end or until we break the loop
+    while ( modelsFound && !fh.atEnd() )
```
adriaandegroot:
```suggestion
    while ( optionSectionFound && !fh.atEnd() )
```

src/modules/keyboard/keyboardwidget/keyboardglobal.cpp
```diff
@@ -199,6 +199,48 @@ parseKeyboardLayouts( const char* filepath )
...
+
+        // check if we start a new section
+        if ( line.startsWith( '!' ) )
+        {
+            break;
+        }
+
+        // here we are in the model section, otherwise we would continue or break
+        QRegExp rx;
+        rx.setPattern( "^\\s+grp:(\\S+)\\s+(\\w.*)\n$" );
```
adriaandegroot: I just ported all the modules to be Qt6 compatible, which means away from `QRegExp` and to `QRegularExpression`. Take a look at 427311f2c31600e87879ddb1587d5a65f30e3645 for how it can be done.

adriaandegroot:
This is looking really good! One more round with some small C++ fixes and one bigger Qt6-compatibility fix (`QRegularExpression` is also in Qt5, so you don't need to worry if you do not personally have a Qt6 build) and it's ready to go!

Please do add one more commit that adds a description of the change to `CHANGES-3.3` under the "Modules" heading so that distro's can see what kind of features / changes to expect. Your name is already in the contributors list from your non-ascii-keyboard things.

Issue #1082, opened 2019-02-09
Flatpak integration


nintyfan:
**Not way to handle cross-distro packages in installer**
Currently, LiveCD distros don't allow us to select additional packages and each distribution could use different package format.

**Use flatpak**
Flatpak provide mechanism to create use stick with software currently installed on system. Someone could order to pack flatpaked version of LibreOffice onto usb stick, went onto another computer and install LibreOffice on it without network access! Why don't use this mechanism or create similar? Distros could integrate flatpaked version of programs on LiveCD and Calamares could asks, which program to move into installed system.
So:
1. User could still use LiveCD session
2. User could select, which program He/She requiring on destination system
3. Flatpak will get (possibly) new packages and will popularize

abucodonosor:
No .. either calamares itself can be a flatpak package see https://github.com/calamares/calamares/issues/999
or will it intergate such thing.

That is _distro_ issue .. calamares is not an package manager but an installer and all this flatpak , appimage , snapd or whatever else containerISH does not belong to calamares.

If your Distribution allows you to install flataks like from a repo then ask them to implement some modules.
99% of all Distributions won't allow you that

kkofler:
I think this RFE can either be implemented as part of the `packages` module or as a separate, similar module. Support can then be added to the `netinstall` module or, again, a separate, modified version of it. (Since `netinstall` uses `packages` to do the actual package installation, the first part is needed first.)

I think this RFE is actually more realistic than #999 (turning Calamares itself into a Flatpak).

abucodonosor:
@kkofler 

While it can be done some hackISH way .. there is not a generic solution for these.

Unlikely Distribution packages the ISO is made for, flapak can be random , not even implemented in that Distribution , or implemented different , not there by default and so on.

If you look closer the request is more about ' I boot some random iso' and I download 
on the live ISO some random flatpak .. which then calmamares 'should inject' into
the final installed system. .. Calamares is not an package manager , is it ?

This kind things doesn't belong to calamares.

A sort : flatpak modules could be implemented but not that way and for sure cannot be to generic.

adriaandegroot:
The package manager for flatpak is called `flatpak` (see http://docs.flatpak.org/en/latest/using-flatpak.html) which looks like it can be run just like any other package manager in the target system. Adding it as a subclass of *PackageManager* in the *packages* module should be straightforward. Since we just use package names in the config, it's straightforward to put the name of a flatpak in there.

[...]

PR #2174, opened 2023-07-26
Wip/packaging flatpak

nintyfan:
This MR is created to work on flatpak integration. 

Not each feature related to: https://github.com/calamares/calamares/issues/1082 will be addressed. This branch aim only to provide a way to install applications from network, which are not installed on LiveCD/LiveUSB (cause these got copied by root copy mechanism).

Other features, like installing only selected packages without distro installation will be addressed by other MR.

### DIFF
--- a/src/libcalamares/packages/Globals.cpp
+++ b/src/libcalamares/packages/Globals.cpp
@@ -6,7 +6,7 @@
  *   Calamares is Free Software: see the License-Identifier above.
  *
  */
-
+#include "string.h"
 #include "Globals.h"
 
 #include "GlobalStorage.h"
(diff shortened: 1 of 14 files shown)


nintyfan:
Currently only shown list of flatpaks inside configured repo.
Do not blame me. I am beginner at developing software. I do not read coding style, etc. yet, but try to match my style your rules.

dalto8:
If we want to add support for flatpak, I think it would be better to add a proper backend for it to the packages module instead of hacking it in like that.  Look at the backends for the other package managers and add it in a similar way.

Could you describe what you are trying to do with the packagechooser module?

> I do not read coding style, etc. yet, but try to match my style your rules.

You need to run clang-format against all the files you changed to apply the proper Calamares formatting.

nintyfan:
About backends. I look inside package(s?) module. It contains python code with classes for different package manager. Problem is: list of packages in packagechooser is currently read from configuration.

dalto8:
> About backends. I look inside package(s?) module. It contains python code with classes for different package manager. Problem is: list of packages in packagechooser is currently read from configuration.

It combines that list with one from global storage.  Prior modules can add package operations.

https://github.com/calamares/calamares/blob/calamares/src/modules/packages/main.py#L742-L743

nintyfan:
So... I would wrote two new module. One for UI of flatpak package manager and next for fill it (list) with packages.

dalto8:
I think you should start with an explanation of what you are hoping to accomplish.  What we understand that, we can maybe propose a methodology.

nintyfan:
I need to accomplish:
Give a user a way to select packages to install from remote repository and made Calamares install it after/while installing system

In future:
1. Select packages to not copy from installation media or to remove it just after installation
2. (if possible) add possibility to install packages from removable media (with distribution) to installed distribution for cases with internet problems (poor countries).
By writing packages I mean flatpak packages.

dalto8:
For example, you want someone to be able browse something like flathub and pick which flatpaks to install?

Are you also allowing them to choose normal packages to install or only flatpaks?

nintyfan:
Only flatpaks.

dalto8:
In that case, I would use the netinstall module.  It is better for browsing a large list of packages.

The netinstall module can already read a list from a global storage variable.  It then feeds the packages module.

Here is what I would do:
* Add support for a flatpak backend in the packages module and create a PR for that
* Create a module that reads your flatpak repo and puts the list of packages into global storage in the appropriate format
  * You can see how this is done if you look at the netinstall-add logic in the packagechooser module

From there it should just all work.  Your module will load the data for netinstall, netinstall will display to the user.  The user makes the selections and then the packages module will install them.

nintyfan:
Thanks, but...
What about making support for multiple backends. Should I ran multiple instances of netinstall/flatpakfiller?
I see netinstall download some yml file. I see I should use local config and put each package into configuration.

dalto8:
Superseded by #2176

PR #2176, opened 2023-08-01
Wip/packaging flatpak 2

nintyfan:
Sorry. I could not find option to change source branch in PR.


nintyfan:
It is related to: https://github.com/calamares/calamares/pull/2174 . I cannot find option to change source branch and would not rewrite history.

I think everything should work, but one think making impossible to check. I do not fill properly config(globalstorage) for netinstall.

nintyfan:
I realized problem. Calamares setup modules in parallel . So netinstall reads configuration before I prepare list of packages.

dalto8:
OK, a few quick things.
* Please split this into 2 PRs, one for the flatpakinfo module and one for the changes to packages
* The flatpakinfo module should probably be PR'd to the [Calamares-extensions repo](https://github.com/calamares/calamares-extensions) instead of this repo
* Don't use `execlp`.  Instead use the calamares built-in API for running commands such as `CalamaresUtils::System::instance()->runCommand()`

nintyfan:
I tries to test new packages modules. I copy packages.conf into ./build/src/modules/packages/super.conf, change ./build/settings.conf adding packages@super in exec section. Calamares resulting with:
ERROR: Installation failed: "Bad unpackfs configuration" 
2023-08-02 - 16:35:52 [6]:     .. - message: "Bad unpackfs configuration" 
2023-08-02 - 16:35:52 [6]:     .. - details: Source filesystem "/home/slawomir/Projekty/Calamares/calamares/CHANGES" does not exist 

(my translation)

dalto8:
> 2023-08-02 - 16:35:52 [6]: .. - message: "Bad unpackfs configuration"

Did you configure the rest of the settings?  You can't run Calamares without configuring the other modules.

If you look in the developer docs there are instructions on how to test a single module without running Calamares.

nintyfan:
diff --git a/src/modules/packages/packages.conf b/src/modules/packages/packages.conf
index 6e62f4b5f..6e1373103 100644
--- a/src/modules/packages/packages.conf
+++ b/src/modules/packages/packages.conf
@@ -35,7 +35,7 @@
 # Not actually a package manager, but suitable for testing:
 #  - dummy       - Dummy manager, only logs
 #
-backend: dummy
+backend: flatpak
 
 #
 # Often package installation needs an internet connection.
@@ -200,14 +200,8 @@ pacman:
 # This will call the package manager once with the package-names "vi" and
 # "binutils", and then a second time for "wget". When installing large numbers
 # of packages, this can lead to a considerable time savings.
-#
 operations:
   - install:
-    - vi
-    - vi-${LOCALE}
-    - wget
-    - binutils
+    - org.neverball.Neverball
   - remove:
-    - vi
-    - wget
-    - binutils
+    - org.neverball.Neverball

nintyfan:
https://pastebin.com/3UNt8a3b

> cat ../build/global.yaml 
rootMountPoint: /

run with:
./loadmodule -g global.yaml packages

src/modules/packages/main.py
```diff
@@ -337,6 +337,22 @@ def update_system(self):
...
 
+class PMFlatpak(PackageManager):
+    backend = "flatpak"
+
+    def install(self, pkgs, from_local=False):
+        check_target_env_call(["flatpak", "install", "--assumeyes"] + pkgs)
+
+    def remove(self, pkgs):
+        check_target_env_call(["flatpak", "uninstall", "--noninteractive"] + pkgs)
+        check_target_env_call(["flatpak", "uninstall", "--noninteractive", "--unused"])
```
dalto8: I don't think we should be forcibly removing unused.  For several reasons:
* It doesn't seem like something an installer should be doing
* Newer versions of flatpak do this automatically making this redundant
* Depending on the situation, this might get called for every package

dalto8 (changes requested):

dalto8 (approved):

nintyfan:
Hi.
It could be problem, when distro creator adds more repo or the same repo both for user and system? I tested this and in case of presence package in system and user repo, flatpak asks to select source. I do not known how to skip this.

Issue #2061, opened 2022-10-17
Use prev-lang:kbd EFI NVRAM variable for language and keyboard layout


probonopd:
**Is your feature request related to a problem? Please describe.**
It is frustrating that on PCs you always have to set (or, as in the case of Calamares, guess) the language and keyboard layout for each operating system, whereas on Macs this is a per-machine setting.

**Describe the solution you'd like**
Read (and write, if it is not there yet) the EFI NVRAM variable `7C436110-AB2A-4BBB-A880-FE41995C9F82:prev-lang:kbd`. This will:
* Make language and keyboard layout "just work" on Macs, which most likely have this variable already set correctly by macOS
* Make language and keyboard layout persistent between different boot disks and/or operating systems (macOS and helloSystem already interpret it if set; possibly we can get more OSes to do the same)

**Additional context**
macOS uses the EFI NVRAM variable `7C436110-AB2A-4BBB-A880-FE41995C9F82:prev-lang:kbd` to store the language and keyboard layout, which is why you get it right even if you boot off e.g., the network or a DVD. There is no reason why we couldn't use the same EFI NVRAM variable on other EFI based machines as well.

kkofler:
Why do you want this done in Calamares? This is a job for `systemd-localed`, which can handle this at every boot (including the live image boot, so you already get the correct setting before starting Calamares), not just once at install time.

probonopd:
* Currently Calamares makes educated guesses about the keyboard layout based on the language; why not use that variable to get precise information?
* In case that variable is not yet set, imho Calamares should set it, so that `systemd-localed` could pick it up

adriaandegroot:
A PR to implement this would be useful. It could go in the *locale* service (e.g. *libcalamares*), or the *locale* module; wherever the current "guess a layout" happens right now.

- [ ] Try to read the variable (I suppose a large-ish part of the PR is making that happen; I guess running `efivar` might do the trick)
- [ ] Map whatever values there are, to available X11 keyboard layouts
- [ ] Use that as the initial guess instead of what it does now

PR #2178, opened 2023-08-06
keyboard: Allow passing in default layout via environment vars

marcan:
This allows pre-selecting the keyboard layout, for platforms (laptops) where this information is available from firmware.

There is no standard mechanism for this, so let's use environment variables and leave it up to the launching script / tooling to populate them.

### DIFF
--- a/src/modules/keyboard/Config.cpp
+++ b/src/modules/keyboard/Config.cpp
@@ -331,6 +331,14 @@ Config::detectCurrentKeyboardLayout()
         }
     }
 
+    QString default_layout = qgetenv("DEFAULT_XKBLAYOUT");
+    QString default_variant = qgetenv("DEFAULT_XKBVARIANT");
+
+    if (default_layout != "") {
+        currentLayout = default_layout;
+        currentVariant = default_variant;
+    }
+
     //### Layouts and Variants
     QPersistentModelIndex currentLayoutItem = findLayout( m_keyboardLayoutsModel, currentLayout );
     if ( !currentLayoutItem.isValid() && ( ( currentLayout == "latin" ) || ( currentLayout == "pc" ) ) )
@@ -434,6 +442,10 @@ Config::guessLocaleKeyboardLayout()
     {
         return;
     }
+    QString default_layout = qgetenv("DEFAULT_XKBLAYOUT");
+    if (default_layout != "")
+        return;
+
     cScopedAssignment returnToIntial( &m_state, State::Initial );
     m_state = State::Guessing;
 
--- a/src/modules/keyboard/KeyboardLayoutModel.cpp
+++ b/src/modules/keyboard/KeyboardLayoutModel.cpp
@@ -120,6 +120,10 @@ KeyboardModelsModel::KeyboardModelsModel( QObject* parent )
     : XKBListModel( parent )
 {
     m_contextname = "kb_models";
+    QString default_model = qgetenv("DEFAULT_XKBMODEL");
+
+    if (default_model == "")
+        default_model = "pc105";
 
     // The models map is from human-readable names (!) to xkb identifier
     const auto models = KeyboardGlobal::getKeyboardModels();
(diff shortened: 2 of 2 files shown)


dalto8:
A couple of things here:
* Please make sure you apply the Calamares code style with an appropriate version of `clang-format`(I am currently using version 10 but I think we support a version slightly newer than that).
* Do we need to use env variables here?  Since almost everything else in Calamares uses the config files, couldn't we add options in the config for this and the external tooling could modify the config?  That would be more consistent with the rest of Calamares.

adriaandegroot:
@marcan One vaguely-related thing, there is issue #2061 which suggests reading an nvram location to get the hardware keyboard layout. Since that's for roughly the same hardware as what you are targeting, perhaps you can comment on its feasibility? 

What @dalto8 means in the previous comment: there are plenty of `launch-calamares.sh` scripts out there that use `sed` to edit the configuration files before actually starting the installer -- perhaps the launch script could do that to the default layout and Calamares can read a default layout from the configuration file, rather than using environment variables here. That just moves the use-of-environment variables around, though.

marcan:
#2061 is for EFI systems, which Apple Silicon systems aren't (we do provide EFI ourselves, but that's just for boot purposes).

The problem with this stuff is there is no standard for it. Keyboard layouts are complicated and there is no standard system for identifying them. Apple Silicon systems use a numeric ID which we then map to the required xkb configs (and that mapping is subject to change as we improve the xkb situation and figure out what the right language variants are, which is its own adventure). We forward that ID to Linux in the device tree, so our launcher script picks it up from there (i.e. `/proc/device-tree/...`). I spoke to the DT folks about this (and whether it'd be worth inventing some kind of standard we could pre-map to, that would then work for any DT system) and the conclusion was that it's not worth it, and we should just pass through Apple's proprietary ID number and figure out the mapping in userspace (i.e. in the Calamares launcher script).

One peculiarity of Apple keyboards, which AIUI goes back to everything HID since their Aluminum series, is that they use the HID "country code" to identify the keyboard *model*. That doesn't tell you the layout but it tells you at least the xkbmodel. That much is at least a standard property (though using it like this is a bit weird) and could be picked up from sysfs on most Apple machines. But I'm not sure just knowing the model is very useful...

I'm happy switching this to use the config file and editing it in our script, that's not an issue :)

marcan:
I'm going to close this because I just realized we have bigger problems. Calamares currently uses X11 tools (setxkbmap) which won't work in a Wayland world.

The "right" way to do this in the modern world is using the locale1 dbus service, which behind the scenes manages the X11 config as well as serves as the backend for Wayland stuff. I'm going to have to add support for this to the plugin, and at that point the logical way forward for the defaults is to just use whatever the current systemwide default config is. Then the launcher script just has to use `localectl` to pre-configure things and Calamares can pick that up, without any config entries needed.

kkofler:
The issue with using D-Bus services is that they do not work in a chroot. But for setting the host system's keyboard settings, it will probably work. It is just not a replacement for making the settings in the installed system. But your pull request, which disables writing the settings only if we have no chroot, looks viable to me.

kkofler:
Though then again, Calamares runs as root, so the invoked `localectl` will also run as root. How can it then change the keyboard map of the Wayland compositor running under the regular user's account?

marcan:
> Though then again, Calamares runs as root, so the invoked `localectl` will also run as root. How can it then change the keyboard map of the Wayland compositor running under the regular user's account?

If the compositor is configured to follow locale1 (e.g. kwin with `--locale1`) it will work. If it's using user settings (e.g. Calamares was started from a regular user session), I don't have an answer for that. I don't think there's actually a uniform way to ask a compositor to change the keyboard layout in Wayland. `setxkbmap` definitely doesn't work (tested in a simple KDE Wayland session), so this is already broken and my PR won't fix it. I'm not interested in that use case (we use Calamares from a dedicated kwin session as a firstboot script), but if other people are then that's going to be a really interesting conversation to have with the Wayland and compositor people... `setxkbmap` is one of those things that just doesn't exist in the Wayland world.

The pragmatic answer is that those people using Calamares in a user session should force XWayland mode. `setxkbmap` works for *X clients only*, so if Calamares itself is running in X11 mode, it will do what you want. And since my patch enables locale1 only for non-X mode by default, it shouldn't break anything.

PR #2180, opened 2023-08-12
keyboard: Add locale1 support & option to keep defaults

marcan:
On non-X11 systems, setxkbmap & friends can't work (it sort of works in XWayland, but not if Calamares is running in Wayland mode, and not properly). The "modern" way to control systemwide keyboard layouts is using the locale1 DBus interface.

This PR implements locale1 support as an alternative to setxkbmap, which is enabled by default on non-X11 platforms. Since the locale daemon already alters systemwide keyboard configs, this also disables actually writing out the config files if there is no explicit root directory set (changing the current layout and committing it to the config files are the same operation in this model).

To use this with Calamares running under standalone kwin_wayland, the latter has to be started with the `--locale1` option.

In addition, this PR also adds support for keeping the currently set keyboard layout, variant, and model instead of guessing/forcing defaults. This allows system implementers to pre-configure the layout if a sensible default is known on any given platform.

Note that the locale1 stuff itself uses `/etc/X11/xorg.conf.d/00-keyboard.conf` to store the XKB settings (yes, really, even on systems without X), so it truly is redundant to try to write out that file. In addition, we request conversion from XKB to VConsole keymaps from the locale1 API, so the VConsole keymap guessing code is also not used in this case. This is why the config write step is redundant in this case, except for the `/etc/default/keyboard` codepath (which is still available if enabled).

### DIFF
--- a/src/modules/keyboard/CMakeLists.txt
+++ b/src/modules/keyboard/CMakeLists.txt
@@ -3,6 +3,9 @@
 #   SPDX-FileCopyrightText: 2020 Adriaan de Groot <groot@kde.org>
 #   SPDX-License-Identifier: BSD-2-Clause
 #
+
+find_package(Qt5 ${QT_VERSION} CONFIG REQUIRED Core DBus)
+
 calamares_add_plugin(keyboard
     TYPE viewmodule
     EXPORT_MACRO PLUGINDLLEXPORT_PRO
@@ -19,6 +22,8 @@ calamares_add_plugin(keyboard
     RESOURCES
         keyboard.qrc
     SHARED_LIB
+    LINK_LIBRARIES
+        Qt5::DBus
 )
 
 calamares_add_test(keyboardtest SOURCES Tests.cpp SetKeyboardLayoutJob.cpp RESOURCES keyboard.qrc)
--- a/src/modules/keyboard/Config.cpp
+++ b/src/modules/keyboard/Config.cpp
@@ -23,8 +23,13 @@
 #include "utils/Variant.h"
 
 #include <QApplication>
+#include <QGuiApplication>
 #include <QProcess>
 #include <QTimer>
+
+#include <QtDBus/QDBusConnection>
+#include <QtDBus/QDBusInterface>
+#include <QtDBus/QDBusReply>
 
 /* Returns stringlist with suitable setxkbmap command-line arguments
  * to set the given @p model.
@@ -163,7 +168,14 @@ Config::Config( QObject* parent )
              {
                  // Set Xorg keyboard model
                  m_selectedModel = m_keyboardModelsModel->key( index );
-                 QProcess::execute( "setxkbmap", xkbmap_model_args( m_selectedModel ) );
+                 if ( m_useLocale1 )
+                 {
+                     locale1Apply();
+                 }
+                 else
+                 {
+                     QProcess::execute( "setxkbmap", xkbmap_model_args( m_selectedModel ) );
+                 }
                  emit prettyStatusChanged();
              } );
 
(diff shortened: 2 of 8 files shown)


marcan:
@kkofler Ping, I know we were discussing this in #2178 but we should probably move discussion to the new PR :)

src/modules/keyboard/CMakeLists.txt
```diff
@@ -3,6 +3,9 @@
 #   SPDX-FileCopyrightText: 2020 Adriaan de Groot <groot@kde.org>
 #   SPDX-License-Identifier: BSD-2-Clause
 #
+
+find_package(Qt5 ${QT_VERSION} CONFIG REQUIRED Core DBus)
```
adriaandegroot, line 7: :ship: Shouldn't be needed, since those are searched-for -- well, ok, DBus is optional, we may as well make it required at top-level then.

src/modules/keyboard/Config.cpp
```diff
@@ -23,9 +23,14 @@
 #include "utils/Variant.h"
 
 #include <QApplication>
+#include <QGuiApplication>
 #include <QProcess>
 #include <QTimer>
 
+#include <QtDBus/QDBusConnection>
+#include <QtDBus/QDBusInterface>
+#include <QtDBus/QDBusReply>
```
adriaandegroot, line 32: Please don't use `QtModule/` paths in includes. They are not needed if you link to the relevant Qt CMake library (e.g. *Qt5::DBus* adds the `QtDBus/` path into the include flags already, and Qt doesn't have conflicting header-file names across modules).

src/modules/keyboard/Config.cpp
```diff
@@ -163,7 +168,14 @@ Config::Config( QObject* parent )
              {
                  // Set Xorg keyboard model
                  m_selectedModel = m_keyboardModelsModel->key( index );
-                 QProcess::execute( "setxkbmap", xkbmap_model_args( m_selectedModel ) );
+                 if ( m_useLocale1 )
```
adriaandegroot, line 171: :ship: Since there's multiple places where `if(m_useLocale)` happens, I wonder if there's a more elegant way to do this (and then realise that that would be the enemy of the good, here). Some interface or proxy object to map the work to an underlying implementation.

src/modules/keyboard/keyboard.conf
```diff
@@ -21,3 +21,13 @@ convertedKeymapPath: "/lib/kbd/keymaps/xkb"
 # found on Debian-related systems.
 # Defaults to true if nothing is set.
 #writeEtcDefaultKeyboard:   true
+
+# Use the Locale1 service instead of directly managing configuration files.
+# This is the modern mechanism for configuring the systemwide keyboard layout,
+# and works on Wayland compositors to set the current layout.
+# Defaults to false on X11 and true otherwise.
+#useLocale1: true
```
adriaandegroot, line 29: The schema should be extended with these two new keys.

adriaandegroot (commented):
This looks really good, and my comments feel like nagging -- sorry 'about that.

kkofler:
The thing is, in pretty much all distributions, Calamares runs from a standard live session running as a regular user, with `kwin-wayland` started without nonstandard options such as `--locale1`, then Calamares gets started as root (using `pkexec` or `kdesu` or `sudo` or something similar). Setting the systemwide default keyboard as root is not going to have any effect whatsoever on the live session already running as a regular user. The only way to communicate there is over the windowing system protocol, as `setxkbmap` does/did. If there is nothing equivalent in the Wayland protocol, then this is something that will have to be addressed in the Wayland spec before we can support this in Calamares. Your pull request is **not** going to help for **any** distribution other than yours.

kkofler:
As far as I can tell, the proper way to set the live user's keyboard layout under `kwin_wayland` is to write it to the user's `kxkbrc` and then to send the `"org.kde.keyboard" "/Layouts" "reloadConfig"` D-Bus signal (also as that user to the user's session bus, not as root) to get `kwin_wayland` to reload the configuration. Of course, this is completely compositor-specific, any other compositor will have a different way to configure this.

marcan:
@kkofler As far as I know there are *two* major use cases for Calamares: either it is used as a live installer (chroot and all that), or as a first-time setup assistant. The former would often run from some kind of live session; the later will almost always run from some kind of dedicated setup session running as root. The latter use case is what I'm interested in here. That's where I can start kwin with `--layout1` (which is the same thing sddm in Wayland mode does, for what it's worth).

Any distribution that wants to use Calamares as a first-time setup wizard in Wayland mode can do the same and benefit from this. Indeed this will not help the live installer use case, but that's a harder problem because, as you've found out, there is *no* Wayland standard for changing keyboard layouts. If someone wants to implement support for specific compositors that way some day for live session use cases they are free to do that, but I don't see why that should block a proper universal solution for the other major Calamares use case as a first time setup wizard, which *is* possible as I demonstrated here.

adriaandegroot:
I'll fix my nagging in the *calamares* branch directly, no worries. Thanks @marcan

marcan:
Thanks! (I was going to fix the nags but you beat me to it :))

PR #2187, opened 2023-08-20
Update LuksBootKeyFileJob.cpp to check for unencrypted boot with encrypted root

bkmo:
Check to not create luks keyfile with unencrypted boot and encrypted root.

### DIFF
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -242,13 +242,6 @@ hasUnencryptedSeparateBoot()
         { return hasMountPoint( partition, QStringLiteral( "/boot" ) ) && !isEncrypted( partition ); } );
 }
 
-STATICTEST bool
-hasEncryptedRoot()
-{
-    return anyPartition( []( const QVariantMap& partition )
-                         { return hasMountPoint( partition, QStringLiteral( "/" ) ) && isEncrypted( partition ); } );
-}
-
 Calamares::JobResult
 LuksBootKeyFileJob::exec()
 {
@@ -300,7 +293,7 @@ LuksBootKeyFileJob::exec()
         cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
         return Calamares::JobResult::ok();
     }
-    
+
     if ( s.devices.first().passphrase.isEmpty() )
     {
         cDebug() << Logger::SubEntry << "No root passphrase.";
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -294,22 +294,13 @@ LuksBootKeyFileJob::exec()
         return Calamares::JobResult::ok();
     }
 
-    if ( hasUnencryptedSeparateBoot() && !hasEncryptedRoot() )
+    if ( hasUnencryptedSeparateBoot() )
     {
-        // /boot partition is not encrypted, keyfile must not be used
-        // But only if root partition is not encrypted
+        // /boot partition is not encrypted, keyfile must not be used.
         cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
         return Calamares::JobResult::ok();
     }
-
-        if ( hasUnencryptedSeparateBoot() && hasEncryptedRoot() )
-    {
-        // /boot partition is not encrypted, keyfile must not be used
-        // But only if root partition is encrypted
-        cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
-        return Calamares::JobResult::ok();
-    }
-
+    
     if ( s.devices.first().passphrase.isEmpty() )
     {
         cDebug() << Logger::SubEntry << "No root passphrase.";
--- a/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
+++ b/src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
@@ -302,6 +302,14 @@ LuksBootKeyFileJob::exec()
         return Calamares::JobResult::ok();
     }
 
+        if ( hasUnencryptedSeparateBoot() && hasEncryptedRoot() )
+    {
+        // /boot partition is not encrypted, keyfile must not be used
+        // But only if root partition is encrypted
+        cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
+        return Calamares::JobResult::ok();
+    }
+
     if ( s.devices.first().passphrase.isEmpty() )
     {
         cDebug() << Logger::SubEntry << "No root passphrase.";
--- a/src/modules/luksbootkeyfile/Tests.cpp
+++ b/src/modules/luksbootkeyfile/Tests.cpp
@@ -158,7 +158,6 @@ LuksBootKeyFileTests::testAnyPartition()
                             { return hasMountPoint( partdata, QStringLiteral( "/" ) ); } ) );
     QVERIFY( !anyPartition( []( const QVariantMap& partdata ) { return hasMountPoint( partdata, QString() ); } ) );
 
-    QVERIFY( !hasEncryptedRoot() );
     QVERIFY( hasUnencryptedSeparateBoot() );
 }
 


bkmo:
I didn't touch this as it didn't seem to work anyways:
`for ( const auto& d : s.devices )
    {
        // Skip setupLuks for root partition if system has an unencrypted /boot
        if ( d.isRoot && hasUnencryptedSeparateBoot() )
        {
            continue;
        }`

dalto8:
Isn't the real issue in the prior condition?

In other words, this is what is currently in the PR:
```
    if ( hasUnencryptedSeparateBoot() && !hasEncryptedRoot() )
    {
        // /boot partition is not encrypted, keyfile must not be used
        // But only if root partition is not encrypted
        cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
    }

     if ( hasUnencryptedSeparateBoot() && hasEncryptedRoot() )
    {
        // /boot partition is not encrypted, keyfile must not be used
        // But only if root partition is encrypted
        cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
    }
```

But isn't that the same as:
```
    if ( hasUnencryptedSeparateBoot() )
    {
        // /boot partition is not encrypted, keyfile must not be used
        cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
    }
```

bkmo:
> Isn't the real issue in the prior condition?
> 
> In other words, this is what is currently in the PR:
> 
> ```
>     if ( hasUnencryptedSeparateBoot() && !hasEncryptedRoot() )
>     {
>         // /boot partition is not encrypted, keyfile must not be used
>         // But only if root partition is not encrypted
>         cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
>         return Calamares::JobResult::ok();
>     }
> 
>      if ( hasUnencryptedSeparateBoot() && hasEncryptedRoot() )
>     {
>         // /boot partition is not encrypted, keyfile must not be used
>         // But only if root partition is encrypted
>         cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
>         return Calamares::JobResult::ok();
>     }
> ```
> 
> But isn't that the same as:
> 
> ```
>     if ( hasUnencryptedSeparateBoot() )
>     {
>         // /boot partition is not encrypted, keyfile must not be used
>         cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
>         return Calamares::JobResult::ok();
>     }
> ```

It looks to in fact do the same.

bkmo:
So we can do away with the first condition having unencrypted root as it doesn't matter with the unencrypted boot.

dalto8:
> So we can do away with the first condition having unencrypted root as it doesn't matter with the unencrypted boot.

Since they are both basically the same, you can remove either one as long as you also remove the check of encrypted root and the associated comment.

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
```diff
@@ -294,14 +294,13 @@ LuksBootKeyFileJob::exec()
...
+    if ( hasUnencryptedSeparateBoot() )
     {
-        // /boot partition is not encrypted, keyfile must not be used
-        // But only if root partition is not encrypted
+        // /boot partition is not encrypted, keyfile must not be used.
         cDebug() << Logger::SubEntry << "/boot partition is not encrypted, skipping keyfile creation.";
         return Calamares::JobResult::ok();
     }
-
+
```
dalto8: Looks like some extra spaces slipped in here.

bkmo:
> > So we can do away with the first condition having unencrypted root as it doesn't matter with the unencrypted boot.
> 
> Since they are both basically the same, you can remove either one as long as you also remove the check of encrypted root and the associated comment.

Ok, I already made the exact change on my end.

bkmo:
> > > So we can do away with the first condition having unencrypted root as it doesn't matter with the unencrypted boot.
> > 
> > 
> > Since they are both basically the same, you can remove either one as long as you also remove the check of encrypted root and the associated comment.
> 
> Ok, I already made the exact change on my end.

I don't know what you were referring to as also a problem with grubcfg

src/modules/luksbootkeyfile/LuksBootKeyFileJob.cpp
dalto8: You marked this as resolved but it still seems to be present.

dalto8:
> I don't know what you were referring to as also a problem with grubcfg

In my testing, grubcfg is enabling cryptodisk even when `/boot` is unencrypted.

bkmo:
> > I don't know what you were referring to as also a problem with grubcfg
> 
> In my testing, grubcfg is enabling cryptodisk even when `/boot` is unencrypted.

`#GRUB_ENABLE_CRYPTODISK=y`   Was added to a clean install with unencrypted boot.

dalto8:
> > > I don't know what you were referring to as also a problem with grubcfg
> > 
> > 
> > In my testing, grubcfg is enabling cryptodisk even when `/boot` is unencrypted.
> 
> `#GRUB_ENABLE_CRYPTODISK=y` Was added to a clean install with unencrypted boot.

Was the root encrypted in addition to /boot being unencrypted?

bkmo:
> Was the root encrypted in addition to /boot being unencrypted?

Yes the root was encrypted.  Now to look into the creation of cryptab.
Looks to be this in encrypt_hook_nokey in initramfs module:
`. /usr/share/initramfs-tools/hook-functions
  if [ -f /etc/crypttab ]
  then
  	cp /etc/crypttab ${DESTDIR}/etc/
  fi`

So if there is an unencrypted boot then we can change the main.py to not do anything instead of the encrypt_hook_nokey, but I am not sure what to do.
` if unencrypted_separate_boot:
            shutil.copy2(
                os.path.join(_path, "encrypt_hook_nokey"),
                target
                )`

dalto8:
The initramfscfg changes should be moved to a separate PR.  However, I am not confident they are correct.  If you open a separate PR for those, be sure to clearly explain why you made those changes and how you tested them.

bkmo:
> The initramfscfg changes should be moved to a separate PR. However, I am not confident they are correct. If you open a separate PR for those, be sure to clearly explain why you made those changes and how you tested them.

I made those changes to test them, not knowing they would be committed. I am sure they are not correct, but am just playing. Now how do I un-commit them?

dalto8:
> Looks to be this in encrypt_hook_nokey in initramfs

The initramfscfg module is not related.  You wouldn't normally be using that module since you are testing on an Arch-based ISO.

bkmo:
> > Looks to be this in encrypt_hook_nokey in initramfs
> 
> The initramfscfg module is not related. You wouldn't normally be using that module since you are testing on an Arch-based ISO.

It is the only place I encountered crypttab, but now I see what you mean.

dalto8:
> Now how do I un-commit them?

Use `git reset --hard` to return to previous commit and then force push your branch.

bkmo:
> > Now how do I un-commit them?
> 
> Use `git reset --hard` to return to previous commit and then force push your branch.

Ok , but I am working from the web gui. I will clone locally and see how it goes.

bkmo:
No luck with deleting the commits. After a hard reset the force push is everything up to date. I will probably just delete the branch and call it a day.

kkofler:
You are not hard-resetting to the correct commit. You have to specify the commit hash for `git reset --hard`.

bkmo:
> You are not hard-resetting to the correct commit. You have to specify the commit hash for `git reset --hard`.

Thanks, I think I have it sorted out now.

bkmo:
I'm done chasing the crypttab creation on unencrypted boot. The only module I found it in was initramfscfg. At least this seem to take care of the keyfile creation.

bkmo:
This is working well but could use a tweak to get rid of a warning on compile. Maybe you can do it for me please  @dalto8 
` LuksBootKeyFileJob.cpp:246:1: warning: ‘bool hasEncryptedRoot()’ defined but not used [-Wunused-function]
  246 | hasEncryptedRoot()`

dalto8:
That means that function isn't used anymore.  You just need to delete that function.

bkmo:
> That means that function isn't used anymore. You just need to delete that function.

I realize that but when I tried I had an error. I'm not a coder.

dalto8:
You need to give me rights to push to your repo.

bkmo:
I added you as a collaborator.

bkmo:
Thank you, the only other issue is there is an entry to crypptab for the keyfile even with unencrypted boot. It is unnecessary in this scenario.

PR #2197, opened 2023-09-06
Fixed typo in initcpiocfg module

Boria138:
If useSystemdHook is true, then the person wants to use systemd hook respectively the variable systemd_hook_allowed should be 
```libcalamares.job.configuration.get("useSystemdHook", True)```

### DIFF
--- a/src/modules/initcpiocfg/main.py
+++ b/src/modules/initcpiocfg/main.py
@@ -164,7 +164,7 @@ def find_initcpio_features(partitions, root_mount_point):
         "keyboard",
     ]
 
-    systemd_hook_allowed = libcalamares.job.configuration.get("useSystemdHook", False)
+    systemd_hook_allowed = libcalamares.job.configuration.get("useSystemdHook", True)
 
     use_systemd = systemd_hook_allowed and target_env_call(["sh", "-c", "which systemd-cat"]) == 0
 


dalto8:
Isn't that the default value?  So it defaults to false unless you make it true in the config file.

Boria138:
> Isn't that the default value?  So it defaults to false unless you make it true in the config file.

Now if useSystemdHook is true it doesn't use systemd, if false it does, from the name of the setting it should be the other way around

dalto8:
> Now if usesystemdhook is true it doesn't use systemd, if false it does, from the name of the setting it should be the other way around

Unless I am missing something, that isn't the case.

When `useSystemdHook` is set to true in the conf file, it will use the systemd hook.

Boria138:
> > Now if usesystemdhook is true it doesn't use systemd, if false it does, from the name of the setting it should be the other way around
> 
> Unless I am missing something, that isn't the case.
> 
> When `useSystemdHook` is set to true in the conf file, it will use the systemd hook.
> 
> 

It's not, I added in calamares useSystemdHook true, but systemd hook was not used when I changed the code in main.py it all worked

Boria138:
If we parse the code that is written now, we can understand that the systemd hook is added after two conditions are met
1) systemd is present in the system at all
2) systemd_hook_allowed = true
And systemd_hook_allowed in its turn is true only if useSystemdHook is false, while it should be true.

dalto8:
Can I see the conf file you used when you tested it?

Boria138:
> Can I see the conf file you used when you tested it?

https://gitlab.com/LunaOS/iso/lunaos-calamares-config/-/blob/main/etc/calamares/modules/initcpiocfg.conf?ref_type=heads

This is my arch linux based distribution

dalto8:
I will take a look at it tonight and see if it is working or needs changes.

The fix in this PR is probably not the answer though.  It changes the default value.

bkmo:
> I will take a look at it tonight and see if it is working or needs changes.
> 
> The fix in this PR is probably not the answer though. It changes the default value.

I got a udev install when I set useSystemdHook to true also.

dalto8:
I can't test it until later but it is probably because I forgot to remove `noconfig: true` from `module.desc`.

adriaandegroot:
For the OP, in general it's a good idea to use `print()` or debug-logging to check what the configuration dictionary actually contains, before twiddling things. As dalto8 points out, all you are doing is changing the default value used when the option is **not** set at all in the configuration.

I'll quote the module readme:

> Modules that have *noconfig* set to true will not attempt to
read a configuration file, and will not warn that one is missing;

so the noconfig change that just landed is the actually-correct fix. That said, @dalto8  there are some weird typo's and a stray "mkinitcpio" in the config file, if you want to clean up the example. And to avoid this problem in future, we might mess with `CalamaresAddModuleSubdirectory.cmake`, which already reads the descriptor file, and we could check (in the elseif block for a `module.desc`) that there are config files only if noconfig is set to false. The CMake code for C++ modules already does that in `CalamaresAddPlugin.cmake`.

PR #2206, opened 2023-10-02
Correct the comment description of the "Replace" partition behavior

geckolinux:
The described behavior of the "Replace" option was incorrect, it does not keep the same filesystem type that the partition already had, rather it uses the `defaultFileSystemType` value. See: https://github.com/calamares/calamares/issues/1970

### DIFF
--- a/src/modules/partition/partition.conf
+++ b/src/modules/partition/partition.conf
@@ -170,8 +170,8 @@ initialSwapChoice: none
 
 # Default filesystem type, used when a "new" partition is made.
 #
-# When replacing a partition, the existing filesystem inside the
-# partition is retained. In other cases, e.g. Erase and Alongside,
+# When replacing a partition, the new filesystem type will be from the 
+# defaultFileSystemType value. In other cases, e.g. Erase and Alongside,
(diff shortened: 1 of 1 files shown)


dalto8:
I am not sure that comment is fully correct either.

Wasn't replace partition was updated to support `availableFileSystemTypes` in 3.3?

geckolinux:
> Wasn't replace partition was updated to support `availableFileSystemTypes` in 3.3?

So the dropdown selector should appear for Replace Partition? That would be ideal.
The behavior I documented was tested in 3.2.61 from Debian Stable.

dalto8:
> So the dropdown selector should appear for Replace Partition?

Yes, it is available in Erase and Replace now.

adriaandegroot:
@geckolinux A lot has changed in the *calamares* branch since the 3.2 days, so please check against one of the alpha's what it does exactly in the current code. Then update the PR (although this is already better than it **was**). I think it needs a double-check that replace-by-other-filesystem actually works, though.

adriaandegroot:
Adding it anyway since better is better than not-improved-at-all.

PR #2210, opened 2023-10-10

bitigchi:
WIP, strings to be added.

@adriaandegroot, I need more context about the strings in `prettyName()`. Where are they displayed exactly, are they titles some sort of? Other than that, please see if the message format in `prettyStatusMessage()` is alright, I presume this is one of the strings displayed during progress bar, right?

### DIFF
--- a/src/libcalamares/ProcessJob.cpp
+++ b/src/libcalamares/ProcessJob.cpp
@@ -36,16 +36,20 @@ ProcessJob::~ProcessJob() {}
 QString
 ProcessJob::prettyName() const
 {
-    return ( m_runInChroot ? tr( "Run command '%1' in target system" ) : tr( " Run command '%1'" ) ).arg( m_command );
+    return ( m_runInChroot ? "Run command '%1' in target system" : "Run command '%1'" ).arg( m_command );
 }
 
 QString
 ProcessJob::prettyStatusMessage() const
 {
     if ( m_runInChroot )
-        return tr( "Running command %1 in chroot…", "@status" ).arg( m_command );
+    {
+        return tr( "Running command %1 in target system…", "@status" ).arg( m_command );
+    }
     else
+    {
         return tr( "Running command %1…", "@status" ).arg( m_command );
+    }
 }
 
 JobResult
--- a/src/libcalamares/ProcessJob.cpp
+++ b/src/libcalamares/ProcessJob.cpp
@@ -36,13 +36,16 @@ ProcessJob::~ProcessJob() {}
 QString
 ProcessJob::prettyName() const
 {
-    return ( m_runInChroot ? tr( "Run command '%1' in target system." ) : tr( " Run command '%1'." ) ).arg( m_command );
+    return ( m_runInChroot ? tr( "Run command '%1' in target system" ) : tr( " Run command '%1'" ) ).arg( m_command );
 }
 
 QString
 ProcessJob::prettyStatusMessage() const
 {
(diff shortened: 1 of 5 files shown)


src/libcalamares/ProcessJob.cpp
```diff
@@ -36,13 +36,16 @@ ProcessJob::~ProcessJob() {}
...
 }
 
 QString
 ProcessJob::prettyStatusMessage() const
 {
-    return tr( "Running command %1 %2" ).arg( m_command ).arg( m_runInChroot ? "in chroot." : " ." );
+    if ( m_runInChroot )
+        return tr( "Running command %1 in chroot…", "@status" ).arg( m_command );
+    else
+        return tr( "Running command %1…", "@status" ).arg( m_command );
```
adriaandegroot: Please add `{}` around all blocks, even 1-line if- and else-blocks.

```suggestion
    if ( m_runInChroot )
    {
        return tr( "Running command %1 in chroot…", "@status" ).arg( m_command );
    }
    else
    {
        return tr( "Running command %1…", "@status" ).arg( m_command );
    }
```

src/libcalamares/ProcessJob.cpp
```diff
@@ -36,13 +36,16 @@ ProcessJob::~ProcessJob() {}
...
-    return ( m_runInChroot ? tr( "Run command '%1' in target system." ) : tr( " Run command '%1'." ) ).arg( m_command );
+    return ( m_runInChroot ? tr( "Run command '%1' in target system" ) : tr( " Run command '%1'" ) ).arg( m_command );
 }
 
 QString
 ProcessJob::prettyStatusMessage() const
 {
-    return tr( "Running command %1 %2" ).arg( m_command ).arg( m_runInChroot ? "in chroot." : " ." );
+    if ( m_runInChroot )
+        return tr( "Running command %1 in chroot…", "@status" ).arg( m_command );
```
adriaandegroot: I think that "in chroot" is unnecessarily technical and that we should change it to "in target system" like it already says in `prettyName()`

adriaandegroot:
This is a fairly subtle question, actually. There are two functions `prettyName()`:

 - For view modules, `prettyName()` should be just the name of the module; this is displayed in the left-hand-pane, the one that shows how far you are in the installation progress. For example, *Welcome*, *Keyboard*, .. 
 - For jobs, `prettyName()` should state what the job **does**. You have picked a job in this MR. A job's name is almost never shown to the user -- I can find it being used in some debug statements, which is where I would **not** want to see translated text anyway. Maybe the `tr()` calls should go away in that function.

Much simpler is `prettyStatus()`, which is only for jobs, and should say what the job is **doing**. This is shown in the progress bar during installation.

src/libcalamares/modulesystem/RequirementsChecker.cpp
```diff
@@ -120,13 +120,13 @@ RequirementsChecker::reportProgress()
     {
         cDebug() << "Remaining modules:" << remaining << Logger::DebugList( remainingNames );
         unsigned int posInterval = ( m_progressTimer->interval() < 0 ) ? 1000 : uint( m_progressTimer->interval() );
-        QString waiting = tr( "Waiting for %n module(s).", "", remaining );
-        QString elapsed = tr( "(%n second(s))", "", m_progressTimeouts * posInterval / 1000 );
+        QString waiting = tr( "Waiting for %n module(s)…", "@status", remaining );
+        QString elapsed = tr( "(%n second(s))", "@status", m_progressTimeouts * posInterval / 999 );
```
adriaandegroot, line 124: I'll try to fix this to properly pluralize myself after merging this.

PR #2219, opened 2023-10-23
Main

ghost:


adriaandegroot:
I personally don't think it makes sense to snap (or containerize in any way) Calamares, but a decent PR would probably be accepted -- talk to Scarlett Moore who snaps most KDE things.

PR #2222, opened 2023-10-23

bitigchi:
Going into more complicated territory, so nitpicking is welcome!


adriaandegroot:
Now that I understand what you're doing, this is easier to review and my check is basic: "does it read ok?"

PR #2226, opened 2023-10-31

bitigchi:


src/modules/fsresizer/ResizeFSJob.cpp
```diff
@@ -176,20 +176,20 @@ ResizeFSJob::exec()
     if ( !m.first || !m.second )
     {
         return Calamares::JobResult::error(
-            tr( "Resize Failed" ),
+            tr( "Resize Failed", "@error" ),
```
bitigchi: Is capitalisation needed here? Does it look like a title?

src/modules/keyboard/Config.cpp
```diff
@@ -502,14 +502,14 @@ QString
 Config::prettyStatus() const
 {
     QString status;
-    status += tr( "Set keyboard model to %1.<br/>" )
+    status += tr( "Keyboard model has been set to %1<br/>", "@label, %1 is keyboard model, as in Apple Magic Keyboard" )
```
bitigchi: Is this progress or just informative label?

Same question for the label below.

src/modules/keyboardq/keyboardq-qt6.qml
```diff
@@ -58,7 +58,7 @@ Item {
         Label {
             id: header
             anchors.horizontalCenter: parent.horizontalCenter
-            text: qsTr("To activate keyboard preview, select a layout.")
+            text: qsTr("Select a layout to activate keyboard preview", "@label")
```
bitigchi, line 61: Is this a title?

src/modules/locale/Config.cpp
```diff
@@ -370,7 +370,7 @@ Config::setLCLocaleExplicitly( const QString& locale )
 QString
 Config::currentLocationStatus() const
 {
-    return tr( "Set timezone to %1/%2." )
+    return tr( "Timezone has been set to %1/%2", "@info" )
```
bitigchi: Progress or label?

src/modules/locale/SetTimezoneJob.cpp
```diff
@@ -29,7 +29,7 @@ SetTimezoneJob::SetTimezoneJob( const QString& region, const QString& zone )
 QString
 SetTimezoneJob::prettyName() const
 {
-    return tr( "Set timezone to %1/%2" ).arg( m_region ).arg( m_zone );
+    return tr( "Setting timezone to %1/%2…", "@status" ).arg( m_region ).arg( m_zone );
```
bitigchi, line 32: Progress or label?

src/modules/keyboard/Config.cpp
adriaandegroot: I *did* say "`prettyStatus()` is for what a job is doing (e.g. progress)" but some of them do just one thing, and it's quick so the progress message is just "doing the thing". So this is progress, but "has been set" is fine as well because by the time the user sees it at all, it has been done -- but is that consistent with other `prettyStatus()` messages?

src/modules/fsresizer/ResizeFSJob.cpp
adriaandegroot: Via a couple of steps this kind of strings in `JobResult::error()` ends up in a pop-up dialog describing the error. The window title is "Error", there's a "Setup Failed" message at the top of the popup, and then the message is shown as *informative text* in the popup. Here, that message is "Resize Failed". So .. no, it's not a title, it is plain text and should be capitalized as normal text and end with a full stop.

src/modules/keyboardq/keyboardq-qt6.qml
adriaandegroot: No. It appears as an instruction / hint at the top of a window, but it's not a title.

adriaandegroot: image

That one there at the top of the screenshot, partly obscured by the list of physical keyboard layouts.

src/modules/locale/SetTimezoneJob.cpp
adriaandegroot: Progress.

src/modules/locale/Config.cpp
adriaandegroot: This one is a bit unusual. I'll call it "info". This particular string is used in the "summary" page telling the user what is going to happen -- not what has happened. So this particular string change is wrong, it should remain "Set timezone to .." and maybe be tagged `@action` or .. um .. something-that-will-be-done.

adriaandegroot:
Switched this to draft because of the one change to summary string that isn't right. The rest looks pretty solid, though, and I hope I have better explained what the different kinds of strings are doing.

bitigchi:
Thanks for the review! Will fix it in a few days!

src/modules/fsresizer/ResizeFSJob.cpp
bitigchi: Fixed.

src/modules/keyboard/Config.cpp
bitigchi: I think it's okay to let the user know that it's done, with a full stop. If anything, it makes the progress step more accurate.

src/modules/keyboardq/keyboardq-qt6.qml
bitigchi: Thank you.

src/modules/locale/SetTimezoneJob.cpp
bitigchi: Thank you.

bitigchi:
By the way, is there a way to test Calamares with a dummy configuration or something? The latest neon unstable does not seem to include my progress bar patch, but the translations seem updated. When I try to launch Calamares from an install, it does not launch.

PR #2227, opened 2023-11-02
If a mountPoint has subvolume: "" don't try to mount the subvolume with a wrong "subvol=," argument.

jpwhiting:
When testing with a / mountPoint set to subvolume: "" it tried and failed to mount the subvolume:

    .. Running ("mount", "-t", "btrfs", "-o", "subvol=,", "/dev/sda3", "/tmp/calamares-root-ylvhpxys/")
    .. Target cmd: ("mount", "-t", "btrfs", "-o", "subvol=,", "/dev/sda3", "/tmp/calamares-root-ylvhpxys/") Exit code: 32 output:
 mount: /tmp/calamares-root-ylvhpxys: wrong fs type, bad option, bad superblock on /dev/sda3, missing codepage or helper program, or other error.

This change should make it not try since there's no subvolume to mount anyway.
Also don't unmount the volume we are about to mount the subvolume if we aren't really going to mount a subvolume.

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -298,7 +298,12 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
                 mount_option_no_subvol = get_mount_options("btrfs_swap", mount_options, partition)
             else:
                 mount_option_no_subvol = get_mount_options(fstype, mount_options, partition)
-            mount_option = f"subvol={s['subvolume']},{mount_option_no_subvol}"
+
+            # Only add subvol= argument if we are not mounting the entire filesystem
+            if s['subvolume']:
+                mount_option = f"subvol={s['subvolume']},{mount_option_no_subvol}"
+            else:
+                mount_option = mount_option_no_subvol
             subvolume_mountpoint = mount_point[:-1] + s['mountPoint']
             mount_options_list.append({"mountpoint": s['mountPoint'], "option_string": mount_option_no_subvol})
             if libcalamares.utils.mount(device,


dalto8:
If we do that, the btrfs volume won't get mounted at all, right?  

I think that case needs proper handling, not to be ignored.

jpwhiting:
> If we do that, the btrfs volume won't get mounted at all, right?
> 
> I think that case needs proper handling, not to be ignored.

Yeah, as I was falling asleep last night I realized I approached this wrong. I wonder if a better idea would be a btrfsNoSubvolumes: True configuration and if set skip subvolume code completely. Otherwise it's hard to know if mounting/unmounting steps should be taken.

dalto8:
You can have both.  The root mounted on `/` and additional subvolumes mounted underneath that.

adriaandegroot:
I'll wait for another round on this PR then -- got an ISO, Jeremy? That generally makes things so much easier to test.

jpwhiting:
> I'll wait for another round on this PR then -- got an ISO, Jeremy? That generally makes things so much easier to test.

It's not going to work as is for cases where there's a legitimate subvolume. I'll rework it to work for that case also. I don't have an iso handy though, sorry, I'm using this in a test project that will hopefully become public at some point, but isn't just yet. I can share the mount.conf I'm using though if that helps.

jpwhiting:
[mount.txt](https://github.com/calamares/calamares/files/13244456/mount.txt)
@adriaandegroot I'm using this (had to s/monut.conf/mount.txt/ for gitlab to accept it...)

src/modules/mount/main.py
```diff
@@ -294,6 +299,9 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
         # Mount the subvolumes
         swap_subvol = libcalamares.job.configuration.get("btrfsSwapSubvol", "/@swap")
         for s in btrfs_subvolumes:
+            # If the subvolume is "", don't try to mount it.
+            if not s["subvolume"]:
+                continue
             if s['subvolume'] == swap_subvol:
```
dalto8: Looking at it in total, isn't this part of the change all we need?

jpwhiting: I thought that still might be wanted, for any subvolumes where the subvolume is "" (which in practice should only be the / mountPoint, but may change at some point, I'm not sure) we probably don't want to try to mount the subvolume, right?

jpwhiting: That said if it causes confusion I'm happy to remove that bit. It's not strictly needed for my use case.

My use case being just supporting this in mount.conf:

btrfsSubvolumes:
    - mountPoint: /
      subvolume: ""

dalto8: When the subvolume is `""`, it means "The root of the partition is mounted on `/`".  I think that is the only thing this can ever mean because we only process the subvolume logic for the partition mounted at `/`.

However, we have already mounted the partition at `/` so I think we just need to "do nothing" for that subvolume.

dalto8: > That said if it causes confusion I'm happy to remove that bit. It's not strictly needed for my use case.

I think this part is all we need.  It is the other parts that aren't needed.

With this change, the other check is superfluous.

jpwhiting: What I was seeing in my testing here was that with the mount.conf I uploaded before without any changes to calamares sources the / mountPoint was mounted, then the subvolume logic would unmount / then rsync would try to rsync the rootfs to /tmp/calamares-foo which was just part of tmpfs, would fill up and fail. So something needs to drop out of this subvolume code and not do the umount either... right?

To fix that I added the dictionary comparison check at the top. I'll remove this bit in the subvolume mounting though since that's not needed for my use case. One moment.

jpwhiting: I pushed up just doing the config check and dropping out early if it matches. Though I haven't yet tested to verify that is how the above mentioned config ends up in dictionary form. I'll test that shortly.

jpwhiting: > When the subvolume is `""`, it means "The root of the partition is mounted on `/`". I think that is the only thing this can ever mean because we only process the subvolume logic for the partition mounted at `/`.
> 
> However, we have already mounted the partition at `/` so I think we just need to "do nothing" for that subvolume.

Do nothing but also not do the umount command.

src/modules/mount/main.py
```diff
@@ -274,6 +274,11 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
 
         # Store created list in global storage so it can be used in the fstab module
         libcalamares.globalstorage.insert("btrfsSubvolumes", btrfs_subvolumes)
+
+        # Don't do any subvolume mounting or unmounting if we don't have any
+        if btrfs_subvolumes == ["mountPoint": "/", "subvolumes": ""]:
+            break
+
```
dalto8: This code isn't needed.  It is not a general solution.

The code you removed was a general solution I think.

dalto8 (changes requested):

src/modules/mount/main.py
jpwhiting: with this and not the top level dictionary check it will unmount / on line 292. We need a way to avoid that. I'll put this back in, but we need some way to not unmount the / mountPoint when not using subvolumes.

jpwhiting: Ok, added back in the subvolume mounting continue, but something also needs to skip the subprocess umount for my use case.

dalto8: Yes, don't do the umount if there is a subvolume `""`

src/modules/mount/main.py
jpwhiting: I see, I misread your comment as "Do we still need this with the other part"

src/modules/mount/main.py
jpwhiting: sorry, rereading the changes I made and the overal issue. The problem is "just" the "subvol=," argument that gets added when we mount the volume lower here.  There's a umount call above it, which made me think by this point / was already mounted, but needed to be unmounted only for the cases where there's a subvolume so that can get mounted properly. But looking at the other fstype cases we probably need to mount it in here, just without the bad "subvol=," argument. I'll rework my change to try to address that instead.

jpwhiting: Ok, I think what I pushed should fix the problem I hit. I'll test on a usb stick now.

jpwhiting: > Yes, don't do the umount if there is a subvolume `""`

Well, when I had that in place you suggested it wasn't needed. Really it's not, as long as we don't try to remount with "subvol=," it should work to unmount then remount. Testing on an image with my latest changes now. but I think it covers the 1 issue I hit last time and the other issue I hit when trying a fix.

jpwhiting: Yep, code as is is working. It does unmount but then it remounts without the subvol=, argument so it does remount and does have the filesystem to sync into.

src/modules/mount/main.py
```diff
@@ -299,6 +300,10 @@ def mount_partition(root_mount_point, partition, partitions, mount_options, moun
             else:
                 mount_option_no_subvol = get_mount_options(fstype, mount_options, partition)
             mount_option = f"subvol={s['subvolume']},{mount_option_no_subvol}"
+
+            # If the subvolume is "" don't add any subvol= arguments
+            if not s['subvolume']:
+                mount_option = mount_option_no_subvol
```
dalto8:
```suggestion
            # Only add subvol= argument if we are not mounting the entire filesystem
            if s['subvolume']:
                 mount_option = f"subvol={s['subvolume']},{mount_option_no_subvol}"
            else:
                mount_option = mount_option_no_subvol
```

jpwhiting: Yep, makes sense, made that change trying it on my usb stick now.

jpwhiting: Yes this seems to have worked here. Marking resolved.

dalto8 (approved):

PR #2234, opened 2023-11-14
[grubcfg] Don't silently ignore missing grub configuration items

### LINKED ISSUE #2233: [REGRESSION] grubcfg module fails to append missing items to the end of /etc/default/grub
**Describe the bug**
When using the grubcfg module to add items to /etc/default/grub that don't exist in the file yet, the added items are silently ignored. For instance, if doing an installation with encrypted /boot, `GRUB_ENABLE_CRYPTODISK=y` has to be set in /etc/default/grub. However, if the `GRUB_ENABLE_CRYPTODISK` option doesn't existing in the config file by default, attempting to add it with a line like `GRUB_ENABLE_CRYPTODISK: true` in the grubcfg YAML configuration file will fail silently, resulting in an installation failure when Calamares attempts to install the bootloader.

This bug is being observed in Calamares 3.3.0-alpha4 on Lubuntu Noble (24.04 development version), causing this bug in Ubuntu: https://bugs.launchpad.net/ubuntu/+source/calamares/+bug/2043301. However, the file that appears to be the source of the bug appears to be identical in git master and in Calamares 3.3.0-alpha4.

**To Reproduce**
1. Download the latest daily ISO of Lubuntu Noble from https://cdimage.ubuntu.com/lubuntu/daily-live/current/.
2. Launch the ISO in a VM with at least 15 GB of disk space.
3. Open QTerminal in the VM.
4. Run `vim /etc/calamares/modules/grubcfg.conf and verify that the contents are as follows:
```
overwrite: false
defaults:
    GRUB_ENABLE_CRYPTODISK: true
```
5. Run `vim /etc/default/grub` and verify that no `GRUB_ENABLE_CRYPTODISK` option can be seen in this file.
6. Run the installer, enabling the "Encrypt system" feature on the "Partitions" screen.
7. Start the actual installation and wait for it to finish.

**Expected behavior**
The installation should finish on the "All done" screen.

**Actual behavior**
The installation fails with an error indicating that the `grub-install` command failed. If you run `sudo vim /root/.cache/calamares/session.log` in a terminal and search for `grub-install` you will find that GRUB is complaining about the `GRUB_ENABLE_CRYPTODISK` option not being set to `y`.

**Suggested fix**
The bug here appears to be in the function `update_existing_config` in the file https://github.com/calamares/calamares/blob/calamares/src/modules/grubcfg/main.py. The function states that "Items that exist in the file will be updated and new items will be appended to the end". However, only the code that updates the existing contents are present, and there is no code that appends any remaining items to the end. I was able to solve this by opening the corresponding file in the Lubuntu filesystem (`/lib/x86_64-linux-gnu/calamares/modules/grubcfg/main.py`), and making two changes:
1. I changed the `if key in grub_config_items.keys():` block to this:
```
if key in grub_config_items.keys():
    print(f"{key}={grub_config_items[key]}")
    del grub_config_items[key]     # this line is added
```
2. Then at the bottom of the function (below the final `else: print(line)`), I added:
```
if len(grub_config_items) != 0:
    with open(default_grub, "a") as grub_file:
        for dict_key, dict_val in grub_config_items.items():
            grub_file.write(f"{dict_key}={dict_val}")
```
After making these modifications, I was then able to install Lubuntu successfully.

**Additional context**
The commit that introduced this bug appears to be https://github.com/calamares/calamares/commit/a45f6a316e6eafd3f01d6013d88d328b7199fef8.

ArrayBolt3:
Fixes https://github.com/calamares/calamares/issues/2233 using the suggested fix provided in the bug report.

### DIFF
--- a/src/modules/grubcfg/main.py
+++ b/src/modules/grubcfg/main.py
@@ -17,6 +17,7 @@ import libcalamares
 import fileinput
 import os
 import re
+import shutil
 
 import gettext
 _ = gettext.translation("calamares-python",
@@ -95,19 +96,32 @@ def update_existing_config(default_grub, grub_config_items):
     :param default_grub: The absolute path to the grub config file
     :param grub_config_items: A dict holding the key value pairs representing the items
     """
-    for line in fileinput.input(default_grub, inplace=True):
-        line = line.strip()
-        if "=" in line:
-            # This may be a key, strip the leading comment if it has one
-            key = line.lstrip("#").split("=")[0].strip()
+    
+    default_grub_orig = default_grub + ".calamares"
+    shutil.move(default_grub, default_grub_orig)
 
-            # check if this is one of the keys we care about
-            if key in grub_config_items.keys():
-                print(f"{key}={grub_config_items[key]}")
-            else:
-                print(line)
-        else:
-            print(line)
+    with open(default_grub, "w") as grub_file:
+        with open(default_grub_orig, "r") as grub_orig_file:
+            for line in grub_orig_file.readlines():
+                line = line.strip()
+                if "=" in line:
+                    # This may be a key, strip the leading comment if it has one
+                    key = line.lstrip("#").split("=")[0].strip()
+
+                    # check if this is noe of the keys we care about
+                    if key in grub_config_items.keys():
+                        print(f"{key}={grub_config_items[key]}", file=grub_file)
+                        del grub_config_items[key]
+                    else:
+                        print(line, file=grub_file)
+                else:
+                    print(line, file=grub_file)
+
+            if len(grub_config_items) != 0:
+                for dict_key, dict_val in grub_config_items.items():
+                    print(f"{dict_key}={dict_val}", file=grub_file)
+
+    os.remove(default_grub_orig)
 
 
 def modify_grub_default(partitions, root_mount_point, distributor):


dalto8:
I commented more fully in the issue but I believe the current functionality is intended unless  `always_use_defaults` is set to `true`

dalto8:
I think we should consider refactoring here instead of re-opening the file and adding the additional lines.

ArrayBolt3:
That makes sense to me. I'm not quite sure how to do that using fileinput, but I'm happy to attempt a rewrite that doesn't use fileinput and just uses normal file I/O. (I tried to find an "isLastLine" property in fileinput that would allow reliably appending to the file, but it didn't appear to exist. That being said, Python isn't exactly the language I'm the best at, so there may be ways of doing it that I've not thought of.)

dalto8:
I suspect you could detect that you were at the end of the loop and issue multiple print's in that scenario.  However, I have never tried it.  It also seems somewhat inelegant.

Probably best to just use a more traditional approach.

ArrayBolt3:
Alright, I finally got around to rewriting the patch into a more traditional and hopefully more elegant way. I tested it and it works on the latest daily of Lubuntu 24.04. Should be ready for review.

dalto8:
Thanks!

PR #2248, opened 2023-11-26
users: Add support for crypt_gensalt for user passwords

EbonJaeger:
This attempts to locate the presense of the crypt_gensalt function in the crypto library in use. Many distributions have switched to libxcrypt, which provides this function. This means that Calamares can use the native library implementation instead of generating password salts itself, which, depending on the distro's configuration, may be more secure.

If the function can not be found, fallback to the current method of generating password salts.

Tested on Solus, which now uses libxcrypt and the yescrypt function by default.

Signed-off-by: Evan Maddock <maddock.evan@vivaldi.net>

### DIFF
--- a/src/modules/users/CMakeLists.txt
+++ b/src/modules/users/CMakeLists.txt
@@ -6,6 +6,16 @@
 find_package(${qtname} ${QT_VERSION} CONFIG REQUIRED Core DBus Network)
 find_package(Crypt REQUIRED)
 
+# Check for crypt_gensalt
+if(Crypt_FOUND)
+    list(APPEND CMAKE_REQUIRED_LIBRARIES crypt)
+    include(CheckSymbolExists)
+    check_symbol_exists(crypt_gensalt crypt.h HAS_CRYPT_GENSALT)
+    if(HAS_CRYPT_GENSALT)
+        add_definitions(-DHAVE_CRYPT_GENSALT)
+    endif()
+endif()
+
 # Add optional libraries here
 set(USER_EXTRA_LIB
     ${kfname}::CoreAddons
@@ -78,7 +88,9 @@ calamares_add_plugin(users
     SHARED_LIB
 )
 
-calamares_add_test(userspasswordtest SOURCES TestPasswordJob.cpp SetPasswordJob.cpp LIBRARIES ${CRYPT_LIBRARIES})
+if(NOT HAS_CRYPT_GENSALT)
+    calamares_add_test(userspasswordtest SOURCES TestPasswordJob.cpp SetPasswordJob.cpp LIBRARIES ${CRYPT_LIBRARIES})
+endif()
 
 calamares_add_test(
     usersgroupstest
--- a/src/modules/users/CMakeLists.txt
+++ b/src/modules/users/CMakeLists.txt
@@ -8,9 +8,11 @@ find_package(Crypt REQUIRED)
 
 # Check for crypt_gensalt
 if(Crypt_FOUND)
+    set(_old_CRL "${CMAKE_REQUIRED_LIBRARIES}")
     list(APPEND CMAKE_REQUIRED_LIBRARIES crypt)
     include(CheckSymbolExists)
     check_symbol_exists(crypt_gensalt crypt.h HAS_CRYPT_GENSALT)
+    set(CMAKE_REQUIRED_LIBRARIES "${_old_CRL}")
     if(HAS_CRYPT_GENSALT)
         add_definitions(-DHAVE_CRYPT_GENSALT)
     endif()
--- a/src/modules/users/SetPasswordJob.cpp
+++ b/src/modules/users/SetPasswordJob.cpp
@@ -44,6 +44,7 @@ SetPasswordJob::prettyStatusMessage() const
     return tr( "Setting password for user %1." ).arg( m_userName );
 }
 
+#ifndef HAVE_CRYPT_GENSALT
 /// Returns a modular hashing salt for method 6 (SHA512) with a 16 character random salt.
 QString
 SetPasswordJob::make_salt( int length )
@@ -67,6 +68,7 @@ SetPasswordJob::make_salt( int length )
     salt_string.append( '$' );
     return salt_string;
 }
+#endif
 
 Calamares::JobResult
 SetPasswordJob::exec()
@@ -90,7 +92,14 @@ SetPasswordJob::exec()
         return Calamares::JobResult::ok();
     }
 
-    QString encrypted = QString::fromLatin1( crypt( m_newPassword.toUtf8(), make_salt( 16 ).toUtf8() ) );
+    QString salt;
+#ifdef HAVE_CRYPT_GENSALT
+    salt = crypt_gensalt( NULL, 0, NULL, 0 );
+#else
+    salt = make_salt( 16 );
+#endif
+
+    QString encrypted = QString::fromLatin1( crypt( m_newPassword.toUtf8(), salt.toUtf8() ) );
 
     int ec = Calamares::System::instance()->targetEnvCall( { "usermod", "-p", encrypted, m_userName } );
     if ( ec )
--- a/src/modules/users/SetPasswordJob.h
+++ b/src/modules/users/SetPasswordJob.h
@@ -22,8 +22,9 @@ public:
     QString prettyName() const override;
     QString prettyStatusMessage() const override;
     Calamares::JobResult exec() override;
-
+#ifndef HAVE_CRYPT_GENSALT
     static QString make_salt( int length );
+#endif /* HAVE_CRYPT_GENSALT */
 
 private:
     QString m_userName;


src/modules/users/CMakeLists.txt
```diff
@@ -6,6 +6,16 @@
 find_package(${qtname} ${QT_VERSION} CONFIG REQUIRED Core DBus Network)
 find_package(Crypt REQUIRED)
 
+# Check for crypt_gensalt
+if(Crypt_FOUND)
+    list(APPEND CMAKE_REQUIRED_LIBRARIES crypt)
```
adriaandegroot, line 12: You'll want to undo this after the call to `check_symbol_exists()`, so first, make a copy:
```suggestion
    set(_old_CRL "${CMAKE_REQUIRED_LIBRARIES}")
    list(APPEND CMAKE_REQUIRED_LIBRARIES crypt)
```

src/modules/users/CMakeLists.txt
```diff
@@ -6,6 +6,16 @@
...
+    include(CheckSymbolExists)
+    check_symbol_exists(crypt_gensalt crypt.h HAS_CRYPT_GENSALT)
```
adriaandegroot, line 14: And undo afterwards:
```suggestion
    check_symbol_exists(crypt_gensalt crypt.h HAS_CRYPT_GENSALT)
    set(CMAKE_REQUIRED_LIBRARIES "${_old_CRL}")
```

src/modules/users/CMakeLists.txt
```diff
@@ -6,6 +6,16 @@
...
+    if(HAS_CRYPT_GENSALT)
+        add_definitions(-DHAVE_CRYPT_GENSALT)
+    endif()
```
adriaandegroot, line 18: Do this later, when you have a target to add the definitions to. Also, you don't need `-D` when adding definitions. CMake is smart enough to know how to add a definition for the compiler being used. So just drop these lines

adriaandegroot: Then down below, add lines
```
if(Crypt_FOUND AND HAS_CRYPT_GENSALT)
target_compile_definitions(users_internal PRIVATE HAVE_CRYPT_GENSALT)
endif()
```

(Note that the rest of the CMake code shows all of the bad habits that I'm telling you not to follow .. don't use it as an example, really)

src/modules/users/SetPasswordJob.cpp
```diff
@@ -44,6 +44,7 @@ SetPasswordJob::prettyStatusMessage() const
     return tr( "Setting password for user %1." ).arg( m_userName );
 }
 
+#ifndef HAVE_CRYPT_GENSALT
```
adriaandegroot, line 47: I would ifdef the implementation, myself: leave the API intact, leave the callers unchanged, but
- fix the misleading comment (it uses `length`, not 16)
- change the comment into a full Doxygen `/** */` comment explaining that ... but if this other thing ...
- internally, use `#ifdef HAVE_CRYPT_GENSALT` to return `crypt_gensalt( NULL, 0, NULL, 0 )` 

Is there no way to specify a length for that function? Does it just always return 16 bytes, or what? 

Also, prefer `nullptr` over `NULL`.

src/modules/users/SetPasswordJob.cpp
```diff
@@ -90,7 +92,14 @@ SetPasswordJob::exec()
         return Calamares::JobResult::ok();
     }
 
-    QString encrypted = QString::fromLatin1( crypt( m_newPassword.toUtf8(), make_salt( 16 ).toUtf8() ) );
```
adriaandegroot, line 93: Leave this alone by replacing the implementation of `make_salt()` instead. It could be a `const QString` though.

src/modules/users/CMakeLists.txt
```diff
@@ -78,7 +88,9 @@ calamares_add_plugin(users
     SHARED_LIB
 )
 
-calamares_add_test(userspasswordtest SOURCES TestPasswordJob.cpp SetPasswordJob.cpp LIBRARIES ${CRYPT_LIBRARIES})
+if(NOT HAS_CRYPT_GENSALT)
+    calamares_add_test(userspasswordtest SOURCES TestPasswordJob.cpp SetPasswordJob.cpp LIBRARIES ${CRYPT_LIBRARIES})
```
adriaandegroot, line 94: What makes the test unsuitable if this function is available?

If it's about testing internal functions, see comments below about how I would prefer keeping `make_salt()` around.

adriaandegroot (commented):
As a whole, looks good -- I'll trust you on the operation of this extra function in libcrypt. Take the review comments as "this could be better", but I realise that they're also all stylistic, not functional. I could chase them down in a post-merge fix, too.

src/modules/users/CMakeLists.txt
EbonJaeger: > Note that the rest of the CMake code shows all of the bad habits that I'm telling you not to follow .. don't use it as an example, really

Ah, see, that's exactly what I did, being unfamiliar with CMake :joy:

src/modules/users/CMakeLists.txt
EbonJaeger: I suppose the question becomes how to test it, or rather, how to make it reproducible. When a null pointer is passed to the `prefix` argument of `crypt_gensalt()`, it will pick the best available hashing method, which could potentially change from system to system (this is probably unlikely), or simply over time as newer, better methods are discovered.

src/modules/users/CMakeLists.txt
EbonJaeger: So to make sure I have this right, the whole thing should look like:

```cmake
# Check for crypt_gensalt
if(Crypt_FOUND)
    set(_old_CRL "${CMAKE_REQUIRED_LIBRARIES}")
    list(APPEND CMAKE_REQUIRED_LIBRARIES crypt)
    include(CheckSymbolExists)
    check_symbol_exists(crypt_gensalt crypt.h HAS_CRYPT_GENSALT)
    set(CMAKE_REQUIRED_LIBRARIES "${_old_CRL}")
endif()

if(Crypt_FOUND AND HAS_CRYPT_GENSALT)
    target_compile_definitions(users_internal PRIVATE HAVE_CRYPT_GENSALT)
endif()
```

src/modules/users/SetPasswordJob.cpp
ReillyBrogan: > Is there no way to specify a length for that function? Does it just always return 16 bytes, or what?

So I think this question comes from a misunderstanding of what the `crypt()` function actually is doing. It takes two arguments, `prefix`, and `password`, and returns a hash which is guaranteed to be identical to output of another call of the function with the same prefix and password. 

The prefix is composed of multiple parts. Previously, we were manually construction a prefix that would look like this: `$6$GkZUpcVNX.ErIfwo`. This breaks down to `$6` as specifying the `sha512crypt` algorithm, and then `GkZUpcVNX.ErIfwo` as the randomly generated salt to use. The `sha512crypt` algorithm only accepts two inputs, the salt and the password so this prefix was sufficient for the `crypt()` function.

What `crypt_gensalt( NULL, 0, NULL, 0 )` is doing instead, is asking libxcrypt to return a cryptographically generated prefix of whatever algorithm libxcrypt considers strongest of the algorithms it was compiled with and the default settings for that algorithm. libxcrypt is almost always compiled with the `--enable-hashes=strong` option, which means that it will consider yescrypt to be the strongest algorithm it has available. 

On such a system an example prefix returned by `crypt_gensalt` would be `$y$j9T$ixuQH9rz2Bfh9yWXs0BJ2/`. This breaks down to `$y` as the identifier for the yescrypt algorithm, `j9T` as parameters of the yescrypt algorithm (the breakdown of what `j9T` mean is a bit outside the scope of this response, but you can read up on it [here](https://unix.stackexchange.com/questions/690679/what-does-j9t-mean-in-yescrypt-from-etc-shadow) if you want). Then `ixuQH9rz2Bfh9yWXs0BJ2/` is a randomly generated salt suitable for cryptographic use. 

It is possible to change the arguments of the `crypt_gensalt` function to get a specific algorithm or salt of a specified length out of it, but it's not really necessary (or even desirable) in this case. The entire point of switching to `crypt_gensalt` is to pass the decision of what algorithm and other default settings to the distro so they can use whatever they feel best (via patches or compile-time build parameters). Solus itself is specifically seeking this feature because we've migrated to yescrypt as the default password hashing algorithm and would like that to be reflected by hashes created by the installer as well.

Some background information on this from the related Fedora change: https://fedoraproject.org/wiki/Changes/yescrypt_as_default_hashing_method_for_shadow

src/modules/users/CMakeLists.txt
ReillyBrogan: I assume the CI system has libxcrypt if it's at all using a recent distribution. Considering that `crypt_gensalt` is supposed to return random output I'm not sure what the best way to wire this up for tests is. I can only think of creating a static yescrypt or sha512crypt prefix and using that to determine if the password hashing function returns expected output?

src/modules/users/CMakeLists.txt
adriaandegroot: Just right! The second `if()` goes lower in the file, because it refers to the target *users_internal*, which must exist by then.

src/modules/users/CMakeLists.txt
adriaandegroot: Aha, so we can check "does it return the thing" but since it is supposed to return the **best** thing at any given time, that thing may change and we'd end up with broken-by-external-changes tests. Not what we want. 

Thanks for the explanation.

src/modules/users/SetPasswordJob.cpp
adriaandegroot: Thanks for the explanation.

adriaandegroot (approved):
LGTM, I'll merge it and then repair any fallout in CI.

src/modules/users/CMakeLists.txt
EbonJaeger: I think I understand. Thanks!

EbonJaeger:
Did you still want me to change the #ifdefs to inside the body of the function?

PR #2249, opened 2023-11-26
bootloader: Add clr-boot-manager support

EbonJaeger:
This adds support for Clear Boot Manager to handle the bootloader installation and configuration.

When this option is selected in the bootloader.conf, clr-boot-manager will be used to install the bootloader (systemd-boot on EFI systems). If the system is non-EFI, Grub must be installed first, because clr-boot-manager doesn't do that, despite it running grub_mkconfig after.

Tested on Solus, which uses clr-boot-manager for bootloader management.

Signed-off-by: Evan Maddock <maddock.evan@vivaldi.net>

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -502,6 +502,35 @@ def get_kernels(installation_root_path):
     return kernel_list
 
 
+def install_clr_boot_manager():
+    """
+    Installs clr-boot-manager as the bootloader for EFI systems
+    """
+    libcalamares.utils.debug("Bootloader: clr-boot-manager")
+
+    installation_root_path = libcalamares.globalstorage.value("rootMountPoint")
+    kernel_dir = os.path.join(installation_root_path, "etc", "kernel", "cmdline.d")
+    kernel_resume_file = os.path.join(kernel_dir, "10_resume.conf")
+
+    partitions = libcalamares.globalstorage.value("partitions")
+    swap_uuid = ""
+
+    # Get the UUID of the swap partition, if present
+    for partition in partitions:
+        if partition["fs"] == "linuxswap" and not partition.get("claimed", None):
+            continue
+        has_luks = "luksMapperName" in partition
+        if partition["fs"] == "linuxswap" and not has_luks:
+            swap_uuid = partition["uuid"]
+
+    # Write out the resume.conf for clr-boot-manager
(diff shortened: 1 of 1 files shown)


src/modules/bootloader/main.py
```diff
@@ -855,7 +884,12 @@ def prepare_bootloader(fw_type):
 
     efi_directory = libcalamares.globalstorage.value("efiSystemPartition")
 
-    if efi_boot_loader == "systemd-boot" and fw_type == "efi":
+    if efi_boot_loader == "clr-boot-manager":
+        if fw_type != "efi":
+            # Grub has to be installed first on non-EFI systems
+            install_grub(efi_directory, fw_type)
```
dalto8, line 883: We shouldn't need this since we already install `grub` when it isn't EFI at line 899.

dalto8 (changes requested):

src/modules/bootloader/main.py
EbonJaeger: The `elif` chain won't fall through to that if `clr-boot-manager` is set as the bootloader in `bootloader.conf`, will it? Even if it does, CBM still has to be invoked after installing Grub. CBM is weird in that it will install systemd-boot for you, but it won't install Grub when it runs.

Clear Boot Manager manages both systemd-boot *and* Grub, it just for some reason only installs one of them.

src/modules/bootloader/main.py
```diff
@@ -502,6 +502,35 @@ def get_kernels(installation_root_path):
...
+    partitions = libcalamares.globalstorage.value("partitions")
+    swap_uuid = ""
+
+    # Get the UUID of the swap partition, if present
+    for partition in partitions:
+        if partition["fs"] == "linuxswap" and not partition.get("claimed", None):
+            continue
+        has_luks = "luksMapperName" in partition
+        if partition["fs"] == "linuxswap" and not has_luks:
+            swap_uuid = partition["uuid"]
```
dalto8: Shouldn't we call `get_kernel_params()` here instead?  That would build a set of params that would work for many different scenarios and also honor the conf file options.

EbonJaeger: It would probably be easier and better, yeah. This is the only thing that the old Solus installer sets, so I just stuck with that for a start, heh.

dalto8 (approved):

PR #2250, opened 2023-11-27
Update CHANGES-3.3

fhdk:
added contribution

### DIFF
--- a/CHANGES-3.3
+++ b/CHANGES-3.3
@@ -13,11 +13,12 @@ the history of the 3.2 series (2018-05 - 2022-08).
 This release contains contributions from (alphabetically by first name):
  - Adriaan de Groot
  - Christophe Marin
+ - Frede Hundewadt
 
 ## Core ##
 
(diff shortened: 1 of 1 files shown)


CHANGES-3.3
```diff
@@ -13,11 +13,12 @@ the history of the 3.2 series (2018-05 - 2022-08).
 This release contains contributions from (alphabetically by first name):
  - Adriaan de Groot
  - Christophe Marin
+ - Frede Hundewadt
 
 ## Core ##
 
 ## Modules ##
-
+initcpiocfg - modernize python usd change bash array to ()
```
adriaandegroot: Please follow the style of existing items when describing a change to a module, and keep the two-blank-lines before the older releases. In any case, this doesn't affect the user (distro, that is) of the module, so I wouldn't mention it specifically here. Just add your name (like you did in line 16) and leave it at that.

fhdk: check

PR #2252, opened 2023-12-08
Add support for systemd-machine-id variations

### LINKED ISSUE #2225: Improvement: Option to set `machine-id` to `uninitialized`
Hi, I use Calamares 3.2.61 on a set of spins built from Debian 12.

I need a custom systemd service to run on first boot using `ConditionFirstBoot=true`. The systemd logic for determining whether it's a first boot is rather weird:
https://www.freedesktop.org/software/systemd/man/latest/machine-id.html#First%20Boot%20Semantics

The system that Calamares installs has a blank `/etc/machine-id` file, I'm not sure where it comes from, but Calamares by default won't touch it whether the `machineid` module is enabled or not, and it also won't modify it if `systemd: false` is set in `machineid.conf` . When the installed system boots for the first time, systemd generates the ID and adds it to `/etc/machine-id`, which is good, but unfortunately the logic it uses says that it's _not_ the the first boot since the `/etc/machine-id` file is present but blank, and therefore my `ConditionFirstBoot=true` service doesn't run.

So it would be handy if there was a `machineid.conf` option for `systemd: uninitialized` to make Calamares fill `/etc/machine-id` with the string `uninitialized` , which would make systemd detect the first boot and also later automatically fill in a random machine ID.

adriaandegroot:
Closes #2225 . Allows blank, literal "uninitialized" and existing practice of writing a machine-id into the target system.

### DIFF
--- a/src/libcalamares/Job.h
+++ b/src/libcalamares/Job.h
@@ -76,7 +76,7 @@ public:
      * Pass in a suitable error code; using 0 (which would normally mean "ok") instead
      * gives you a GenericError code.
      */
-    static JobResult internalError( const QString&, const QString& details, int errorCode );
+    static JobResult internalError( const QString& message, const QString& details, int errorCode );
 
 protected:
     explicit JobResult( const QString& message, const QString& details, int errorCode );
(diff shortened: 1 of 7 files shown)


dalto8 (approved):

Conan-Kudo (commented):
LGTM.

PR #2253, opened 2023-12-09
Fix mount defaults

Conan-Kudo:
I noticed that this defaults were a bit wrong, so here are fixes with details in each commit.

### DIFF
--- a/src/modules/mount/mount.conf
+++ b/src/modules/mount/mount.conf
@@ -116,7 +116,7 @@ mountOptions:
     - filesystem: efi
       options: [ defaults, umask=0077 ]
     - filesystem: btrfs
-      options: [ defaults, compress=lzo ]
+      options: [ defaults, compress=zstd:1 ]
     - filesystem: btrfs_swap
       options: [ defaults, noatime ]
 
--- a/src/modules/mount/mount.conf
+++ b/src/modules/mount/mount.conf
@@ -95,9 +95,9 @@ btrfsSwapSubvol: /@swap
 #    - filesystem: efi
 #      options: [ defaults, umask=0077 ]
 #    - filesystem: ext4
-#      options: [ defaults, noatime ]
+#      options: [ defaults ]
 #    - filesystem: btrfs
-#      options: [ defaults, noatime, compress=zstd:1 ]
+#      options: [ defaults, compress=zstd:1 ]
 #      ssdOptions: [ discard=async ]
 #      hddOptions: [ autodefrag ]
 #    - filesystem: btrfs_swap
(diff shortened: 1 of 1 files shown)


dalto8:
Originally, we didn't switch to zstd because some distros didn't have support for it.  That should be a non-issue at this point.

I think both these changes are good.

What do you think about removing some of the superfluous `defaults` as well?

Conan-Kudo:
> What do you think about removing some of the superfluous defaults as well?

My understanding is that this gets directly written into `fstab(5)`, right? I think `defaults` needs to exist when there are no other options because of `fstab(5)` formatting requirements. It doesn't hurt anything to be there for everything, so ehh?

Conan-Kudo:
To be clear, the result of these changes aligns a lot more to what Fedora and openSUSE do today. Pretty much the only thing from downstream Fedora I'm not propagating up to Calamares is the adjustments to the default subvolume arrangement, mostly because I'm not sure I should.

dalto8:
> the only thing from downstream Fedora I'm not propagating up to Calamares is the adjustments to the default subvolume arrangement, mostly because I'm not sure I should.

Yeah, probably not.  I think many distros try to be compatible with Timeshift, unfortunately.

dalto8 (approved):

src/modules/mount/mount.conf
```diff
@@ -108,15 +108,15 @@ btrfsSwapSubvol: /@swap
...
     - filesystem: default
-      options: [ defaults, noatime ]
+      options: [ defaults ]
     - filesystem: efi
       options: [ defaults, umask=0077 ]
     - filesystem: btrfs
-      options: [ defaults, noatime, compress=lzo ]
+      options: [ defaults, compress=zstd:1 ]
     - filesystem: btrfs_swap
       options: [ defaults, noatime ]
```
davide125, line 121: I'm pretty sure `noatime` is meaningless for swap, as it's not actually a filesystem

dalto8: btrfs_swap isn't a swap partition.  It is the subvolume that holds the btrfs swapfile.

That being said, most mount options aren't subvolume specific.  I am not sure about `noatime` though.

adriaandegroot:
Landing ahead of 3.3.0 -- and distro's can aways patch the example config anyway.

ghost:
@dalto8 @davide125  Keeping noatime might have been a good idea:

https://github.com/kdave/btrfs-progs/issues/377

Conan-Kudo:
It is up to a distribution to make this choice, not Calamares.

ghost:
@Conan-Kudo And so they do, these are just suggested defaults. It is strange this reaction in your request where you promote your preferences, with "fix" in the title 🙄

Anyway, I'm not even dev and I didn't come here to argue. I'm just making a vague suggestion and providing info that might be relevant.

Conan-Kudo:
This was a recommendation from the upstream Btrfs developers. I'm aware of that discussion, but generally they would prefer that `noatime` is only used if the user has specifically asked for it, since it breaks a lot of tools if `atimes` aren't updated (such as filesystem indexers).

PR #2260, opened 2023-12-27
Improve context

bitigchi:


bitigchi:
I think this concludes the string context improvements. Please let me know if there are any sections I've missed.

Please bear with me on this one, as my mind stopped reading contexts after a while. There might be repetitive, unnecessary or missed lines.

Thanks in advance!

src/modules/partition/jobs/SetPartitionFlagsJob.cpp
```diff
@@ -62,38 +62,38 @@ SetPartFlagsJob::prettyDescription() const
...
         }
-        return tr( "Clear flags on new partition." );
+        return tr( "Clearing flags on new partition…", "@status" );
     }
 
     if ( !partition()->partitionPath().isEmpty() )
     {
-        return tr( "Flag partition <strong>%1</strong> as "
-                   "<strong>%2</strong>." )
+        return tr( "Flagging partition <strong>%1</strong> as "
```
dalto8, line 80: "Flagging" seems like a strange word to use in this context.

If we want to change it like that perhaps it should be `Setting flags on partition"

src/modules/umount/UmountJob.cpp
```diff
@@ -36,7 +36,7 @@ UmountJob::~UmountJob() {}
 QString
 UmountJob::prettyName() const
 {
-    return tr( "Unmount file systems." );
+    return tr( "Unmounting file systems…", "@status" );
```
dalto8, line 39: Is changing this to "Unmounting" always correct in all ways that `prettyName()` is used?

adriaandegroot: Since this is a `Job` subclass, the aPI documentation in `Job.h` applies:

- `prettyName` is a human-readable name/description of what the job does (not what it is **doing**),
- `prettyDescription` is a longer humand-readable description used in the summary page
- `prettyStatusMessage` is a human-readable description of what the job is **doing** and can contain numbers / update information

There is some overlap between them, and some fallback mechanisms so you don't necessarily have to write all three (in particular, the description is rarely used because it's for things-that-will-be-show-in-the-summary-page).

This will look a bit weird in the logs, because the name is used to list which jobs will run. This will look less weird during installation, because this particular job does not have a description or status, so its name gets used instead. This PR would be better if it added `prettyStatusMessage()` and used this status-style message for that, leaving the original name unchanged.

src/modules/partition/jobs/SetPartitionFlagsJob.cpp
adriaandegroot: This string needs to be taken in context, it is in the `prettyDescription()` function, so this is a longer human-readable description of what the job will do during installation. So "flagging" is not right. "Set flag %2 .." is better and matches the "clear flag" description it can also give. The job name is "Set flags on partition" and the status messages are "Clearing ..." and "Setting ..."

adriaandegroot:
Merging, and I'll fix up the comments in the *calamares* branch myself.

PR #2273, opened 2024-01-19
Encourage, not enforce, following NIST password guidelines

es20490446e:
Encourages the user to follow [NIST password guidelines](https://blog.netwrix.com/2022/11/14/nist-password-guidelines/).

But allows to ignore the guidelines, by unchecking the box "use secure password".

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@ Calamares maintainers through the years:
 Calamares has received contributions of code, documentation, artwork
 and moral support from (alphabetically by first name or nickname):
 
+ - Alberto Salvia Novella
  - Alf Gaida
  - aliveafter1000
  - Allen Welkie
@@ -73,5 +74,3 @@ and moral support from (alphabetically by first name or nickname):
  - vtriolet
  - Walter Lapchynski
  - Waneon Kim
-
- > This list was updated to revision 6e8d820737dea0f3e08f12b10768facef19be684 on May 28th 2022.
--- a/AUTHORS
+++ b/AUTHORS
@@ -74,3 +74,5 @@ and moral support from (alphabetically by first name or nickname):
  - vtriolet
  - Walter Lapchynski
  - Waneon Kim
+
+ > This list was updated to revision 6e8d820737dea0f3e08f12b10768facef19be684 on May 28th 2022.
--- a/src/modules/users/users.conf
+++ b/src/modules/users/users.conf
@@ -145,7 +145,8 @@ doAutologin: false
 #  - set 'allowWeakPasswords' to false (the default)
 #  - configure password-checking, e.g. with NIST settings
 #
-# The defaults follow NIST-800-63B guidelines.
+# The defaults follow NIST password guidelines:
+# (https://pages.nist.gov/800-63-3/sp800-63b.html)
 
 passwordRequirements:
     minLength: 8
--- a/src/modules/users/users.conf
+++ b/src/modules/users/users.conf
@@ -148,6 +148,22 @@ doAutologin: false
 #
 # The defaults follow NIST password guidelines:
 # (https://pages.nist.gov/800-63-3/sp800-63b.html)
+#
+# And are:
+# passwordRequirements:
+#     minLength: 8
+#     maxLength: 64
+#     libpwquality:
+#         - minlen=6
+#         - maxrepeat=3
+#         - maxsequence=3
+#         - usersubstr=4
+#         - badwords=user password linux
+#
+# For no password requirements:
+# passwordRequirements:
+#     minLength: -1
+#     maxLength: -1
 
 passwordRequirements:
     minLength: 8
--- a/src/modules/users/users.conf
+++ b/src/modules/users/users.conf
@@ -106,7 +106,7 @@ doReusePassword: true
 #  - true to check or
 #  - false to uncheck
 # These set the **initial** state of the checkbox.
-doAutologin:     true
+doAutologin: false
 
 # These are optional password-requirements that a distro can enforce
 # on the user. The values given in this sample file set only very weak
@@ -115,8 +115,7 @@ doAutologin:     true
 # Calamares itself supports two checks:
 #   - minLength
 #   - maxLength
-# In this sample file, the values are set to -1 which means "no
-# minimum", "no maximum". This allows any password at all.
+# A value of -1 disables the check.
 # No effort is done to ensure that the checks are consistent
 # (e.g. specifying a maximum length less than the minimum length
 # will annoy users).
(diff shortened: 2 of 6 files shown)


es20490446e:
I don't know if the `modules/usersq` is also relevant here...

demmm:
> I don't know if the `modules/usersq` is also relevant here...

In general, the `q` modules use the same .conf as their widgets counterpart, settings are included, mostly no repetition of instructions, so yes, it is relevant for this adjustment.

es20490446e:
Let me change that too...

AUTHORS
```diff
@@ -73,5 +74,3 @@ and moral support from (alphabetically by first name or nickname):
  - vtriolet
  - Walter Lapchynski
  - Waneon Kim
-
- > This list was updated to revision 6e8d820737dea0f3e08f12b10768facef19be684 on May 28th 2022.
```
adriaandegroot: I would prefer doing this in a separate merge, when it's actually a complete update (e.g. removes BlueSystems mention, adds all authors since that commit, updates the stamp).

src/modules/users/users.conf
```diff
@@ -145,47 +144,27 @@ doAutologin:     true
 # To require specific password characteristics:
 #  - set 'allowWeakPasswords' to false (the default)
 #  - configure password-checking, e.g. with NIST settings
+#
+# The defaults follow NIST-800-63B guidelines.
```
adriaandegroot: Please keep the link to this document (which you removed lower down) because NIST doesn't necessarily mean anything to distro-makers.

src/modules/users/users.conf
```diff
@@ -115,8 +115,7 @@ doAutologin:     true
 # Calamares itself supports two checks:
 #   - minLength
 #   - maxLength
-# In this sample file, the values are set to -1 which means "no
-# minimum", "no maximum". This allows any password at all.
+# A value of -1 disables the check.
```
adriaandegroot: Probably 'length check" for good measure.

src/modules/users/users.conf
```diff
@@ -145,47 +144,27 @@ doAutologin:     true
...
 #  - set 'allowWeakPasswords' to false (the default)
 #  - configure password-checking, e.g. with NIST settings
+#
+# The defaults follow NIST-800-63B guidelines.
 
-
-# These are very weak -- actually, none at all -- requirements
 passwordRequirements:
-    minLength: -1  # Password at least this many characters
-    maxLength: -1  # Password at most this many characters
```
adriaandegroot: At the very least this should remain as an example, for instance commented-out.

src/modules/users/users.conf
```diff
@@ -145,47 +144,27 @@ doAutologin:     true
...
+    maxLength: 64
     libpwquality:
-        - minlen=0
-        - minclass=0
-
-# These are "you must have a password, any password" -- requirements
-#
-# passwordRequirements:
-#     minLength: 1
-
```
adriaandegroot: This example should stay as well.

src/modules/users/users.conf
```diff
@@ -145,47 +144,27 @@ doAutologin:     true
...
-# These are requirements the try to follow the suggestions from
-#     https://pages.nist.gov/800-63-3/sp800-63b.html , "Digital Identity Guidelines".
-# Note that requiring long and complex passwords has its own cost,
-# because the user has to come up with one at install time.
```
adriaandegroot, line 166: We can lose this editorializing, but then we need to have documentation explaining the choices distro's can make, e.g. "the uncommented settings ask for a strong password and disable autologin, but the user can explicitly change that to enable weak(er) passwords and/or autologin." Again, with examples.

src/modules/users/users.conf
```diff
@@ -145,47 +144,27 @@ doAutologin:     true
...
+        - maxrepeat=3
+        - maxsequence=3
+        - usersubstr=4
+        - badwords=user password linux
 
 # You can control the visibility of the 'strong passwords' checkbox here.
 # Possible values are:
-#  - true to show or
-#  - false to hide  (default)
+#  - true to show  (default)
```
adriaandegroot, line 184: Then you need to change the default, too. That's in the C++ code.

adriaandegroot:
The *usersq* configuration should be roughly the same as the *users* configuration, only *usersq* is stripped of documentation because it points to the other file. (At some point in the future, I see the Widgets and QML modules merging into one, but that's a 3.4 timespan thing).

But overall, these example configurations are exactly that: examples. I have long maintained that distro's should be writing their own configurations in a separate repository and should not use `INSTALL_CONFIG` -- but some patch the config files as part of package building. This kind of change will annoy the patchers and have no effect on the separate-config-repo. (It may be **good** to annoy the patchers every now and then, as a warning that the config has evolved).

Finally, this would need an entry in the `CHANGES-3.3` file, under *modules*.

AUTHORS
es20490446e: Fixed 👌

src/modules/users/users.conf
es20490446e: Fixed at: https://github.com/es20490446e/calamares/commit/07fb4a9d6348a8ff696ca08529b576a1d6da7588

src/modules/users/users.conf
es20490446e: Fixed at: https://github.com/es20490446e/calamares/commit/7c823f0c911679fff4393004dead44046c751744

src/modules/users/users.conf
es20490446e: Fixed at: https://github.com/es20490446e/calamares/commit/749a8a47de82162f1cbd4eb0aced1592f8ccee72

src/modules/users/users.conf
es20490446e: Fixed at: https://github.com/es20490446e/calamares/commit/77b29e05b6ad8b7e74dd73af98ce0d622b15f77a

src/modules/users/users.conf
es20490446e: https://github.com/es20490446e/calamares/commit/5c115b4c0d43886a68547fc4adb95a165bb3eb1a

src/modules/users/users.conf
```diff
@@ -145,47 +145,48 @@ doAutologin:     true
...
+#
+# For accepting any password:
+# passwordRequirements:
+#     minLength: 1
+
+passwordRequirements:
+    minLength: 8
+    maxLength: 64
+    libpwquality:
+        - minlen=6
```
dalto8, line 176: It seems a bit strange to set this differently than minLength.

es20490446e: With `minlen=6` it will only accept passwords of length 8, I have checked it out.

If I increase that number, it asks for a longer password than that.

es20490446e: Because:

> minlen=X
> The minimum acceptable size for the new password (plus one if credits are not disabled which is the default). In addition to the number of characters in the new password, credit (of +1 in length) is given for each different kind of character (other, upper, lower and digit).

src/modules/users/users.conf
es20490446e: What if we did the opposite? That we tried to make comments more concise?

I mean something like:
[users-proposed.conf](https://github.com/es20490446e/calamares/blob/calamares/src/modules/users/users-proposed.conf)

adriaandegroot:
Can I encourage you to take this apart into multiple PRs instead? There's a couple of themes I can identify:

- Changing the default checkbox-state of *autologin* from true to false. That would be mergeable, with the following parts:
  - Update the example `.conf` file
  - Change the default in the schema
  - Change the default in code
  - Add a line in `CHANGES-3.3` 
- Changing the defaults around allowing weak passwords. This is something that only really makes sense when there are password restrictions in place, but it is something that can be merged separately, with the following parts:
  - Update the example `.conf` file
  - Change the default in the schema
  - Change the default in code
  - Add a line in `CHANGES-3.3`
  - I'm not a big fan of the two existing key names in the configuration files about weak passwords. That's outside the scope of this (current) PR, but might be in-scope for a PR changing these settings specifically. A key like `strongPasswordRequiredCheckbox` could have subkeys `show` and `default`. That makes it easier to explain in one spot, but there's more code to write (also to accept previous-style config files). This expands the scope a lot, but gives the opportunity to dig through the code improving names (it's a mix of strong/weak and allow/permit/require, which is a mess) and defaults.
- Changing obvious things in the documentation in the example `.conf`, for instance you have a much nicer and succinct description for the minimum-password-length of -1. There's probably a half-dozen typo's, misplaced spaces, and too-wordy bits. Like I said elsewhere in this PR, there is also editorializing in the config file that can be removed.

Then there's somewhat more subjective things, like
- Re-ordering the config file (your proposed do-it-differently does that -- I think this is worth discussing at a higher level beforehand so it is visible that the PR does what it says on the tin)
- Changing the example config file so it comments-out the current weak settings and the active suggestion is something stronger.

Doing all of these things at once is hard to review -- even from the perspective of "this looks ok-ish". Your proposed "more succinct" configuration file is a no-go: as written, it is missing an introduction, missing a YAML section separator, and moves the documentation for a setting below the setting, which is unlike any of the other example files.

es20490446e:
Let me have a look...

es20490446e:
I will leave this as is. I will attach the conf just in case.

[users-proposed.conf.txt](https://github.com/calamares/calamares/files/14836592/users-proposed.conf.txt)

PR #2277, opened 2024-01-25
port slideshow to Qt 6

jriddell:
simple code copy so we have a Qt 5 and 6 version, although since the difference is only 1 line it might be better just to remove that one line.


dalto8:
Maintaining parallel QML files seems like a nightmare.

jriddell:
a nightmare but it's what is done with the qml modules

src/CMakeLists.txt
```diff
@@ -14,8 +14,13 @@ include(CalamaresAddTranslations)
 add_subdirectory(libcalamares)
 add_subdirectory(libcalamaresui)
 
-# all things qml
-add_subdirectory(qml/calamares)
+if(WITH_QT6)
+  # all things qml
+  add_subdirectory(qml/calamares-qt5)
```
demmm: Is this meant to be for WITH_QT6?  Seems this is for Qt5

adriaandegroot:
So, to be clear:
- makes a copy of all the QML things into a `calamares-qt5` directory
- tries to pick one of the two directories at build time based on the `WITH_QT6` flag

I would suggest renaming and copying, so we end up with a `calamars-qt5/` and a `calamares-qt6/` under the `qml/` directory.

And yes, maintaining Qt5 and Qt6 QML is a nightmare because of the source-incompatibilities. It is a road already taken, though, with the modules. I'd be willing to entertain a Qt6-only Calamares version (3.4?) once KDE Frameworks 6 happens and the KDE Megarelease is done.

src/CMakeLists.txt
jriddell: yes, fixed

jriddell:
@adriaandegroot yes that seems right.  I renamed the directory as you suggested

Issue #1938, opened 2022-05-01
Add checkbox to not encrypt /boot in GUI


sezanzeb:
**Is your feature request related to a problem? Please describe.**

1. Not encrypting boot requires knowledge of the partitions that need to be created
2. People might encrypt their disk and then run into the very user unfriendly non-optimized way of decrypting via grub (see https://github.com/calamares/calamares/issues/1311), not knowing that there is a way to improve that.
3. I'd like to avoid having to do this configuration every time I set up my computer

**Describe the solution you'd like**

A checkbox to not encrypt `/boot` with a tooltip that explains the benefits and risk

**Describe alternatives you've considered**

Having templates for the manual partitioning could also improve it, but would still be not as user friendly as simply a checkbox to tick.

sezanzeb:
Maybe like this

![image](https://user-images.githubusercontent.com/28510156/167107682-9117cb34-dba4-4b92-a5e5-526cef04c7a3.png)

And hovering it opens a tooltip that says "This will allow for faster booting, but poses a security risk"

---

or below it. Whatever text you would think works here

![image](https://user-images.githubusercontent.com/28510156/167108646-ef1d5c47-0fa9-4f7d-9366-675248c59ae8.png)

sezanzeb:
"This makes the decryption more user friendly, but makes grub vulnerable for malicious modification if physical access to the device is given"

"This makes the system decrypt the master key faster, but /boot will be vulnerable if physically accessed"

sezanzeb:
Ah, oops, the other way around!

GloriousEggroll:
The following patch is not a toggle, however it does disable /boot and swap from being encrypted, enables btrfs as the default, adds a 1gb ext4 boot partition, configures the remaining space as a btrfs partition, and sets "" for the root subvolume (essentially allowing usage with snapper). This more or less aligns with Fedora's default Anaconda installation using btrfs (+ encryption), and while boot and swap are not encrypted, this achieves the nice plymouth password gui rather than forcing the grub prompt.

The only difference between this and Fedora is that 

(1) Fedora creates subvolumes for root (/) and home (/home) by default. Those would look something like this in modules/mount.conf:
```
btrfsSubvolumes:
     - mountPoint: /
       subvolume: "/@root"
     - mountPoint: /home
       subvolume: "/@home"
```
(2) Fedora does not apply any swap by default because the Fedora-shipped kernel already enables zswap:

[...]

PR #2280, opened 2024-02-05
[partition] Allow specifying unencrypted partitions when encryption is used

ArrayBolt3:
Implements feature request https://github.com/calamares/calamares/issues/1938 sorta. There's no user-visible checkbox for encrypting or not encrypting `/boot`, instead there is an additional `noEncrypt` option added to the `partitionLayout` section of `partition.conf`, allowing a distro to specify that a certain partition should remain unencrypted even when encryption is enabled.


src/modules/partition/core/PartitionLayout.h
ArrayBolt3, line 1: gah, Vim struck again and did a weird mix of spaces plus tabs. Fixing...

ArrayBolt3: Fixed.

ArrayBolt3:
I'm realizing I didn't make this work right with ZFS partitions. I probably should go through and fix that.

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -116,6 +118,7 @@ PartitionLayout::init( FileSystem::Type defaultFsType, const QVariantList& confi
                           Calamares::getUnsignedInteger( pentry, "attributes", 0 ),
                           Calamares::getString( pentry, "mountPoint" ),
                           Calamares::getString( pentry, "filesystem", "unformatted" ),
+                          Calamares::getBool( pentry, "noEncrypt", true ),
```
dalto8: This should be `false` by default to mirror the prior behavior.

ArrayBolt3:
alright, that should work with ZFS encryption too. I did **not** test whether this results in unencrypted ZFS or not yet, but  I believe it will work as the logic is very straightforward.

src/modules/partition/partition.conf
```diff
@@ -293,6 +295,7 @@ defaultFileSystemType:  "ext4"
 #       - if "unknown" (or an unknown FS name, like "elephant") then the
 #         default filesystem type, or the user's choice, will be applied instead
 #         of "unknown" (e.g. the user might pick ext4, or xfs).
+#   - noEncrypt: whether this partition is exempt from encryption if enabled (optional parameter)
```
dalto8: We should mention the default state.

src/modules/partition/core/PartitionLayout.cpp
```diff
@@ -296,7 +299,7 @@ PartitionLayout::createPartitions( Device* dev,
         Partition* part = nullptr;
 
         // Encryption for zfs is handled in the zfs module
-        if ( luksPassphrase.isEmpty() || correctFS( entry.partFileSystem ) == FileSystem::Zfs )
+        if ( luksPassphrase.isEmpty() || correctFS( entry.partFileSystem ) == FileSystem::Zfs || entry.partNoEncrypt )
```
dalto8, line 302: We need to update the comment above this to reflect these changes.

dalto8 (changes requested):

src/modules/partition/core/PartitionLayout.cpp
ArrayBolt3: Good catch, that was an accident. Fixing now.

ArrayBolt3:
Got all changes implemented. (The default variable thing was a mistake since I went from implementing this as an `isEncryptable` option that defaulted to true, to a `noEncrypt` option that defaulted to false, and then forgot to change the default.)

dalto8:
I will take a look at it more depth tomorrow.  If memory serves, that isn't the only place we do the partition layout.(I could be wrong though)

The partition module is quite a mess.

ArrayBolt3:
Thanks!

I did a quick look through and the only other place I saw where
encrypted partitions are made is in PartitionActions.cpp, where an
(optionally encrypted) swap partition is made. I don't particularly
mind if the swap partition is *always* encrypted when encryption is
used - unencrypted swap breaks the security of FDE pretty much
entirely (at best it's a gaping security hole), and if someone's that
determined to configure their system in a broken way, they can use the
manual partitioning screen.

In case it's helpful, I did patch Ubuntu's Calamares locally with (the
earlier version of) these changes and successfully used it to make an
encrypted Kubuntu installation with unencrypted /boot without doing
manual partitioning.

dalto8 (approved):

dalto8 (approved):

PR #2284, opened 2024-02-13
[RFC] networkcfg: Configure NetworkManager to be the default Netplan renderer

slyon:
When Netplan is installed in the target system:

In case NM is not yet set to be the default Netplan renderer (e.g. through a `/usr/lib/netplan/00-network-manager-all.yaml` file shipped by an installed package), create the `/etc/netplan/01-network-manager-all.yaml` configuration and copy over all other Netplan configuration from the installer system.

This change would enable to install target systems similar to what is being proposed for the `debian-installer`: https://salsa.debian.org/installer-team/netcfg/-/merge_requests/9 but always using the NetworkManager backend, as that's what we do on Calamares when installing a desktop system.

### DIFF
--- a/src/modules/networkcfg/main.py
+++ b/src/modules/networkcfg/main.py
@@ -138,7 +138,27 @@ def run():
     target_netplan = os.path.join(root_mount_point, source_netplan.lstrip('/'))
 
     if os.path.exists(source_netplan) and os.path.exists(target_netplan):
-        for cfg in glob.glob(os.path.join(source_netplan, "90-NM-*")):
+        # Set NetworkManager to be the default renderer if Netplan is installed
+        # TODO: We might rather do that inside the network-manager package, see:
+        # https://bugs.launchpad.net/ubuntu/+source/ubuntu-settings/+bug/2020110
+        default_renderer = os.path.join(root_mount_point, "usr/lib/netplan",
+                                        "00-network-manager-all.yaml")
+        if not os.path.exists(default_renderer):
+            renderer_file = os.path.join(target_netplan,
+                                         "01-network-manager-all.yaml")
+            nm_renderer = """# This file was written by calamares.
+# Let NetworkManager manage all devices on this system.
+# For more information, see netplan(5).
+network:
+  version: 2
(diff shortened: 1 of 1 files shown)


slyon:
@tsimonq2 This is related to your commit https://github.com/calamares/calamares/commit/e5db326a7e4ea8ae70f88378d18020cbdde8a1bf What do you think about this?

tsimonq2:
This does make sense, yes :smile: I do think we should aim to have this fixed within the network-manager package itself, let's follow up elsewhere on how that should be done (ideally before the 18th, which is the next planned Calamares release.)

Also, for upstream folks, this only affects installs with [Netplan](https://netplan.io/) + Network Manager, which is the default stack in Ubuntu (and thus, derivatives of 24.04.) Specifically, the stanza beginning with this line...
```python3
for cfg in glob.glob(os.path.join(source_netplan, "90-NM-*")):
```
...runs iff there are Netplan configuration files. It **is** modified to this...
```python3
for cfg in glob.glob(os.path.join(source_netplan, "*.yaml")):
```
...which is quite acceptable. I originally wrote it as `90-NM-*` just to be safe.

The original intention behind my commit was simply to unblock; I do appreciate the eyes on this! If you find anything else, I would be happy to review again.

### +1, thank you!

slyon:
> I do think we should aim to have this fixed within the network-manager package itself, let's follow up elsewhere on how that should be done (ideally before the 18th, which is the next planned Calamares release.)

Agreed. Thanks for updating the corresponding bug report in [LP: #2020110](https://bugs.launchpad.net/ubuntu/+source/ubuntu-settings/+bug/2020110). I do think we do not need to focus on the 18th deadline, though. As the two implementations can be kept in parallel, providing a clean migration path, as outlined in [LP: merge#443210](https://code.launchpad.net/~waveform/ubuntu/+source/ubuntu-settings/+git/ubuntu-settings/+merge/443210), where the "unowned" Calamares file would be deleted/moved on package upgrade, whenever that happens.

tsimonq2:
@slyon That's quite rational, thanks!

PR #2288, opened 2024-02-20
[packages] Add a DNF5 PM backend

### LINKED ISSUE #2145: Prepare for DNF 5
DNF 5 is a new package manager that will replace DNF 4 in Fedora 39+. Starting in Fedora 39, the `dnf` command will be provided by the `dnf5` package rather than the `dnf` package, and `dnf5` will obsolete `dnf`. Since Calamares currently depends on DNF 4, it should choose one of the following strategies to avoid breaking the approaching Fedora upgrade and maintain compatibility with future versions of Fedora:
- Add support for DNF 5, and depend on the `dnf5` package in Fedora 39+ instead of dnf. The command-line interface isn't changing a whole lot between DNF 4 and DNF 5, so in Calamares's case, this change may be pretty quick. Builds of DNF 5 are available in this COPR repository: https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf5-unstable, and documentation is available here: https://dnf5.readthedocs.io/en/latest/. You can call DNF 5 explicitly at `/usr/bin/dnf5`.
- Change the Calamares package to depend on `python3-dnf` instead of `dnf`, and call the `dnf-3` binary instead of `dnf`. The old DNF 4 command will still be available in the distribution, but only as `dnf-3`. (The binary is called `dnf-3`  rather than `dnf4` for historical reasons; it was originally the Python 3 version of DNF). The first option is preferred to this one; it is not recommended to modify installed software using both DNF 4 and DNF 5 on the same system.
  - I drafted up a small PR that should implement this change: https://github.com/calamares/calamares/pull/2144

We are planning to replace DNF with DNF5 in Fedora Rawhide very soon, in the next couple weeks, and the `calamares` package will break as long as it still depends on the `dnf` package.

After the necessary changes are made here in upstream, I'm happy to file the bugs in the Fedora issue tracker to get the package dependencies changed.

For more information about the switch to DNF 5, see https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5.

adriaandegroot:
I just made this up, checked that the `dnf5` command in Fedora 39 exists and can run the command-lines created by this PM backend. (I did install dnf5 into that Fedora 39 system first)

FIXES #2145

### DIFF
--- a/src/modules/packages/main.py
+++ b/src/modules/packages/main.py
@@ -279,6 +279,11 @@ class PMApt(PackageManager):
 
 
 class PMDnf(PackageManager):
+    """
+    This is "legacy" DNF, called DNF-4 even though the
+    executable is dnf-3 in modern Fedora. Executable dnf
+    is a symlink to dnf-3 in systems that use it.
+    """
     backend = "dnf"
 
     def install(self, pkgs, from_local=False):
@@ -296,6 +301,30 @@ class PMDnf(PackageManager):
 
     def update_system(self):
         check_target_env_call(["dnf-3", "-y", "upgrade"])
+
+
+class PMDnf5(PackageManager):
+    """
+    This is "modern" DNF, DNF-5 which is for Fedora 41 (presumably)
+    and later. Executable dnf is a symlink to dnf5 in systems that use it.
+    """
+    backend = "dnf5"
+
+    def install(self, pkgs, from_local=False):
+        check_target_env_call(["dnf5", "-y", "install"] + pkgs)
+
+    def remove(self, pkgs):
+        # ignore the error code for now because dnf thinks removing a
+        # nonexistent package is an error
+        target_env_call(["dnf5", "--disablerepo=*", "-C", "-y",
+                         "remove"] + pkgs)
+
+    def update_db(self):
+        # Doesn't need updates
+        pass
+
+    def update_system(self):
+        check_target_env_call(["dnf5", "-y", "upgrade"])
 
 
 class PMDummy(PackageManager):
--- a/src/modules/packages/packages.conf
+++ b/src/modules/packages/packages.conf
@@ -23,6 +23,7 @@
 #  - apk         - Alpine Linux package manager
 #  - apt         - APT frontend for DEB and RPM
 #  - dnf         - DNF, the new RPM frontend
+#  - dnf5        - DNF5, the newer new RPM frontend
 #  - entropy     - Sabayon package manager (is being deprecated)
 #  - luet        - Sabayon package manager (next-gen)
 #  - packagekit  - PackageKit CLI tool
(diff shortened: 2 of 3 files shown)


adriaandegroot:
@evan-goode can you take a look, too? You mention that `dnf` will be `dnf5` in #2145, but the *packages* module now calls the executble `dnf-3`. I just duplicated the package-manager class and wrote `dnf5` instead, and that seems to work.

kkofler (approved):
I think this is OK.

So the thing is that `dnf` is now ambiguous. In older Fedora releases, `dnf` is a symlink to `dnf-3` which is DNF 4 (used to be DNF 3 a long long time ago). In newer Fedora releases, `dnf` will be a symlink to `dnf5` (without the dash) instead. So this now makes a separate backend for the new version, after the old backend was made to force `dnf-3`.

Whether we will actually need different switches between the two DNF versions is not clear. For now, the switches are the same.

dalto8:
Unless there is a divergence in usage do we really need a separate backend?  We don't maintain separate backends for the different versions of other package managers.

As an alternative approach, couldn't we drive the correct binary to use via a config value?

kkofler:
Well, DNF5 is not really just a new version of DNF, it is largely a rewrite. DNF ≤ 4 was written in Python, with underlying libraries (libdnf, libsolv, …) in C or C++. There was a separate C frontend to the C/C++ libraries called MicroDNF, with reduced functionality, for use in cloud images and the like. In DNF 5, the CLI frontend was rewritten in C++ to eliminate the Python dependency and the duplication between DNF and MicroDNF.

dalto8:
How it got from version 4 to version 5 isn't really material.

To me, the question is, do we need to maintain two separate classes for it?  Unless the usage is different as it relates to Calamares, I would say we don't.  Other package managers also change/add/remove functionality as they evolve and we don't do it for those.

evan-goode:
LGTM

> To me, the question is, do we need to maintain two separate classes for it? Unless the usage is different as it relates to Calamares, I would say we don't. Other package managers also change/add/remove functionality as they evolve and we don't do it for those.

These classes are pretty tiny, so I'd say it doesn't make much difference either way. If you want to add some new feature that depends on functionality that's different between DNF 4 and DNF 5, then you'll want the two separate classes. But for now, it looks like the commands you call here work interchangeably with both DNF 4 and DNF 5.

My vote is to merge this, but another option is to leave the old DNF implementation alone and call `/usr/bin/dnf` instead of `/usr/bin/dnf-3` so DNF 5 will be used on systems where it is the default (where it provides `/usr/bin/dnf`).

adriaandegroot:
I merged this based on Evan's comment, but also because we would have to do **something**: the current DNF backend explicitly (now) runs `dnf-3` which was a reaction to the upcoming dnf5 which was supposed to be more-different than it turned out. So we would have to patch the DNF backend to run `dnf` again, or .. something. 

Well, I guess it is slightly more complicated now because a distro needs to know which version of DNF it is running.

PR #2292, opened 2024-02-26
[partition] Make default partition entry have `partNoEncrypt` be false

vlinkz:
#2280 added the `noEncrypt` field to `PartitionEntry`s. However, the current initialization in the `PartitionLayout` class leaves the value uninitialized, leading to the default system partition not being included in encryption.

Before my patch:
before patch, system partition is not encrypted with LUKS

After my patch:
after patch, system partition is encrypted with LUKS

### DIFF
--- a/src/modules/partition/core/PartitionLayout.h
+++ b/src/modules/partition/core/PartitionLayout.h
@@ -37,7 +37,7 @@ public:
         quint64 partAttributes = 0;
         QString partMountPoint;
         FileSystem::Type partFileSystem = FileSystem::Unknown;
-        bool partNoEncrypt;
+        bool partNoEncrypt = false;
         QVariantMap partFeatures;
         Calamares::Partition::PartitionSize partSize;
         Calamares::Partition::PartitionSize partMinSize;


dalto8:
Is this the right fix or would it be better to properly set the default entry here:
https://github.com/calamares/calamares/blob/calamares/src/modules/partition/core/PartitionLayout.cpp#L136

As a side note, I wonder if we should even have a default config here.  If a distro has a broken config, shouldn't we fail in that case so they know to fix their config?

vlinkz:
As for the other alternative, it could be initialized in the constructor instead with a default as false for the `addEntry` instance you linked.
https://github.com/calamares/calamares/blob/calamares/src/modules/partition/core/PartitionLayout.cpp#L44-L81
```diff
PartitionLayout::PartitionEntry::PartitionEntry( FileSystem::Type fs,
                                                 const QString& mountPoint,
                                                 const QString& size,
                                                 const QString& minSize,
                                                 const QString& maxSize )
     : partAttributes( 0 )
     , partMountPoint( mountPoint )
     , partFileSystem( fs )
+    , partNoEncrypt( false )
     , partSize( size )
     , partMinSize( minSize )
     , partMaxSize( maxSize )
{
}

PartitionLayout::PartitionEntry::PartitionEntry( const QString& label,
                                                 const QString& uuid,
                                                 const QString& type,
                                                 quint64 attributes,
                                                 const QString& mountPoint,
                                                 const QString& fs,
                                                 const bool& noEncrypt,
                                                 const QVariantMap& features,
                                                 const QString& size,
                                                 const QString& minSize,
                                                 const QString& maxSize )
    : partLabel( label )
    , partUUID( uuid )
    , partType( type )
    , partAttributes( attributes )
    , partMountPoint( mountPoint )
+   , partNoEncrypt( noEncrypt )
    , partFeatures( features )
    , partSize( size )
    , partMinSize( minSize )
    , partMaxSize( maxSize )
 {
     PartUtils::canonicalFilesystemName( fs, &partFileSystem );
-    partNoEncrypt = noEncrypt;
 }
 bool
```

> As a side note, I wonder if we should even have a default config here. If a distro has a broken config, shouldn't we fail in that case so they know to fix their config?

That may be a good idea, or perhaps store the default config within the [partition.conf](https://github.com/calamares/calamares/blob/calamares/src/modules/partition/partition.conf#L243-L306)

adriaandegroot:
This fixes UB, and I'll follow up with @vlinkz 's second suggestion as well. The constructors are now a weird mix of initializer-list and other initializations of data members.

dalto8:
> The constructors are now a weird mix of initializer-list and other initializations of data members.

Yeah, I noticed that above.  That is my fault.  I should have caught that during the code review.

PR #2297, opened 2024-03-02
Create SECURITY.md

adonis-jimenez:


demmm:
A separate file is not the best option for this in my opinion.  Why not add to the bug report file or main README?

dalto8:
Either way, this is not mergable in it's current state.  We shouldn't merge boilerplate.  If we are going to merge it, in addition to deciding where this information should be located, it needs to updated to have the specific details for Calamares.

kkofler:
Invalid pull request, boilerplate nonsense, closing.

PR #2299, opened 2024-03-04
initcpiocfg: Add microcode module

ptr1337:
As mentioned in the recent mkinitcpio and archlinux news, the ALL_MICROCODE way got deprecated and there is now a 'microcode' hook.
This should be adapted.

https://archlinux.org/news/mkinitcpio-hook-migration-and-early-microcode/

### DIFF
--- a/CHANGES-3.3
+++ b/CHANGES-3.3
@@ -12,6 +12,7 @@ the history of the 3.2 series (2018-05 - 2022-08).
 This release contains contributions from (alphabetically by first name):
  - Adriaan de Groot
  - Evan James
+ - Peter Jung
 
 ## Core ##
  - Calamares logs more information about how the executable was created
@@ -22,6 +23,7 @@ This release contains contributions from (alphabetically by first name):
 ## Modules ##
  - *displaymanager* module can configure an alternate SDDM configuration file.
  - *networkcfg* a bug affecting NetPlan + NetworkManager was fixed.
+ - *initcpiocfg* Add microcode hook to initcpiocfg
 
 
 # 3.3.4 (2024-02-27)
--- a/src/modules/initcpiocfg/main.py
+++ b/src/modules/initcpiocfg/main.py
@@ -148,6 +148,7 @@ def find_initcpio_features(partitions, root_mount_point):
     """
     hooks = [
         "autodetect",
+        "microcode",
         "kms",
         "modconf",
         "block",


dalto8:
<s>We probably shouldn't make this change unless we also change `bootloader` at the same time.</s>

Is there anything that needs to be changed in `grubcfg`?  I can't think of anything off the top of my head.

Also, does this need to be done conditionally?  What about a distro that isn't using the latest `mkinitcpio`?  It isn't only Arch that uses it.

ptr1337:
> We probably shouldn't make this change unless we also change `bootloader` at the same time.
> 
> Is there anything that needs to be changed in `grubcfg`? I can't think of anything off the top of my head.
> 
> Also, does this need to be done conditionally? What about a distro that isn't using the latest `mkinitcpio`? It isn't only Arch that uses it.

At systemd-boot we are handling this over systemd-boot-manager, otherwise manual action is required.
For grub I dont find anything in the default entry, but yes that should be also changed.

Since most arch based distrubutions are following the arch repo, they will also get the recent updates. I think it is pretty important to pull this in.

I dont see anything currently in the bootloader module, which is handling that.

demmm:
Just a question, so this won't cause build failures for some, are you sure all mkinitcpio users, use the latest Arch packages?
EDIT:
@dalto8 was quicker :)

dalto8:
> At systemd-boot we are handling this over systemd-boot-manager

We also need to consider distros that use the Calamares bootloader module which needs changes to support this.

> Since most arch based distrubutions are following the arch repo

Again, what about the distros that aren't and the non-Arch distros which also use `mkinitcpio`?

demmm:
Here a list:
https://repology.org/project/mkinitcpio/versions
(KaOS does not use mkinitcpio in Calamares or installs, so no need to include that one)

ptr1337:
> Here a list: https://repology.org/project/mkinitcpio/versions

Manjaro does maintain its own version of calamares anyways, they can handle it on their own.
All other seems to follow the archlinux releases.

dalto8:
Actually, `bootloader` uses `kernel-install` so it doesn't need changes.

ptr1337:
grub seems to automatically add it, someone created an issue to avoid adding the ucode to /boot/*-ucode.img.

I think this can be merged, no changes required on the bootloader module.

ptr1337:
@dalto8 
For grub following can be used:
https://github.com/CachyOS/cachyos-calamares/commit/f2896c3bcc150bb31a618abb146f8a507bc6e2cc

dalto8:
> @dalto8 For grub following can be used: [CachyOS/cachyos-calamares@f2896c3](https://github.com/CachyOS/cachyos-calamares/commit/f2896c3bcc150bb31a618abb146f8a507bc6e2cc)

It really seems like that should be changed by default in Arch's grub package.  There was an issue opened about it yesterday but it was closed by the person who opened it.

dalto8 (approved):

PR #2300, opened 2024-03-07
Active Directory Support

tsimonq2:
Probably needs to use an existing QProcess type instead of pure QProcess, otherwise this code does work.

### DIFF
--- a/src/modules/users/ActiveDirectoryJob.cpp
+++ b/src/modules/users/ActiveDirectoryJob.cpp
@@ -1,0 +1,91 @@
+/* === This file is part of Calamares - <https://calamares.io> ===
+ *
+ *   SPDX-FileCopyrightText: 2024 Simon Quigley <tsimonq2@ubuntu.com>
+ *   SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include "ActiveDirectoryJob.h"
+
+#include "Config.h"
+
+#include "GlobalStorage.h"
+#include "JobQueue.h"
+#include "utils/Logger.h"
+#include "utils/Permissions.h"
+#include "utils/System.h"
+
+#include <QDateTime>
+#include <QDir>
+#include <QFile>
+#include <QFileInfo>
+#include <QTextStream>
+#include <QProcess>
+
+ActiveDirectoryJob::ActiveDirectoryJob(QStringList& activeDirectoryInfo)
+    : Calamares::Job()
+    , m_activeDirectoryInfo(activeDirectoryInfo)
+{
+}
+
+QString
+ActiveDirectoryJob::prettyName() const
+{
+    return tr( "Enroll system in Active Directory" );
+}
+
+QString
+ActiveDirectoryJob::prettyDescription() const
+{
+    return tr( "Enroll system in Active Directory" );
+}
+
+QString
+ActiveDirectoryJob::prettyStatusMessage() const
+{
+    return tr( "Enrolling system in Active Directory" );
+}
+
+Calamares::JobResult
+ActiveDirectoryJob::exec()
+{
+    QString username = m_activeDirectoryInfo.value(0);
+    QString password = m_activeDirectoryInfo.value(1);
+    QString domain = m_activeDirectoryInfo.value(2);
+    QString ip = m_activeDirectoryInfo.value(3);
+
+    Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
+    QString rootMountPoint = gs ? gs->value("rootMountPoint").toString() : QString();
+
+    if (!ip.isEmpty()) {
+        QString hostsFilePath = !rootMountPoint.isEmpty() ? rootMountPoint + "/etc/hosts" : "/etc/hosts";
+        QFile hostsFile(hostsFilePath);
+        if (hostsFile.open(QIODevice::Append | QIODevice::Text)) {
+            QTextStream out(&hostsFile);
+            out << ip << " " << domain << "\n";
+            hostsFile.close();
+        } else {
+            return Calamares::JobResult::error("Failed to open /etc/hosts for writing.");
+        }
+    }
+
+    QString installPath = !rootMountPoint.isEmpty() ? rootMountPoint : "/";
+    QStringList args = {"join", domain, "-U", username, "--install=" + installPath, "--verbose"};
+
+    QProcess process;
+    process.start("realm", args);
+    process.waitForStarted();
+
+    if (!password.isEmpty()) {
+        process.write((password + "\n").toUtf8());
+        process.closeWriteChannel();
+    }
+
+    process.waitForFinished(-1);
+
+    if (process.exitCode() == 0) {
+        return Calamares::JobResult::ok();
+    } else {
+        QString errorOutput = process.readAllStandardError();
+        return Calamares::JobResult::error(QString("Failed to join realm: %1").arg(errorOutput));
+    }
+}
--- a/src/modules/users/ActiveDirectoryJob.h
+++ b/src/modules/users/ActiveDirectoryJob.h
@@ -1,0 +1,29 @@
+/* === This file is part of Calamares - <https://calamares.io> ===
+ *
+ *   SPDX-FileCopyrightText: 2024 Simon Quigley <tsimonq2@ubuntu.com>
+ *   SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ *   Calamares is Free Software: see the License-Identifier above.
+ *
+ */
+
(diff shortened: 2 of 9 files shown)


src/modules/users/ActiveDirectoryJob.cpp
```diff
@@ -0,0 +1,91 @@
...
+{
+    QString username = m_activeDirectoryInfo.value(0);
+    QString password = m_activeDirectoryInfo.value(1);
+    QString domain = m_activeDirectoryInfo.value(2);
+    QString ip = m_activeDirectoryInfo.value(3);
+
+    Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
+    QString rootMountPoint = gs ? gs->value("rootMountPoint").toString() : QString();
+
+    if (!ip.isEmpty()) {
```
tsimonq2, line 59: taking suggestions for a better way to do this

src/modules/users/ActiveDirectoryJob.cpp
```diff
@@ -0,0 +1,91 @@
...
+            hostsFile.close();
+        } else {
+            return Calamares::JobResult::error("Failed to open /etc/hosts for writing.");
+        }
+    }
+
+    QString installPath = !rootMountPoint.isEmpty() ? rootMountPoint : "/";
+    QStringList args = {"join", domain, "-U", username, "--install=" + installPath, "--verbose"};
+
+    QProcess process;
```
tsimonq2, line 74: should be a Calamares process not a plain QProcess

src/modules/users/ActiveDirectoryJob.cpp
```diff
@@ -0,0 +1,91 @@
...
+ActiveDirectoryJob::ActiveDirectoryJob(QStringList& activeDirectoryInfo)
+    : Calamares::Job()
+    , m_activeDirectoryInfo(activeDirectoryInfo)
+{
+}
+
+QString
+ActiveDirectoryJob::prettyName() const
+{
+    return tr( "Enroll system in Active Directory" );
```
adriaandegroot, line 33: These strings typically get an annotation (`@title` for `prettyName()`, for instance). Take a look at the way translations are done in other modules. This helps the translators with a little context to show how to translate otherwise-identical-in-English strings. You can ask @bitigchi (also in KDE-translation channels, I think) for hints.

src/modules/users/ActiveDirectoryJob.h
```diff
@@ -0,0 +1,29 @@
...
+    Q_OBJECT
+public:
+    ActiveDirectoryJob( QStringList& activeDirectoryInfo );
+    QString prettyName() const override;
+    QString prettyDescription() const override;
+    QString prettyStatusMessage() const override;
+    Calamares::JobResult exec() override;
+
+private:
+    QStringList m_activeDirectoryInfo;
```
adriaandegroot, line 26: It is probably worthwhile to write out the four strings you're interested in, rather than glomming them in a string-list. That is, `QString m_username;`, ...

src/modules/users/ActiveDirectoryJob.cpp
```diff
@@ -0,0 +1,91 @@
...
+    return tr( "Enrolling system in Active Directory" );
+}
+
+Calamares::JobResult
+ActiveDirectoryJob::exec()
+{
+    QString username = m_activeDirectoryInfo.value(0);
+    QString password = m_activeDirectoryInfo.value(1);
+    QString domain = m_activeDirectoryInfo.value(2);
+    QString ip = m_activeDirectoryInfo.value(3);
```
adriaandegroot, line 54: I think it would be better to use four separate data-members for this. Or, failing that, make these `const`.

src/modules/users/ActiveDirectoryJob.cpp
adriaandegroot: I think there's a whole family of "edit file in the target system" functions that might be useful in Calamares (also, eventually, when Calamares loses root access) but this isn't the MR for that.

src/modules/users/ActiveDirectoryJob.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
+    QString rootMountPoint = gs ? gs->value("rootMountPoint").toString() : QString();
```
adriaandegroot, line 57: Can be `const`.

src/modules/users/ActiveDirectoryJob.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    if (!ip.isEmpty()) {
+        QString hostsFilePath = !rootMountPoint.isEmpty() ? rootMountPoint + "/etc/hosts" : "/etc/hosts";
```
adriaandegroot, line 60: It doesn't really matter if `rootMountPoint` is empty (if it is, `rootMountPoint + "something"` is "something"). So this can simplify to `const QString hostsFilePath = rootMountPoint + "/etc/hosts";` or even simpler, `Calamares::System::instance()->targetPath(QStringLiteral("/etc/hosts"));` which is the API for getting paths to the target system.

adriaandegroot: "even simpler" is relative, of course :/

src/modules/users/ActiveDirectoryJob.cpp
adriaandegroot: `System::instance()->runCommand()` or some variant thereof, it can handle stdin and run the program in either the host (like you do here) or the target. You can probably simplify `installPath` using `targetPath("/")`, too.

src/modules/users/ActiveDirectoryJob.cpp
```diff
@@ -0,0 +1,91 @@
...
+    process.waitForStarted();
+
+    if (!password.isEmpty()) {
+        process.write((password + "\n").toUtf8());
+        process.closeWriteChannel();
+    }
+
+    process.waitForFinished(-1);
+
+    if (process.exitCode() == 0) {
```
adriaandegroot, line 85: If you use `runCommand()` you get a `ProcessResult` and can call `return explainProcess()` to wrangle stderr and exit codes for you.

src/modules/users/Config.cpp
```diff
@@ -656,6 +657,59 @@ Config::setRootPasswordSecondary( const QString& s )
...
+}
+
+void
+Config::setActiveDirectoryIP( const QString & s )
+{
+    m_activeDirectoryIP = s;
+}
+
+QStringList&
+Config::getActiveDirectory() const
```
adriaandegroot, line 703: Is this used somewhere? It seems like a very weird API given the setters and getters above.

src/modules/users/Config.cpp
```diff
@@ -656,6 +657,59 @@ Config::setRootPasswordSecondary( const QString& s )
...
+}
+
+void
+Config::setActiveDirectoryAdminUsername( const QString & s )
+{
+    m_activeDirectoryUsername = s;
+}
+
+void
+Config::setActiveDirectoryAdminPassword( const QString & s )
```
adriaandegroot, line 685: This API (the others, too) seems a bit peculiar to me because the name of the method says "Admin" but the data member does not. Whose password **is** it then?

src/modules/users/Config.h
```diff
@@ -343,6 +355,14 @@ public Q_SLOTS:
 
     bool m_isReady = false;  ///< Used to reduce readyChanged signals
 
+    mutable QStringList m_activeDirectorySettings;
```
adriaandegroot, line 358: Why `mutable`? (This isn't Rust; C++ has the defaults wrong, so everything it mutable by default)

src/modules/users/UsersPage.cpp
```diff
@@ -162,6 +162,34 @@ UsersPage::UsersPage( Config* config, QWidget* parent )
             config, &Config::requireStrongPasswordsChanged, ui->checkBoxRequireStrongPassword, &QCheckBox::setChecked );
     }
 
+    // Active Directory is not checked or enabled by default
+    ui->useADCheckbox->setVisible(m_config->getActiveDirectoryEnabled());
+    ui->domainLabel->setVisible(false);
```
adriaandegroot, line 167: This collection of `setVisible()` as well as the one connected to `toggled()` would be simplified if you put the collection of widgets that is relevant for AD into a parent widget (a `QFrame` for instance) so you can show / hide the parent widget.

adriaandegroot:
So overall -- assuming `realm` does what it says -- this looks fine, although I have lots of specific comments about the code.

PR #2301, opened 2024-03-10
on Qt6 we need to build with position independent code

hsitter:
otherwise linking may fail with

> copy relocation against non-copyable protected symbol `qt_resourceFeatureZstd@@Qt_6'

(don't ask me why, we did the same in KDE code and I didn't bother to ask why XD)


adriaandegroot:
I guess it's because this is an object library that doesn't link against anything -- once the C++ is generated by rcc there isn't anything strictly-Qt-related in there. So we need to manually add some flags that would otherwise be there from the Qt build.

PR #2303, opened 2024-03-18
[mount] Ensure efi partition gets correct options

dalto8:
It was recently pointed out to me that I failed to upstream this fix to allow the correct options to be set for the EFI partition.

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -74,13 +74,14 @@ def is_ssd_disk(partition):
         return False
 
 
-def get_mount_options(filesystem, mount_options, partition):
+def get_mount_options(filesystem, mount_options, partition, efi_location = None):
     """
     Returns the mount options for the partition object and filesystem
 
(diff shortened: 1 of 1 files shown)


ezsit55:
Thanks for looking into this. I have been thinking about this issue and I have come to the conclusion that using the setting:

mountOptions:
      - filesystem: vfat
      options: [ defaults, umask=0077 ]

In the mount.conf file is actually preferable to using - filesystem: efi since vfat is the actual filesystem name and efi is not a filesystem. Maybe just the mount.conf file needs to be changed and the code that selects the correct partition to apply the options to is already sufficient. Just a thought. Thanks again.

dalto8:
If you the only vfat partition your distro has is the EFI partition, then it works the same no matter which option you choose.

However, if you have another vfat partition, that approach doesn't work since the ESP needs more restrictive permissions than you would probably want on other partitions.

Either way, I don't think we should leave an existing option broken.

PR #2304, opened 2024-03-24
Update Presentation.qml

manchicken:
The `padding` property is present in the qt5 version, but not the qt6 version. I notice a syntax error as a result of this missing propertly.


erikdubois:
I suppose it is this message?
image

manchicken:
naw, it was this:

ima3178212991179264277

PR #2305, opened 2024-03-29
plymouthcfg: Use plymouth-set-default-theme instead of sed

ptr1337:
Currently we are running into an issue, when setting the default theme via "sed", since the "[DAEMON]" entry in /etc/plymouth/plymouthd.conf is commented, which results into that the theme is not correctly applied.


adriaandegroot:
Seems reasonable. I'll do some post-merge tweaks.

Issue #2310, opened 2024-04-10
Module fstab should not add keyfile/password options in crypttab when luksbootkeyfile module is disabled


eugenesan:
I am working on Kubuntu installer with systemd-boot, encrypted root, no boot partition.
So far I managed to get it working by altering only config files but there is one issue that requires code change.

When I remove `luksbootkeyfile`, `fstab` still adds keyfile/password option to crypttab, making target unbootable without manual edit of crypttab.

I guess the following logic
https://github.com/calamares/calamares/blob/1645874973669475954e039fbfbd5b3d126fa938/src/modules/fstab/main.py#L159
should check when `luksbootkeyfile` is disabled or/and when bootloader doesn't require access to /boot as it is with systemd-boot which boots linux/initrd from EFI.

Tested on Ubuntu 24.04, Calamares 3.3.5

dalto8:
I think we should fix line: https://github.com/calamares/calamares/blob/1645874973669475954e039fbfbd5b3d126fa938/src/modules/fstab/main.py#L154

It should be "none" by default and only get set to the keyfile path if the keyfile exists.

PR #2313, opened 2024-04-12
Add support for splash for non-grub bootloaders and fix adding crypto_keyfile when it's missing

### LINKED ISSUE #2311: Feature request: module bootloader should check for plymouth and splash (similarly to grubcfg)
Currently when using systemd-boot, bootloader module doesn't check if plymouth is installed and doesn't add "splash" to kernel options.

I suggest replicating the functionality from grubcfg
https://github.com/calamares/calamares/blob/1645874973669475954e039fbfbd5b3d126fa938/src/modules/grubcfg/main.py#L149

eugenesan:
The following enables adding "splash" to cmdline when plymouth is installed (independently of grubcfg) and fixes adding crypto_keyfile when it's missing (for example when luksbootkeyfile is disabled).
Fixes https://github.com/calamares/calamares/issues/2311 and https://github.com/calamares/calamares/issues/2310 respectively.

TLDR;
Improvements for systemd-boot support.
With those fixes, it's possible to configure Kubuntu/Lubuntu installer to utilize 2 partition config (EFI + Encrypted root).

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -127,7 +127,6 @@ def is_zfs_root(partition):
 
 def get_kernel_params(uuid):
     kernel_params = libcalamares.job.configuration.get("kernelParams", ["quiet"])
-    kernel_params.append("rw")
 
     partitions = libcalamares.globalstorage.value("partitions")
     swap_uuid = ""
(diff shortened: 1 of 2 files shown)


demmm:
Since it is a common practice to keep commits per module (see https://github.com/calamares/calamares/wiki/Develop-Code, last paragraph), it would be better to split this PR up in one for fstab & one for the bootloader module.

The fstab changes probably won't need much discussing, adding to the bootloader module might.

dalto8:
Yes, the changes for crypto keyfile and the changes for splash are unrelated.  Please split them into 2 PRs.

For the keyfile changes, we should refactor it so we don't set it wrong in the first place, instead of modifying it at the end like that.

eugenesan:
Obsoleted by https://github.com/calamares/calamares/pull/2314 and https://github.com/calamares/calamares/pull/2315

eugenesan:
Obsoleted by https://github.com/calamares/calamares/pull/2314 and https://github.com/calamares/calamares/pull/2315

PR #2314, opened 2024-04-12
Refactor crypto_keyfile addition in fstab module and skip it when it's missing

### LINKED ISSUE #2310: Module fstab should not add keyfile/password options in crypttab when luksbootkeyfile module is disabled

eugenesan:
Fixes https://github.com/calamares/calamares/issues/2310
Replaces https://github.com/calamares/calamares/pull/2313

### DIFF
--- a/src/modules/fstab/main.py
+++ b/src/modules/fstab/main.py
@@ -151,8 +151,12 @@ class FstabGenerator(object):
         if not mapper_name or not luks_uuid:
             return None
 
-        password = "/crypto_keyfile.bin"
         crypttab_options = self.crypttab_options
+        # Make sure to not use missing keyfile
+        if os.path.isfile(os.path.join(self.root_mount_point, "crypto_keyfile.bin")):
+            password = "/crypto_keyfile.bin"
+        else:
+            password = "none"
 
         # Set crypttab password for partition to none and remove crypttab options
         # if root partition was not encrypted


src/modules/fstab/main.py
```diff
@@ -151,23 +151,24 @@ def generate_crypttab_line_info(self, partition):
...
+           # it's missing
+           ( not os.path.isfile(os.path.join(self.root_mount_point, "crypto_keyfile.bin")) ) or \
+           # root partition was not encrypted
+           ( any([p["mountPoint"] == "/"
+                  and "luksMapperName" not in p for p in self.partitions]) ) or \
+           # it's on root partition when /boot is unencrypted
+           ( partition["mountPoint"] == "/"
+                  and any([p["mountPoint"] == "/boot"
+                  and "luksMapperName" not in p for p in self.partitions]) )
+            ):
```
dalto8: I haven't tested this, but would something like this work to keep it more readable?

```suggestion
        crypttab_options = self.crypttab_options
        if os.path.isfile(os.path.join(self.root_mount_point, "crypto_keyfile.bin")):
            password = "/crypto_keyfile.bin"
        else:
            password = "none"

        # if root partition was not encrypted
        if any([p["mountPoint"] == "/" and
               "luksMapperName" not in p for p in self.partitions]):
            crypttab_options = ""
       # on root partition when /boot is unencrypted
        elif partition["mountPoint"] == "/" and
             any([p["mountPoint"] == "/boot"
                 and "luksMapperName" not in p for p in self.partitions]):
            crypttab_options = ""
```

dalto8:
Thanks for the contribution!

I made some suggested changes.  Please take a look when you have time.

eugenesan:
> Thanks for the contribution!
> 
> I made some suggested changes. Please take a look when you have time.

Looks good. Also seems like you fixed passing of the options which might not be keyfile related.

src/modules/fstab/main.py
```diff
@@ -151,23 +151,27 @@ def generate_crypttab_line_info(self, partition):
...
-        elif partition["mountPoint"] == "/":
-            if any([p["mountPoint"] == "/boot"
-                   and "luksMapperName" not in p
-                   for p in self.partitions]):
-                password = "none"
-                crypttab_options = ""
+        # Setup keyfile
+        else:
+            password = "/crypto_keyfile.bin"
+            crypttab_options = self.crypttab_options
```
dalto8: I think these lines and the two directly above are no longer needed.

eugenesan: I agree.
I missed it, GitHub's change review interface could use some work ;-)

src/modules/fstab/main.py
```diff
@@ -151,23 +151,23 @@ def generate_crypttab_line_info(self, partition):
...
+            password = "none"
 
-        # Set crypttab password for partition to none and remove crypttab options
         # if root partition was not encrypted
-        if any([p["mountPoint"] == "/"
-                and "luksMapperName" not in p
-                for p in self.partitions]):
-            password = "none"
+        if any([p["mountPoint"] == "/" and
+               "luksMapperName" not in p for p in self.partitions]):
```
adriaandegroot: I think it would be worth adding a helper function 

```
def is_unencrypted(partition):
    return "luksMapperName" not in partition

def is_partition_encrypted(mountpoint : str, partitions):
    return any(p["mountPoint"] == mountpoint and is_unencrypted(p) for p in self.partitions)
```

eugenesan: It sure would be nice but it won't reduce amount of code or simplify it.
In fact, in the future (when systemd-boot is the standard or grub gains full LUKS2 support) we might be able to drop 1/2 of the code.

Also, testing fstab module is a real pain.
Full installation for each possible setup is required after each change.
That's why my initial PR was so "ugly" but minimal.

If someone volunteers to testing, I am all for it.

dalto8 (approved):

eugenesan:
@adriaandegroot & @dalto8 
Thanks!
PR can be closed.

PR #2315, opened 2024-04-12
Mirror GRUB's splash screen functionality for non-grub booloaders

### LINKED ISSUE #2311: Feature request: module bootloader should check for plymouth and splash (similarly to grubcfg)

eugenesan:
Resolves https://github.com/calamares/calamares/issues/2311
Replaces https://github.com/calamares/calamares/pull/2313

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -127,7 +127,6 @@ def is_zfs_root(partition):
 
 def get_kernel_params(uuid):
     kernel_params = libcalamares.job.configuration.get("kernelParams", ["quiet"])
-    kernel_params.append("rw")
 
     partitions = libcalamares.globalstorage.value("partitions")
     swap_uuid = ""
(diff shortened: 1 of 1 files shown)


dalto8:
Why are we moving "rw"?  Can't we leave that part alone?

eugenesan:
It's for cosmetic reasons only.
'quite' is added automatically during cmdline initialization and if we do not move 'rw' we'll end up with `quite rw splash...`  which looks odd so I decided to move it.

adriaandegroot:
So it turns into `quiet splash rw` :shrug:  I'm going to merge this and then do some minor post-merge tidying of the module.

PR #2316, opened 2024-04-13
add possibility to pre check encryption checkox

SixK:
This PR add possibility to pre check Encryption checkbox when enableLuksAutomatedPartitioning is true.  
By default preCheckEncryption is false and has to be uncommented with true value for the checkbox to be pre checked.  
By default the option does not change actual behavior.

This option is mainly useful in enterprise to help users to not forget to cypher their disk.

### DIFF
--- a/src/modules/partition/PartitionViewStep.cpp
+++ b/src/modules/partition/PartitionViewStep.cpp
@@ -698,8 +698,6 @@ PartitionViewStep::setConfigurationMap( const QVariantMap& configurationMap )
                 Calamares::getBool( configurationMap, "alwaysShowPartitionLabels", true ) );
     gs->insert( "enableLuksAutomatedPartitioning",
                 Calamares::getBool( configurationMap, "enableLuksAutomatedPartitioning", true ) );
-    gs->insert( "preCheckEncryption",
-                Calamares::getBool( configurationMap, "preCheckEncryption", false ) );
 
     QString partitionTableName = Calamares::getString( configurationMap, "defaultPartitionTableType" );
     if ( partitionTableName.isEmpty() )
--- a/src/modules/partition/PartitionViewStep.cpp
+++ b/src/modules/partition/PartitionViewStep.cpp
@@ -698,6 +698,8 @@ PartitionViewStep::setConfigurationMap( const QVariantMap& configurationMap )
                 Calamares::getBool( configurationMap, "alwaysShowPartitionLabels", true ) );
     gs->insert( "enableLuksAutomatedPartitioning",
                 Calamares::getBool( configurationMap, "enableLuksAutomatedPartitioning", true ) );
+    gs->insert( "preCheckEncryption",
+                Calamares::getBool( configurationMap, "preCheckEncryption", false ) );
 
     QString partitionTableName = Calamares::getString( configurationMap, "defaultPartitionTableType" );
     if ( partitionTableName.isEmpty() )
--- a/src/modules/partition/gui/ChoicePage.cpp
+++ b/src/modules/partition/gui/ChoicePage.cpp
@@ -90,7 +90,7 @@ ChoicePage::ChoicePage( Config* config, QWidget* parent )
     auto gs = Calamares::JobQueue::instance()->globalStorage();
 
     m_enableEncryptionWidget = gs->value( "enableLuksAutomatedPartitioning" ).toBool();
-    m_preCheckEncryption = gs->value( "preCheckEncryption" ).toBool();
+    m_preCheckEncryption = m_config->preCheckEncryption();
 
     // Set up drives combo
     m_mainLayout->setDirection( QBoxLayout::TopToBottom );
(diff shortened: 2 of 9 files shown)


src/modules/partition/PartitionViewStep.cpp
```diff
@@ -698,6 +698,8 @@ PartitionViewStep::setConfigurationMap( const QVariantMap& configurationMap )
                 Calamares::getBool( configurationMap, "alwaysShowPartitionLabels", true ) );
     gs->insert( "enableLuksAutomatedPartitioning",
                 Calamares::getBool( configurationMap, "enableLuksAutomatedPartitioning", true ) );
+    gs->insert( "preCheckEncryption",
+                Calamares::getBool( configurationMap, "preCheckEncryption", false ) );
```
dalto8: The ChoicePage has access to the config.  I don't think we need to use global storage here.

dalto8 (changes requested):
Thanks for the contribution!

As mentioned in the specific comment, I think we can refactor and simplify if we avoid using global storage for this.

SixK:
You are probably right. 
In fact, I tried to keep the same logic than "enableLuksAutomatedPartitioning" variable handling.
gs->insert( "enableLuksAutomatedPartitioning",
                Calamares::getBool( configurationMap, "enableLuksAutomatedPartitioning", true ) );

dalto8:
I would have to look and see if that is used anywhere else to know if it makes sense for it to be in global storage.  However, even if it doesn't, we shouldn't duplicate bad code.

SixK:
You're definitly right, preCheckEncryption variable don't have to be stored on Global Storage since no other module has to know value selected/configured.
Commit cc835cee6f6272ef8692e88f3008f36fffe01800 should remove use of Global Storage and only use partition.conf file.

src/modules/partition/gui/ChoicePage.cpp
```diff
@@ -468,6 +470,11 @@ ChoicePage::onActionChanged()
         {
             m_encryptWidget->setFilesystem( FileSystem::typeForName( m_replaceFsTypesChoiceComboBox->currentText() ) );
         }
+        
+        if ( m_preCheckEncryption )
```
dalto8: Do we need `m_preCheckEncryption` still?  Can't we just call `m_config->preCheckEncryption()` here?

SixK: Your right, m_preCheckEncryption variable can be removed from ChoicePage files.
Modification done with commit 5ac23cd3d60f956fd5525b66c738e0178e07d491

dalto8:
Thanks, it generally looks good to me but can we clarify about what your intentions are for when you expect the box to become checked?

SixK:
Before doing anything, please wait, I have to find how to trigger stateChanged() when forcing checkbox to true. 
Next button is available even if passphrase is not set !?
I may need help to find a solution for this.


Now to reply to your question:
Encryption checkbox has to be only checked when preCheckEncryption is set to true and enableLuksAutomatedPartitioning is true too in config file. (when Encryption Widget is visible).

Purpose is to help people to not forget to cypher their disk when installing customized distribution in enterprise.

SixK:
Finally found a way to handle Next button for all combinations of enableLuksAutomatedPartitioning and preCheckEncryption variables.
I did few simplifications and tested that it introduced no regressions.

Maybe you will find better way to handle all this.

Here is a short table of problems encountered and solved:

enableLuksAutomatedPartitioning | preCheckEncryption | problem
-- | -- | --
X | X | Next button was active by default
X | - | Next button was still active if checking Encryption Checkbox for the first time (pb with encryption state)

src/modules/partition/gui/ChoicePage.cpp
```diff
@@ -1575,7 +1576,9 @@ ChoicePage::calculateNextEnabled() const
         }
     }
 
-    if ( m_config->installChoice() != InstallChoice::Manual && m_encryptWidget->isVisible() )
+    if ( m_config->installChoice() != InstallChoice::Manual 
+         && (m_encryptWidget->isVisible() ||
+                    m_encryptWidget->isEncryptionCheckboxChecked()))
```
dalto8, line 1581: This seems wrong to me.  If the encryption widget isn't visible, encryption should be disabled.

dalto8 (changes requested):

src/modules/partition/gui/ChoicePage.cpp
SixK: First time we go on this condition, we are at opening of the application.
Page is not visible so widget is not visible.
This is not a problem when Encryption is not checked (default behaviour), Next button will be accessible.
But if Encryption is pre Check, Next button will still be accessible, even if we have no
Encryption passphrase set. 
We can't let user to go to Next page until the passphrase is set or Encryption Widget is not check.

Only way I found to have the right behaviour in all combinations is to verify Encryption Checkbox at this point.

dalto8 (approved):

src/modules/partition/gui/ChoicePage.cpp
dalto8: OK, I tested this and the behavior seems fine.  Although, it is a little aggressive about rechecking that checkbox.

PR #2322, opened 2024-04-22
[partition] Don't crash when going back without selecting a partition

dalto8:
While testing something else I discovered an issue where if you choose "Install Alongside" and then hit the back button without choosing a partition, it causes a crash do to the assertions in `doAlongsideApply()`

### DIFF
--- a/src/modules/partition/gui/ChoicePage.cpp
+++ b/src/modules/partition/gui/ChoicePage.cpp
@@ -679,7 +679,11 @@ ChoicePage::onLeave()
 {
     if ( m_config->installChoice() == InstallChoice::Alongside )
     {
-        doAlongsideApply();
+        if ( m_afterPartitionSplitterWidget->splitPartitionSize() >= 0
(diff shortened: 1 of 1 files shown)


src/modules/partition/gui/ChoicePage.cpp
```diff
@@ -679,7 +679,11 @@ ChoicePage::onLeave()
 {
     if ( m_config->installChoice() == InstallChoice::Alongside )
     {
-        doAlongsideApply();
+        if ( m_afterPartitionSplitterWidget->splitPartitionSize() >= 0
+             && m_afterPartitionSplitterWidget->newPartitionSize() >= 0 )
+        {
+            doAlongsideApply();
+        }
```
adriaandegroot, line 686: I think it would be good to log something in the `else` case, so we (or the user) can know what happened, anyway.

adriaandegroot:
LGTM, with one suggestion. Feel free to merge at your leisure.

PR #2323, opened 2024-04-26
make first arg of accumulate func obj to be prvalue

vnepogodin:
since C++20 first argument to std::accumulate cannot be lvalue, before C++20 it could be lvalue.

see 2) in https://en.cppreference.com/w/cpp/algorithm/accumulate


adriaandegroot:
Calamares is C++17, so I'm hesitant to apply changes that are relevant for a later standard. Thank you for pointing them out, though -- in retrospect I think using `std::accumulate()` there is more of a smoke-screen than a useful application of the algorithm. So I'm rejecting this PR, but applying relevant fixes, instead.

PR #2341, opened 2024-06-03
Fix mounting encrypted swap partition

vlinkz:
#2102 allows for mounting a swap partition during installation, however luks encrypted swap partitions are not mounted. This change enables the mounting of luks and luks2 encrypted swap partitions.

### DIFF
--- a/src/modules/mount/main.py
+++ b/src/modules/mount/main.py
@@ -345,7 +345,8 @@ def run():
                 _("No partitions are defined for <pre>{!s}</pre> to use.").format("mount"))
 
     # Find existing swap partitions that are part of the installation and enable them now
-    swap_devices = [p["device"] for p in partitions if (p["fs"] == "linuxswap" and p.get("claimed", False))]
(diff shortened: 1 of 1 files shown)


dalto8:
Thanks for the contribution!

I like your idea but I think we have exceeded the readability of a one-liner there.  Can we either reformat it or refactor it to make it more obvious what it is doing?

vlinkz:
Sure! Perhaps something more like this?
```python
    # Find existing swap partitions that are part of the installation and enable them now
    swap_devices = []    
    for p in partitions:
        if p["fs"] == "linuxswap" and p.get("claimed", False) and p["fsName"] == "linuxswap":
            swap_devices.append(p["device"])
        elif p["fs"] == "linuxswap" and p.get("claimed", False) and p["fsName"] in ["luks", "luks2"]:
            swap_devices.append("/dev/mapper/" + p["luksMapperName"])
```

dalto8:
Yes, that is much more readable I think.

PR #2345, opened 2024-06-23
[partition] Fix crash on filesystem changes when encryption is checked

dalto8:
This fixes the current crash when selecting encryption and changing the filesystem.

There is one remaining issue which is if the `preCheckEncryption` conf option is set to `true` AND you select erase disk AND you try to uncheck the encryption checkbox, you need to click it twice.

Ideally this would be fixed but since the current situation is an application crash on a fairly common use case and the above is an edge case, we should probably merge this unless someone has time to track that last bit down.


SixK:
I have to check again on a clean install, 
but the fix seem's to introduce a regression.

When preCheckEncryption is not set and checking encryption checkbox next button is still available. 
This let people leaving blank passphrase.

When preCheckEncryption is set to true and unchecking encryption checkbox, encryption checkbox become checked again.
uncheking twice, let the checkbox unchecked, but next button is still unavailable.

Issue #2371, opened 2024-09-17
LANG env not set with selected locale


SixK:
When Calamares set grub.cfg in bootloader module, 
it seem's that $LANG variable is not set with selected locale.

chroot take LANG from host env.
In my case, I run Calamares with a custom Ubuntu 24.04 cdrom iso using default $LANG that is "LANG=C.UTF-8".
As the result, grub-mkconfig don't set lang section (and locale) in grub.cfg

To set grub.cfg content, grub-mkconfig use "/etc/grub.d/00-header" file containing this snippets:
```
grub_lang=`echo $LANG | cut -d . -f 1`
# Gettext variables and module
if [ "x${LANG}" != "xC" ] &&  [ "x${LANG}" != "x" ]; then
  cat << EOF
  set locale_dir=\$prefix/locale
  set lang=${grub_lang}
  insmod gettext
EOF
fi
```

Final result is that grub is not using selected language and show a "IE" keyboard (Ireland instead of "FR" in my case),
when unlocking partitions at boot. 
That may be confusing even if unlock partition is using right keyboard layout.

Is there a quick way to fix this behavior using any calamares or calamares modules .cfg files ?
Would it be a good idea to change env LANG value on the host side when language is selected by user in calamares ?
Would it be a good idea to do something like check_target_env_call(["env", f"LANG={libcalamares.globalstorage.value('localeConf.LANG')}",libcalamares.job.configuration["grubMkconfig"], "-o", output_file]) ?
Is there something better to do ?

demmm:
What version of Calamares is this referring to? (please fill out all asked for info when submitting)

SixK:
Calamares base version I use is 3.3.5

SixK:
Found in documentation here https://github.com/calamares/calamares/blob/calamares/src/modules/README.md#running-commands-in-python

``` 
    target_env_call(command, stdin, timeout) returns the exit code, does not raise.
    check_target_env_call(command, stdin, timeout) raises on a non-zero exit code.
    check_target_env_output(command, stdin, timeout) returns a single string with the output of command, raises on a non-zero exit code.

All of the API functions for running commands set the environment LC_ALL and LANG to "C" for the called command.```

This mean that `check_target_env_call([libcalamares.job.configuration["grubMkconfig"], "-o", output_file])` from https://github.com/calamares/calamares/blob/calamares/src/modules/bootloader/main.py
can't set grub locales and add lang lines in grub.cfg

SixK:
[...]

PR #2372, opened 2024-09-18
Should fix grub not adding lang section in /boot/grub/grub.cfg file as

SixK:
chroot take host env LANG variable and this value is usually LANG=C.UTF-8 when booting a live installer.

Pb is only visible at boot when prompted to unlock Luks partition. 
In this case, screen show a IE keymap instead of keymap of LANG selected by user in Calamares.

Note that in this case, on ubuntu,  a "setupcon --save-only" and "update-initramfs -u" may be missing to generate /etc/console-setup/cached_UTF-8_del.kmap.gz and make this file integrated in initramfs to be able to unlock Luks partitions with the keymap selected in calamares.

This quick fix is related to https://github.com/calamares/calamares/issues/2371

### DIFF
--- a/src/modules/localecfg/main.py
+++ b/src/modules/localecfg/main.py
@@ -133,6 +133,9 @@ def run():
             'LC_IDENTIFICATION': 'en_US.UTF-8'
         }
 
+    # Should fix grub lang (set as IE at boot when not set) and locales in /boot/grub/grub.cfg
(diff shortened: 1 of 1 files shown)


SixK:
Cancel this PR, seem keyboard seen at start to unlock luks partition is not related to grub lang.
I have to investigate more.

PR #2376, opened 2024-09-25
Partition module fixes for encryption checkbox

### LINKED ISSUE #2375: After checking "Encrypt system", the user can proceed past the Partitions screen without entering a passphrase
**Describe the bug**
If the user checks "Encrypt System", the "Next" button remains enabled and they can proceed past the partitions screen without entering a passphrase. Unchecking and rechecking the box results in the "Next" button being grayed out as expected.

**To Reproduce**
1. Launch Calamares.
2. Click "Next" until you get to the partitions screen.
3. Click "Erase Disk".
4. Check the "Encrypt system" checkbox.
5. Click "Next" (do not enter a passphrase first).

**Expected behavior**
"Next" button should be disabled and unable to be clicked.

**Screenshots and Logs**
Screenshot_20240925_175929

**Additional context**
I debugged this issue using gdb and traced it to a logic error in `src/modules/partition/gui/EncryptWidget.cpp`:

```c++
    // In function EncryptWidget::updateState( const bool notify ):
    // ...
    if ( newState != m_state )
    {   
        m_state = newState;
        if ( notify ) 
        {   
            Q_EMIT stateChanged( m_state );
        }       
    }
```
The `if ( newState != m_state )` is causing the issue.

In the constructor, `m_state` is initialized to `Encryption::Disabled`, but when the constructor calls `updateState()` in EncryptWidget.cpp line 68, `m_state` gets changed to `Encryption::Unconfirmed`. This is because `updateState()` is called while the "Encrypt system" checkbox is still invisible, thus triggering the following code path:

```c++
EncryptWidget::Encryption
EncryptWidget::state() const
{
    Encryption newState = Encryption::Unconfirmed;

    // This `if` statement is entered because the "Encrypt system" check box is invisible since the app is still starting up
    if ( m_ui->m_encryptCheckBox->isChecked() || !m_ui->m_encryptCheckBox->isVisible() )
    {
        // And naturally, the passphrase text boxes are also empty, so this `if` statement will be skipped over...
        if ( !m_ui->m_passphraseLineEdit->text().isEmpty()
             && m_ui->m_passphraseLineEdit->text() == m_ui->m_confirmLineEdit->text() )
        {
            newState = Encryption::Confirmed;
        }
        else // ...and this one is run instead, thus returning `Encryption::Unconfirmed` which is what m_state is then set to.
        {
            newState = Encryption::Unconfirmed;
        }
    }
    else
    {
        newState = Encryption::Disabled;
    }

    return newState;
}
```

So ultimately `m_state` ends up being set to `Encryption::Unconfirmed` once the encryption widget's constructor is finished. Then when the user checks the "Encrypt system" checkbox later on, `EncryptWidget::state()` properly determines that the state is `Encryption::Unconfirmed`, but because `m_state` is already `Encryption::Unconfirmed`, it does **not** emit a `stateChanged` signal. This means that `ChoicePage::onEncryptWidgetStateChanged` is not called when the user first checks the checkbox, and thus the "Next" button is not disabled.

Simply removing the `if ( newState != m_state )` statement from EncryptWidget.cpp should be enough to resolve the bug.

ArrayBolt3:
Fixes https://github.com/calamares/calamares/issues/2375, https://github.com/calamares/calamares/issues/2379

### DIFF
--- a/src/modules/partition/gui/ChoicePage.cpp
+++ b/src/modules/partition/gui/ChoicePage.cpp
@@ -1031,9 +1031,10 @@ ChoicePage::updateActionChoicePreview( InstallChoice choice )
         if ( m_enableEncryptionWidget )
         {
             m_encryptWidget->show();
-            if ( m_config->preCheckEncryption() )
+            if ( m_config->preCheckEncryption() && ! m_preCheckActivated )
             {
                 m_encryptWidget->setEncryptionCheckbox( true );
+                m_preCheckActivated = true;
             }
         }
         m_previewBeforeLabel->setText( tr( "Current:", "@label" ) );
@@ -1090,9 +1091,10 @@ ChoicePage::updateActionChoicePreview( InstallChoice choice )
         if ( shouldShowEncryptWidget( choice ) )
         {
             m_encryptWidget->show();
-            if ( m_config->preCheckEncryption() )
+            if ( m_config->preCheckEncryption() && ! m_preCheckActivated )
             {
                 m_encryptWidget->setEncryptionCheckbox( true );
+                m_preCheckActivated = true;
             }
         }
         m_previewBeforeLabel->setText( tr( "Current:", "@label" ) );
--- a/src/modules/partition/gui/ChoicePage.h
+++ b/src/modules/partition/gui/ChoicePage.h
@@ -173,6 +173,7 @@ private:
     QString m_osproberOneEntryName;
 
     bool m_enableEncryptionWidget = false;
+    bool m_preCheckActivated = false;
 
     QMutex m_coreMutex;
 };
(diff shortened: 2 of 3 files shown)


ArrayBolt3:
Test plan: Ensure UI reacts as expected, system installation succeeds, and the installed system is bootable, in all of the following scenarios:

* `[x]` Erase disk, no encryption, pre-encrypt off, BIOS firmware
* `[x]` Erase disk, encryption, pre-encrypt off, UEFI firmware
* `[x]` Erase disk, no encryption, pre-encrypt on, UEFI firmware
* `[x]` Erase disk, encryption, pre-encrypt on, BIOS firmware
* `[x]` Install alongside, no encryption, pre-encrypt off, BIOS firmware
* `[x]` Install alongside, no encryption, pre-encrypt on, UEFI firmware
* `[x]` Replace partition, no encryption, pre-encrypt off, BIOS firmware
* `[x]` Replace partition, encryption, pre-encrypt off, UEFI firmware
* `[x]` Replace partition, no encryption, pre-encrypt on, UEFI firmware
* `[x]` Replace partition, encryption, pre-encrypt on, BIOS firmware

PR #2390, opened 2024-10-31
Add include keyword for unpackfs

tsimonq2:
This is used for stacked squashfses, where the user can select minimal or standard and not have to remove packages later.

I have tested this, it works.


adriaandegroot:
Post-merge, I changed the keyword to *condition* because I found it confusing when used alongside *exclude* which becomes an rsync option.

PR #2392, opened 2024-11-02
[partition] Use the same last sector for root vs. swap partitions

### LINKED ISSUE #2367: automatic partition task fail with swap
> Hi! Thank you for helping improve Calamares. If you are seeing a problem in installing a specific distribution, you should **probably** report the problem in the distribution's bug tracker, first. That helps filter out issues with packaging, mis-configuration, etc. that Calamares has no control over. If you are a distribution packager or maintainer, this page is for you.

**Describe the bug**
If i choose automatic partitioning, e.g. erase disk, use swap "no hibernate", calamares install fails in creating partitions module.  Without swap or swapfile all is working as it should

**To Reproduce**
1. Start Calamares
2. Choose erase disk, with swap
3. Go fourther
4. Calamares Install will fail

**Expected behavior**
Install should be successfull.

**Screenshots and Logs**
```
Create a new partition (3.17 GiB, linuxswap) on ‘/dev/nvme0n1’


Job: Create new partition on device ‘/dev/nvme0n1’


Command: sfdisk --force --append /dev/nvme0n1

Failed to add partition ‘New Partition’ to device ‘/dev/nvme0n1’.

Failed to add partition ‘New Partition’ to device ‘/dev/nvme0n1’.

```
**Additional context**

Used Systems:  \

KDE neon User

```
TUXEDO OS (Ubuntu 22.04 with neon user packages)
Operating System: TUXEDO OS 3
KDE Plasma Version: 6.1.5
KDE Frameworks Version: 6.5.0
Gear Version: 24.08.0
Qt Version: 6.7.2
Kernel Version: 6.11.0-100005-tuxedo (64-bit)
Graphics Platform: X11
Processors: 6 × AMD Ryzen 7 2700X Eight-Core Processor
Memory: 7.7 GiB of RAM
Graphics Processor: virgl
Manufacturer: TUXEDO
Product Name: Tuxedo Qemu VM
```

jpetso:
If we have already determined that `lastSectorForRoot` should be a lower value for GPT tables, the calculation of swap partition boundaries shouldn't keep using `dev->totalLogical()` for its own size.

Fixes #2367

### DIFF
--- a/src/modules/partition/core/PartitionActions.cpp
+++ b/src/modules/partition/core/PartitionActions.cpp
@@ -108,7 +108,7 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
         partType = isEfi ? PartitionTable::gpt : PartitionTable::msdos;
     }
     // last usable sector possibly allowing for secondary GPT using 66 sectors (256 entries)
-    qint64 lastSectorForRoot = dev->totalLogical() - (partType == PartitionTable::gpt ? 67 : 1);
+    const qint64 lastUsableSector = dev->totalLogical() - ( partType == PartitionTable::gpt ? 67 : 1 );
 
     // Looking up the defaultFsType (which should name a filesystem type)
     // will log an error and set the type to Unknown if there's something wrong.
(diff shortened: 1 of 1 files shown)


jpetso:
Carlos of KDE neon built a Calamares .deb for me with this patch, and it looks like this fixes the creation of the swap partition. My plasmashell crash after `sfdisk` invocation is probably unrelated, but perhaps it would be a good idea to have someone else try out this patch before merging.

For testing, select "Erase disk" and "Swap (with Hibernate)" to exercise this code path. [Edit: I should probably mention this context in the commit message itself.]

demmm:
Tested this patch with latest kpmcore build without the reverting patch.
Used install on virtualbox was erase disk, with swap (no hibernate & hibernate), XFS filesystem, no longer any crashes/segfaults.

demmm (approved):

PR #2398, opened 2024-11-12
Update calamares_zh_CN.ts

nekrozis:
Corrected zh_CN typo


demmm:
Normally, Calamares uses transifex for translations:

So, if possible, it would be better to work there though it might be (as an exception), this PR will be merged.

nekrozis:
OK, I will fix it there and close this PR

nekrozis:
I found it a bit of a hindrance to register an account on the site, as it requires a business email address to do so.

adriaandegroot:
I have merged and also applied the change in Transifex. I see a couple more reasons to start moving away from Transifex, so I think we should soon change the documentation to name something else as preferred place for translations.

Issue #2397, opened 2024-11-10
Allow distros to restrict what filesystems can be used in manual partitioning


ArrayBolt3:
**Is your feature request related to a problem? Please describe.**
Calamares currently allows one to pick from a large number of different filesystems when setting up a system with manual partitioning. For the most part this is good, however because not all filesystems work well with all distros, it is possible for the user to shoot themselves in the foot by picking a filesystem that won't work. For instance, currently nothing in Calamares prevents the user from attempting to install a Debian-based distro like Kicksecure onto a FAT32-formatted root partition.

![Screenshot_20241110_173418](https://github.com/user-attachments/assets/16f77507-654b-4754-afcc-e99b2950f4f7)

Obviously this won't work in the long run, however the installation actually tries to proceed and gets quite far. The error message that the installation eventually fails with is less-than-obvious:

![Screenshot_20241110_173630](https://github.com/user-attachments/assets/e3dcd4f7-6fa1-4e04-bc64-69951644ef8a)

**Describe the solution you'd like**
I think the best solution here would be to allow distros to specify what filesystems can be used at what mountpoints. Ideally, the manual partitioning screen would prevent the user from selecting a "bad" filesystem, however it might be sufficient to simply warn them that what they're doing is likely to fail in weird ways if they continue. Calamares could use a default "deny-map" (or whatever you want to call it) that would prevent some common blunders (such as using a non-UNIX filesystem for the root filesystem), but the map would be fully configurable by the distro developer.

**Describe alternatives you've considered**
The simplest solution would be to simply prohibit using fat32 as a filesystem for the root filesystem, however this doesn't seem like the right decision to make since in theory some distros might actually want to support a fat32 root filesystem.

**Additional context**
The issue mentioned above has tripped up at least one user in the wild: https://forums.kicksecure.com/t/kicksecure-installation-cannot-set-timezone-link-creation-failed-target-usr-share-timezone-link-name-etc-localtime/652

dalto8:
Yeah, this would be great.  It isn't only non-working filesystems that are the issue here.  It is filesystems the ISO or target system doesn't have the proper utilities for installed.

We still need to ship an ISO with a ton of filesystem utilities on the ISO because it can't be restricted.

ArrayBolt3:
If this would be worthwhile, I can spend time trying to implement a proof-of-concept. Any suggestions or pointers on how the implementation should work?

dalto8:
Hmm...I haven't given it a lot of thought.

[...]

PR #2400, opened 2024-11-21
Allow restricting the filesystem's usable in manual partitioning

ArrayBolt3:
This is my first attempt at implementing https://github.com/calamares/calamares/issues/2397. I tried to follow the existing conventions when it came to how configuration was fetched, making a new class for the map of restricted filesystems for mountpoints and handling default configuration values in much the same way as other parts of the code handled them. I don't like the way I had to plumb PartitionCoreModule and Config objects into areas of the code that didn't used to use them, but it was the only way I could see to do the job and still be consistent.

There are some major shortcomings still left in the code:

* <s>For mountpoint-specific filesystem whitelists, I'm not taking into account that un-mountable filesystems should always be allowed, even though I document that this is taken into account. This **must** be fixed before this can be merged.</s> Resolved.
* <s>It doesn't take into account the case when mountpoint `/a` can be filesystem X, mountpoint `/a/b` cannot be filesystem X, and the user chooses to only make a partition for mountpoint `/a` and leave `/a/b` on the same filesystem as `/a`. In that instance the user could set `/a` to be filesystem X, and thus `/a/b` would also be filesystem X, despite that being prohibited.</s> Resolved, the UX isn't awesome but it does work.
* <s>No testing has been done yet, at all. I don't even know if Calamares will launch with this.</s> Did some testing, it's working decent so far.

I'm not sure if this is the right approach to this, or if I should be approaching the problem in a fundamentally different way. I'm happy to scrap the patch and try again if this isn't the right way to do this.

EDIT: The code is mostly working now. It still needs a good test plan written and thoroughly tested before it's merge-ready, but it seems to work right, and is ready for review.

### DIFF
--- a/src/modules/partition/core/DirFSRestrictLayout.cpp
+++ b/src/modules/partition/core/DirFSRestrictLayout.cpp
@@ -1,0 +1,234 @@
+/* === This file is part of Calamares - <https://calamares.io> ===
+ *
+ *   SPDX-FileCopyrightText: 2014-2017 Teo Mrnjavac <teo@kde.org>
+ *   SPDX-FileCopyrightText: 2017-2018 Adriaan de Groot <groot@kde.org>
+ *   SPDX-FileCopyrightText: 2018-2019 Collabora Ltd <arnaud.ferraris@collabora.com>
+ *   SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ *   Calamares is Free Software: see the License-Identifier above.
+ *
+ */
+
+#include "utils/Logger.h"
+
+#include "core/DirFSRestrictLayout.h"
+
+#include "core/KPMHelpers.h"
+#include "core/PartUtils.h"
+
+#include "utils/Variant.h"
+
+#include "GlobalStorage.h"
+#include "JobQueue.h"
+
+#include <kpmcore/fs/filesystem.h>
+#include <kpmcore/fs/filesystemfactory.h>
+
+#include <QSet>
+
+DirFSRestrictLayout::DirFSRestrictLayout() {}
+
+DirFSRestrictLayout::DirFSRestrictLayout( const DirFSRestrictLayout& layout )
+    : m_dirFSRestrictLayout( layout.m_dirFSRestrictLayout )
+{
+}
+
+DirFSRestrictLayout::~DirFSRestrictLayout() {}
+
+DirFSRestrictLayout::DirFSRestrictEntry::DirFSRestrictEntry()
+{
+}
+
+DirFSRestrictLayout::DirFSRestrictEntry::DirFSRestrictEntry( const QString& path,
+                                                             QList< FileSystem::Type > allowedFSTypes,
+                                                             bool onlyWhenMountpoint )
+    : dirPath( path )
+    , dirAllowedFSTypes( allowedFSTypes )
+    , useOnlyWhenMountpoint( onlyWhenMountpoint )
+{
+}
+
+void
+DirFSRestrictLayout::init( const QVariantList& config )
+{
+    m_dirFSRestrictLayout.clear();
+    bool efiNeedsSet = true;
+
+    for ( const auto& r : config )
+    {
+        QVariantMap pentry = r.toMap();
+        if ( !pentry.contains( "directory" ) || !pentry.contains( "allowedFilesystemTypes" ) )
+        {
+            cError() << "Directory filesystem restriction layout entry #" << config.indexOf( r )
+                     << "lacks mandatory attributes, switching to default layout.";
+            m_dirFSRestrictLayout.clear();
+            break;
+        }
+
+        QString directory = Calamares::getString( pentry, "directory" );
+        QStringList allowedFSTypeStrings = Calamares::getStringList( pentry, "allowedFilesystemTypes" );
+        QList< FileSystem::Type > allowedFSTypes;
+        if ( allowedFSTypeStrings.length() == 1 && allowedFSTypeStrings[0] == "all" )
+        {
+            allowedFSTypes = fullFSList();
+        }
+        else
+        {
+            for ( const auto& fsStr : allowedFSTypeStrings )
+            {
+                FileSystem::Type allowedFSType;
+                PartUtils::canonicalFilesystemName( fsStr, &allowedFSType );
+                if ( allowedFSType == FileSystem::Type::Unknown )
+                {
+                    continue;
+                }
+               allowedFSTypes.append( allowedFSType );
+            }
+        }
+        bool onlyWhenMountpoint = Calamares::getBool( pentry, "onlyWhenMountpoint", false );
+        if ( directory == "efi" )
+        {
+            efiNeedsSet = false;
+        }
+        DirFSRestrictEntry restrictEntry( directory, allowedFSTypes, onlyWhenMountpoint );
+        m_dirFSRestrictLayout.append( restrictEntry );
+    }
+
+    if ( efiNeedsSet )
+    {
+        QList< FileSystem::Type > efiAllowedFSTypes = { FileSystem::Fat32 };
+        DirFSRestrictEntry efiRestrictEntry( "efi", efiAllowedFSTypes, true );
+        m_dirFSRestrictLayout.append( efiRestrictEntry );
+    }
+}
+
+QList< FileSystem::Type >
+DirFSRestrictLayout::allowedFSTypes( const QString& path, const QStringList& existingMountpoints, bool overlayDirs )
+{
+    QSet< FileSystem::Type > typeSet;
+    bool foundTypeList = false;
+
+    for ( const auto& entry : m_dirFSRestrictLayout )
+    {
+        QString dirPath = entry.dirPath;
+        if ( dirPath == "efi" )
+        {
+            dirPath = Calamares::JobQueue::instance()->globalStorage()->value( "efiSystemPartition" ).toString();
+        }
+        if ( dirPath == path || ( !entry.useOnlyWhenMountpoint && overlayDirs && path.startsWith( QStringLiteral( "/" ) ) && dirPath.startsWith( path ) && !existingMountpoints.contains( dirPath ) ) )
+        {
+            QSet< FileSystem::Type > newTypeSet = QSet< FileSystem::Type >( entry.dirAllowedFSTypes.cbegin(), entry.dirAllowedFSTypes.cend() );
+            foundTypeList = true;
+            if ( typeSet.isEmpty() )
+            {
+                typeSet = newTypeSet;
+                if ( !overlayDirs )
+                {
+                    break;
+                }
+            }
+            else
+            {
+                typeSet.intersect( newTypeSet );
+            }
+        }
+    }
+
+    if ( overlayDirs )
+    {
+        QList< FileSystem::Type > anyTypeList = anyAllowedFSTypes();
+        QSet< FileSystem::Type > anyTypeSet = QSet< FileSystem::Type >( anyTypeList.cbegin(), anyTypeList.cend() );
+        if ( !foundTypeList )
+        {
+            typeSet = anyTypeSet;
+            foundTypeList = true;
+        }
+        else
+        {
+            typeSet.intersect( anyTypeSet );
+        }
+    }
+
+    if ( foundTypeList )
+    {
+        return QList< FileSystem::Type >( typeSet.cbegin(), typeSet.cend() );
+    }
+    else
+    {
+        // This directory doesn't have any allowed filesystems explicitly
+        // configured, so all filesystems are valid.
+        return fullFSList();
+    }
+}
+
+QString
+DirFSRestrictLayout::diagnoseFSConflict( const QString& path, const FileSystem::Type& fsType, const QStringList& existingMountpoints )
+{
+    QSet< FileSystem::Type > typeSet;
+    bool foundTypeList = false;
+
+    for ( const auto& entry : m_dirFSRestrictLayout )
+    {
+        QString dirPath = entry.dirPath;
+        if ( dirPath == "efi" )
+        {
+            dirPath = Calamares::JobQueue::instance()->globalStorage()->value( "efiSystemPartition" ).toString();
+        }
+        if ( dirPath == path || ( !entry.useOnlyWhenMountpoint && path.startsWith( QStringLiteral( "/" ) ) && ( dirPath.startsWith( path ) || dirPath == "any" ) && !existingMountpoints.contains( dirPath ) ) )
+        {
+            QSet< FileSystem::Type > newTypeSet = QSet< FileSystem::Type >( entry.dirAllowedFSTypes.cbegin(), entry.dirAllowedFSTypes.cend() );
+            foundTypeList = true;
+            if ( typeSet.isEmpty() )
+            {
+                typeSet = newTypeSet;
+            }
+            else
+            {
+                typeSet.intersect( newTypeSet );
+            }
+        }
+
+        if ( foundTypeList && !typeSet.contains( fsType ) )
+        {
+            if ( typeSet.isEmpty() )
+            {
+                cWarning() << "no filesystems are valid for path '" << path << "', check directoryFilesystemRestrictions for issues";
+            }
+            // At this point, we've found the first mountpoint that, when
+            // taken into account, results in the currently chosen filesystem
+            // being invalid. Return that mountpoint.
+            return dirPath;
+        }
+    }
+
+    return QString();
+}
+
+QList< FileSystem::Type >
+DirFSRestrictLayout::anyAllowedFSTypes()
+{
+    for ( const auto& entry : m_dirFSRestrictLayout )
+    {
+        if ( entry.dirPath == "any" )
+        {
+            return entry.dirAllowedFSTypes;
+        }
+    }
+
+    // No global filesystem whitelist defined, so all filesystems are
+    // considered valid unless a mountpoint-specific whitelist is used to
+    // restrict the allowed filesystems.
+    return fullFSList();
+}
+
+QList< FileSystem::Type >
+DirFSRestrictLayout::fullFSList()
+{
+    QList< FileSystem::Type > typeList;
+    FileSystemFactory::init();
+    for ( auto fs : FileSystemFactory::map() )
+    {
+        typeList.append( fs->type() );
+    }
+    return typeList;
+}
--- a/src/modules/partition/core/DirFSRestrictLayout.cpp
+++ b/src/modules/partition/core/DirFSRestrictLayout.cpp
@@ -36,10 +36,6 @@ DirFSRestrictLayout::DirFSRestrictLayout( const DirFSRestrictLayout& layout )
 
 DirFSRestrictLayout::~DirFSRestrictLayout() {}
 
-DirFSRestrictLayout::DirFSRestrictEntry::DirFSRestrictEntry()
-{
-}
-
 DirFSRestrictLayout::DirFSRestrictEntry::DirFSRestrictEntry( const QString& path,
                                                              QList< FileSystem::Type > allowedFSTypes,
                                                              bool onlyWhenMountpoint )
--- a/src/modules/partition/core/DirFSRestrictLayout.cpp
+++ b/src/modules/partition/core/DirFSRestrictLayout.cpp
@@ -3,6 +3,7 @@
  *   SPDX-FileCopyrightText: 2014-2017 Teo Mrnjavac <teo@kde.org>
  *   SPDX-FileCopyrightText: 2017-2018 Adriaan de Groot <groot@kde.org>
  *   SPDX-FileCopyrightText: 2018-2019 Collabora Ltd <arnaud.ferraris@collabora.com>
+ *   SPDX-FileCopyrightText: 2024 Aaron Rainbolt <arraybolt3@gmail.com>
  *   SPDX-License-Identifier: GPL-3.0-or-later
  *
  *   Calamares is Free Software: see the License-Identifier above.
--- a/src/modules/partition/core/DirFSRestrictLayout.h
+++ b/src/modules/partition/core/DirFSRestrictLayout.h
@@ -2,6 +2,7 @@
  *
  *   SPDX-FileCopyrightText: 2018-2019 Collabora Ltd <arnaud.ferraris@collabora.com>
  *   SPDX-FileCopyrightText: 2019 Adriaan de Groot <groot@kde.org>
+ *   SPDX-FileCopyrightText: 2024 Aaron Rainbolt <arraybolt3@gmail.com>
  *   SPDX-License-Identifier: GPL-3.0-or-later
  *
  *   Calamares is Free Software: see the License-Identifier above.
--- a/src/modules/partition/core/DirFSRestrictLayout.h
+++ b/src/modules/partition/core/DirFSRestrictLayout.h
@@ -29,10 +29,10 @@ public:
     {
         QString dirPath;
         QList< FileSystem::Type > dirAllowedFSTypes;
-        bool useOnlyWhenMountpoint;
+        bool useOnlyWhenMountpoint = false;
 
         /// @brief All-zeroes DirFSRestrictEntry
-        DirFSRestrictEntry();
+        DirFSRestrictEntry() = default;
         /** @brief Parse @p path, @p allowedFSTypes, and @p onlyWhenMountpoint to their respective member variables
          *
          * Sets a specific set of allowed filesystems for a mountpoint.
--- a/src/modules/partition/core/DirFSRestrictLayout.h
+++ b/src/modules/partition/core/DirFSRestrictLayout.h
@@ -1,0 +1,86 @@
+/* === This file is part of Calamares - <https://calamares.io> ===
+ *
+ *   SPDX-FileCopyrightText: 2018-2019 Collabora Ltd <arnaud.ferraris@collabora.com>
+ *   SPDX-FileCopyrightText: 2019 Adriaan de Groot <groot@kde.org>
+ *   SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ *   Calamares is Free Software: see the License-Identifier above.
+ *
+ */
+
+#ifndef DIRFSRESTRICTLAYOUT_H
+#define DIRFSRESTRICTLAYOUT_H
+
+#include "Config.h"
+
+// KPMcore
+#include <kpmcore/fs/filesystem.h>
+
+// Qt
+#include <QList>
+#include <QObject>
+#include <QVariantMap>
+
+class DirFSRestrictLayout
+{
+public:
+    struct DirFSRestrictEntry
+    {
+        QString dirPath;
+        QList< FileSystem::Type > dirAllowedFSTypes;
+        bool useOnlyWhenMountpoint;
+
+        /// @brief All-zeroes DirFSRestrictEntry
+        DirFSRestrictEntry();
+        /** @brief Parse @p path, @p allowedFSTypes, and @p onlyWhenMountpoint to their respective member variables
+         *
+         * Sets a specific set of allowed filesystems for a mountpoint.
+         */
+        DirFSRestrictEntry( const QString& path,
+                            QList< FileSystem::Type > allowedFSTypes,
+                            bool onlyWhenMountpoint );
+        /// @brief Copy DirFSRestrictEntry
+        DirFSRestrictEntry( const DirFSRestrictEntry& e ) = default;
+    };
+
+    DirFSRestrictLayout();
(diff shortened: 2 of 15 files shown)


dalto8:
I haven't reviewed the code in detail yet.  But instead of creating a separate array to track the mountpoint settings, couldn't we just add an element to the `partitionLayout` array?

ArrayBolt3:
> I haven't reviewed the code in detail yet. But instead of creating a separate array to track the mountpoint settings, couldn't we just add an element to the `partitionLayout` array?

@dalto8 Yeah, that would work. I was trying to mimic what Calamares already did with `availableFileSystemTypes` and `partitionLayout`, but I'm happy to go with the other option. It would make the code a *lot* simpler in some places, so I think that's a great idea actually.

ArrayBolt3:
Well, that took longer than expected. Took me about an hour to figure out I was missing a `FileSystemFactory::init();` call at one point, but I figured it out. This now just uses a special node in the main whitelist map, rather than using a separate list for the global filesystem whitelist. I've tested it a bit, and it seems to work.

ArrayBolt3:
Small update, I hit a bug where this patch wouldn't let me use a filesystem that was prohibited on `/` even if I had no mount point selected at all and was trying to make a filesystem that wasn't intended to be mounted. This was due to a string matching error, which is now fixed.

ArrayBolt3:
Test plan to ensure the changes work as expected:

* [x] When a `mountpointFilesystemRestrictions` entry is missing `mountpoint`, ignores section entirely and falls back to defaults?
* [x] When a `mountpointFilesystemRestrictions` entry is missing `allowedFilesystemTypes`, ignores section entirely and falls back to defaults?
* [x] Unrecognized filesystems are ignored and don't cause problems?
* [x] Leaving out the `any` entry in `mountpointFilesystemRestrictions` results in all filesystems being permissible in general?
* [x] Not allowing a filesystem for a specific mountpoint results in being unable to proceed with manual partitioning if the prohibited filesystem is specified for that mountpoint?
* [x] Not allowing a filesystem in general (by setting the `any` entry and restricting the allowable filesystems there) results in being unable to proceed with manua partitioning if the prohibited filesystem is specified anywhere?
* [x] Adding a filesystem to the list of allowable filesystems for a mountpoint but leaving it out of the list of allowable filesystems in the `any` entry results in that filesystem being prohibited even for that mountpoint?
* [x] Set the mount point `/` to allow filesystems ext4, xfs, and btrfs, but set `/home` to only allow ext4 and xfs. Cannot create a `/` with btrfs until a separate `/home` partition with ext4 or xfs is created first?
* [x] Set the mount point `/` to allow filesystems ext4, xfs, and btrfs, set `/home` to only allow ext4 and xfs, and set `propagate = false` for `/home`. Can create `/` with btrfs even without creating a separate `/home` partition first?
* [x] Set the mount point `/` to allow filesystems ext4, xfs, and btrfs, set `/home` to only allow ext4 and xfs, and set `propagate = false` for `/home`. Cannot create separate `/home` with btrfs?
* [x] Set the mount point `/` to allow filesystems ext4, xfs, and btrfs, set `/home` to only allow ext4 and xfs, and set `propagate = false` for `/home`. Cannot create a `/home` partition with btrfs?
* [x] Set the mount point `/` to allow filesystems ext4, xfs, and btrfs, but set `/home` to only allow ext4 and xfs. Do *NOT* set the `any` entry at all. Can create an `/opt` partition with minix filesystem?
* [x] Filesystem checks are done both when changing mountpoint *and* when changing filesystem, for both partition creation and partition edit dialogs?
* [x] Set the mount point `/` to allow filesystems ext4, xfs, and btrfs. Create a partition, set the mountpoint as `/`. and then select an unmountable partition type such as `unformatted`. No denial message shown, can create partition?
* [x] Install a distro on an EFI machine with default filesystem restrictions, creating a single fat32 partition at `/boot/efi` and a single ext4 partition at `/`. Installation succeeds?
* [x] Code builds against both Qt5 and Qt6?

dalto8:
I am little confused with some of the recent changes.

What I was proposing above is that the filesystem-specific limitations be moved inside the existing `partitionLayout` key.  However, thinking back on it, that probably doesn't make sense so this is OK.

It seems what we actually did was remove the global filesystem restriction key which I think was important.  Many distros will only want to globally limit filesystems not specify per-partition limits.

Lastly, I don't understand the propagate example.  In the context of Calamares, how can two mountpoints share the same filesystem?  Btrfs and zfs have special handling so the mountpoints don't even apply in those cases.

EDIT: I see what you are doing with propagate I think.  However, this seems very strange to me.  These settings should only apply to mountpoints, not to directories inside those mountpoints.

ArrayBolt3:
@dalto8 Sorry, I think I must have misunderstood you.

I don't think these options can be combined into the `partitionLayout` key. `partitionLayout` specifies the layout for automatic partitioning, not manual partitioning. Having `partitionLayout` handle both scenarios results in the awkward situation of being unable to restrict manual partitioning settings without also creating new partitions when automatic partitioning is used. For instance, a distro might restrict `/boot/efi` to only using the fat32 filesystem. Not all installations will *have* a `/boot/efi` partition (notably those that are on legacy BIOS hardware), but for those that do, this restriction is important. Having this setting exist in the `partitionLayout` key means that now you have to specify a separate `/boot/efi` partition in the automatic partitioning settings just to be able to restrict the filesystems that can be used there, meaning *all* installations will now have a `/boot/efi` partition, even those that don't need it.

I didn't remove the global filesystem restriction key, I just moved it into `mountpointFilesystemRestrictions`. This vastly simplified some parts of the code compared to the initial implementation.

For the propagate example, let's say that you have a distro where the `/` filesystem can be almost any UNIX-compatible filesystem the user wants, but `/home` *has* to be ext4 for some reason (maybe software that is intended to be used on the OS is only compatible with ext4, but also only manages files under `/home`, the Dropbox client used to have limitations like this). My initial, naive implementation of the restrictions would mean that a user could make a single `/` partition with no separate `/home` partition, and make it with the BTRFS filesystem. Thus `/home` would wind up being BTRFS, even though the distro tried to configure it to only be ext4, and now the filesystem guarantee for the `/home` mountpoint is broken. To avoid this, the current implementation sees that the filesystem restrictions for `/home` are different than the ones for `/`, and when `/` and `/home` are on the same filesystem, it overlays their restrictions. That way you can't make a `/` that is BTRFS unless you split `/home` out into a separate ext4 partition.

This is all well and good, but now consider how this more advanced implementation would work once `/boot/efi` is involved. `/boot/efi` has to be FAT32. `/` has to **not** be FAT32. If you overlay those two sets of restrictions, you end up with no filesystem being allowable for `/` at all. In this instance, you actually *want* to be able to say "even though `/` and `/boot/efi` share the same filesystem, let me break `/boot/efi`'s restrictions." That's what the propagate key does - it allows you to specify whether Calamares should propagate the restrictions up to containing mountpoints or not.

ArrayBolt3:
> EDIT: I see what you are doing with propagate I think. However, this seems very strange to me. These settings should only apply to mountpoints, not to directories inside those mountpoints.

That's a valid point. However I think it's worth thinking about how a distro would be likely to wield this feature. You specify what filesystems you want for a specific mountpoint because you want to guarantee that any files saved under that mountpoint are saved on one of a particular set of filesystems. If a distro said specifically "allow `/` to be ext4 or xfs, allow `/home` to be ext4 only", it's reasonable for the developers of the distro to assume that now Calamares won't ever install their distro with a `/home` that's actually on xfs. That requires the overlay / propagation gymnastics to properly accomplish, otherwise you can end up with misconfigured systems.

dalto8:
> I don't think these options can be combined into the `partitionLayout` key. `partitionLayout` specifies the layout for automatic partitioning, not manual partitioning.

I agree.  I noted that above but you had probably already started responding at that point.

> For the propagate example, let's say that you have a distro where the `/` filesystem can be almost any UNIX-compatible filesystem the user wants, but `/home` _has_ to be ext4 for some reason (maybe software that is intended to be used on the OS is only compatible with ext4, but also only manages files under `/home`, the Dropbox client used to have limitations like this). My initial, naive implementation of the restrictions would mean that a user could make a single `/` partition with no separate `/home` partition, and make it with the BTRFS filesystem. Thus `/home` would wind up being BTRFS, even though the distro tried to configure it to only be ext4, and now the filesystem guarantee for the `/home` mountpoint is broken. To avoid this, the current implementation sees that the filesystem restrictions for `/home` are different than the ones for `/`, and when `/` and `/home` are on the same filesystem, it overlays their restrictions. That way you can't make a `/` that is BTRFS unless you split `/home` out into a separate ext4 partition.

That means these rules have nothing to do with mountpoints.  We are applying them to directories.  

For me, that is bridge too far.

Even if we decided we did want this to be about directories, I think the propagate concept is overlay complicated.

ArrayBolt3:
> I agree. I noted that above but you had probably already started responding at that point.

True, I'm currently running through the test plan so I happened to have this page open when you replied.

> That means these rules have nothing to do with mountpoints. We are applying them to directories.
>
> For me, that is bridge too far.

I mean, in a way that's true. I guess I am misusing the term "mountpoint" here to mean "somewhere a filesystem *can be* mounted", whereas the true definition is "somewhere a file system *is* mounted".

A couple possible solutions I can think of here:

* Rename the option to "directoryFilesystemRestrictions" to be more accurate
* Cut out the propagation handling code entirely and put a warning in the documentation that this option controls what filesystems can be used *if* the specified mountpoint is on its own partition, and that it does not necessarily guarantee that the specified directory will use the specified filesystem

I'd rather avoid the latter because of the issues I mentioned in my previous reply about how distros would use this feature, but this isn't a hill I'll die on, so I'm happy to remove the code if desired.

dalto8:
If we are going to enforce this on directories and not partitions, I think it would both simpler and more flexible to simply add a key named `onlyMountpoints` or something similar which would describe a given config should apply only to mountpoints or to the directory in general.  If someone configures a manual setup that validates that you could just throw an error and let them know what rules they violated.  Then you(and the distros) don't need to worry about how propagation is applied.

Either way, I think we should add back the global restrictions.

I also think instead of specifying the ESP mountpoint by path, we should just handle implicitly.  We already know what the ESP mountpoint is and we already know what filesystem it needs to be.

Lastly, I don't think we should have a list of hardcoded restrictions for `/`.  If the config doesn't specify anything, let any filesystem in the global settings be used.

ArrayBolt3:
> Either way, I think we should add back the global restrictions.

They're still there, you specify them like this:

```
mountpointFilesystemRestrictions:
    - mountpoint: "any"
      allowedFilesystemTypes: ["ext4","xfs","btrfs","fat32"]
```

That's what I meant by the key being moved into `mountpointFilesystemRestrictions`. I *can* move them back out, but the code becomes a lot messier in some spots because now the `CreatePartitionDialog` and `EditExistingPartitionDialog` classes need to have both `m_core` *and* `m_config` passed into them, rather than just `m_core`. You don't get any additional functionality from that, so I don't really see why we should do that. All of what the global key used to do, the `mountpoint: "any"` key still does.

I think I see what you mean about the overlay idea being too complicated, and `onlyMountpoints` being preferable. Now that you say that, I realize my implementation of `propagate` actually does pretty much exactly what you're describing, and does *not* behave the way I documented it behaving (there's a complicated edge case I only just now thought of that my code handles perfectly if the option is called `onlyMountpoints`, but that it fails at if you call it `propagate`). I'm more than happy to rename the option, that saves me a lot of work and will help distros avoid footguns.

> I also think instead of specifying the ESP mountpoint by path, we should just handle implicitly. We already know what the ESP mountpoint is and we already know what filesystem it needs to be.

Sure, that makes sense. I would like for the distro to be able to override that implicit handling though by setting a `/boot/efi` key, just in case machines of the future allow the ESP to use a filesystem other than fat32. Also we should probably implicitly handle `/efi` since some distros use that rather than `/boot/efi` (the Arch Wiki even has a note about it: https://wiki.archlinux.org/title/EFI_system_partition).

> Lastly, I don't think we should have a list of hardcoded restrictions for /. If the config doesn't specify anything, let any filesystem in the global settings be used.

hmm... assuming the user has fat32 in the list of global restrictions for the sake of the EFI system partition, that would allow people to try to install `/` with fat32, which will not work (or at least not be desirable) on most distros. The implementation won't let you use a filesystem on a mountpoint that is prohibited globally even if the mountpoint explicitly "allows" it (global restrictions take priority), so that doesn't pose a problem. Still, this isn't a hill I'll die on either, so I'll remove that if desirable.

kkofler:
I think it makes sense to require file systems for specific directories and everything below it, **but** the EFI System Partition is special and must always use its own separate requirements, independently of what is required for `/` or `/boot`.

I think I would just hardcode the requirements for the special partitions (EFI System Partition, BIOS Boot Partition, and the like), especially if they are UEFI-specific or legacy-BIOS-specific.

Also, the only case where I would intersect requirements is if the directories are on the same partition. Otherwise, I would use only the most specific directory, e.g., if we have requirements for `/` and for `/usr`, a separate partition for `/usr` or `/usr/local` should use only the ones for `/usr`. If the distro actually wants the intersection, they will have to intersect the sets explicitly and allow only the intersection for `/usr`.

dalto8:
> Sure, that makes sense. I would like for the distro to be able to override that implicit handling though by setting a `/boot/efi` key, just in case machines of the future allow the ESP to use a filesystem other than fat32. Also we should probably implicitly handle `/efi` since some distros use that rather than `/boot/efi` (the Arch Wiki even has a note about it: https://wiki.archlinux.org/title/EFI_system_partition).

The ESP is most commonly at `/efi`, `/boot` or `/boot/efi`.  However, it can be mounted anywhere.  But there is a key in global storage that tells you where it is mounted.  We should honor that to find the location.  Some distros don't even have a fixed location for the ESP, so we should honor the global storage for the location.

ArrayBolt3:
@kkofler 

> I think it makes sense to require file systems for specific directories and everything below it, but the EFI System Partition is special and must always use its own separate requirements, independently of what is required for / or /boot.

I don't like the idea of treating any partitions specially except when it comes to default settings that are user-overridable. If we're going to have `(/boot)/efi` treated specially in a way that can't be overridden, why don't we treat special partitions used on other platforms the same way? Raspberry Pis require that `/boot` be fat32 because that's what their firmware expects, which is no different than a PC expecting `/boot/efi` to be fat32. I'm sure lots of other boards have other requirements, some of which are almost certainly incompatible with each other (should the fact that Raspberry Pis require `/boot` to be fat32 mean *everyone* should use a fat32 `/boot`?). Really, the user should always be able to override the defaults, even if that means shooting themselves in the foot, unless the footgun is so dangerous or is so much of a support problem for the project that they can't afford to provide it.

> Also, the only case where I would intersect requirements is if the directories are on the same partition.

Agreed, this is what the current implementation does. If `/` is set to allow ext4, xfs, or btrfs, but `/home` only allows ext4, the current implementation will require that you make a `/home` partition with ext4 first, and *then* you can make `/` with BTRFS if you so desire. This does have the annoying side effect that in this setup, you can't make a btrfs `/` first and then an ext4 `/home` in as straightforward of a way as one might hope, but you can technically still do it by making `/` ext4, then making `/home` ext4, then changing `/` to be btrfs.

kkofler:
But these settings are **not** configurable by the **user** (at least not intended to be), only by the distro packager.

If specific hardware has specific restrictions, that can only be handled in code that actually detects the hardware and sets restrictions accordingly. A single config file setting is not going to work unless we are talking about a special distro image only for that specific hardware. But these days we are moving the opposite way, to generic images even for more and more ARM devices. So we cannot put hardware specifics unconditionally into a config file.

kkofler:
Oh, and UEFI is not x86-specific, it has always been used on some other platforms, and is now used more and more often on ARM as well.

ArrayBolt3:
> But these settings are not configurable by the user (at least not intended to be), only by the distro packager.

Right, that's what I meant. Sorry about that, I should have been more clear.

> If specific hardware has specific restrictions, that can only be handled in code that actually detects the hardware and sets restrictions accordingly. A single config file setting is not going to work unless we are talking about a special distro image only for that specific hardware. But these days we are moving the opposite way, to generic images even for more and more ARM devices. So we cannot put hardware specifics unconditionally into a config file.

I don't think I've ever seen a distro that has a "universal" distro image that works across many different mutually incompatible devices. Even if it is true though, how does hardcoding defaults for this sort of thing directly in Calamares help? If putting the specifics in a config file is bad, putting them in the binary is many times worse.

> Oh, and UEFI is not x86-specific, it has always been used on some other platforms, and is now used more and more often on ARM as well.

I am aware, but I don't know how this has any bearing on the rest of the discussion.

ArrayBolt3:
@dalto8 

> The ESP is most commonly at /efi, /boot or /boot/efi. However, it can be mounted anywhere. But there is a key in global storage that tells you where it is mounted. We should honor that to find the location. Some distros don't even have a fixed location for the ESP, so we should honor the global storage for the location.

I'm not sure if this is actually possible at the level I'm applying the restrictions at. I'm applying them on the manual partitioning screen, such that the user isn't even allowed to select a combination of mountpoints and filesystems that prohibited by the configuration. Trying to make a fat32 `/` will result in an error message, and the "OK" button on the partition creation or partition change dialog is grayed out until the user changes their settings. To my awareness, there's no way for Calamares to know *where* the EFI system partition is located at this point, since the user is actively deciding where to locate it right then and there. I guess we could special-case the situation where the user is installing on UEFI hardware and sets the `boot` flag on a partition, which indicates that the partition should be an EFI system partition. Then only fat32 would be allowable as a filesystem. I'd still rather have it where distros could choose what filesystems to allow on the ESP though in case UEFI systems start supporting more filesystems for the ESP.

The test plan specified earlier has now passed in its entirety. I used a slightly modified local version of the code for the testing, where I fixed the one TODO about code duplication with the unmountable filesystem list, however functionally the code should be exactly equivalent to what's currently in the PR. So I think I'm ready to make whatever final changes are ready to make this mergable, and then it should be ready. Is the following a good summary of the changes tthat need to be made?

* Rename `propagate` to `onlyMountpoints` and change the documentation to more closely reflect what that option does
* Remove the list of hardcoded restrictions for `/`
* Do something about ESP handling, whether that be using GlobalStorage or a boot flag somehow, or also treating `/efi` as an ESP location and restricting it to fat32

dalto8:
> the current implementation will require that you make a `/home` partition with ext4 first, and _then_ you can make `/` with BTRFS if you so desire.

That is not at all workable from a usability perspective.  This can be easily avoided.

Apply the mountpoint restrictions in the individual mount boxes and the directory restrictions when you exit manual partitioning.  Calamares already is doing a bunch of checks at that point so it should be straightforward to add a few more.

> To my awareness, there's no way for Calamares to know _where_ the EFI system partition is located at this point, since the user is actively deciding where to locate it right then and there

That isn't how Calamares works.  Even in manual partitioning, you can't choose your own ESP location.  You must mount it at the location specified in global storage.

ArrayBolt3:
> Apply the mountpoint restrictions in the individual mount boxes and the directory restrictions when you exit manual partitioning. Calamares already is doing a bunch of checks at that point so it should be straightforward to add a few more.

Sounds good, implementation should be straightforward, will work that in.

> That isn't how Calamares works. Even in manual partitioning, you can't choose your own ESP location. You must mount it at the location specified in global storage.

Neat. No idea how that works (I should probably find that part of the code and read it), but if that will work, I'll get that worked in too.

ArrayBolt3:
@dalto8 Alright, see what you think of that. It did NOT end up being straightforward to implement the post-manual partitoining validator, but it was doable. I haven't tested this thoroughly, and I haven't tried building it against Qt6, but I have given it some testing, and so far it seems to work correctly.

dalto8:
Thanks.  I will take a look as soon as I have a free minute.

One quick thing is that the description for `onlyWhenMountpoint` seems backwards to me from a boolean perspective.  Is the description wrong or do we need to rename the option?

ArrayBolt3:
:facepalm: The description is backwards. I wrote the description in a way that matched the semantics of the old `propagate` attribute, then reversed the semantics of the attribute when renaming it `onlyWhenMountpoint`. Fixing that now. Edit: fixed.

adriaandegroot:
I have followed the comments thread, and if there's some consensus that this is ready, I can merge it for 3.3.13 (lucky 13!). Just a couple of general comments:
- with manual partitioning, we have always allowed people to shoot themselves in the foot; there are some warnings that Calamares pops up when something is likely to go wrong, but the footguns are there;
- automated partitioning should DTRT as configured by the distro, and I don't think this whole MR has anything to change there.
- if you are retiring a configuration key, the best thing to do is to check if the retired key is present, warn about it, and then write the value of the retired key into the configuration map where it should have been (e.g. moving a key into a sub-key).

ArrayBolt3:
> with manual partitioning, we have always allowed people to shoot themselves in the foot; there are some warnings that Calamares pops up when something is likely to go wrong, but the footguns are there;

True, right now this does block some footguns but I can change one line of code to them back I believe (warn the user that their partition configuration is unsupported but let them click "OK" nonetheless).

> automated partitioning should DTRT as configured by the distro, and I don't think this whole MR has anything to change there.

Correct.

> if you are retiring a configuration key, the best thing to do is to check if the retired key is present, warn about it, and then write the value of the retired key into the configuration map where it should have been (e.g. moving a key into a sub-key).

The bit about moving a key into a sub-key was just changing how this particular PR was implemented. I had added two new keys, but then changed to only adding one new key. So no existing keys have been retired.

This is almost ready to merge, but I do want to do one more thorough test first. I'll get that done today most likely.

ArrayBolt3:
Test plan:

* [x] When a directoryFilesystemRestrictions entry is missing directory entry, ignores section entirely and falls back to defaults?
* [x] When a directoryFilesystemRestrictions entry is missing allowedFilesystemTypes, ignores section entirely and falls back to defaults?
* [x] Unrecognized filesystems are ignored and don't cause problems?
* [x] Leaving out an entry for EFI filesystem restrictions results in only fat32 being allowed for the EFI system partition, even if other partitions have restrictions set?
* [x] Leaving out the any entry in directoryFilesystemRestrictions results in all filesystems being permissible in general?
* [x] Not allowing a filesystem for a specific directory results in a warning being shown if the prohibited filesystem is specified for that directory?
* [x] Not allowing a filesystem in general (by setting the any entry and restricting the allowable filesystems there) results in a warning being shown if the prohibited filesystem is specified anywhere?
* [x] Adding a filesystem to the list of allowable filesystems for a directory but leaving it out of the list of allowable filesystems in the any entry results in a warning being shown if that filesystem is selected even for that directory?
* [x] Set the directory / to allow filesystems ext4, xfs, and btrfs, but set /home to only allow ext4 and xfs. Creating a / with BTRFS and no separate /home partition results in a warning when finishing manual partitioning?
* [x] Set the directory / to allow filesystems ext4, xfs, and btrfs, set /home to only allow ext4 and xfs, and set onlyWhenMountpoint = true for /home. No warning is shown when creating / with btrfs even without creating a separate /home?
* [x] Set the directory / to allow filesystems ext4, xfs, and btrfs, set /home to only allow ext4 and xfs, and set onlyWhenMountpoint = true for /home. Creating a separate /home with btrfs results in a warning?
* [x] Set the directory / to allow filesystems ext4, xfs, and btrfs, but set /home to only allow ext4 and xfs. Do NOT set the any entry at all. Can create an /opt partition with minix filesystem?
* [x] Filesystem checks are done both when changing mountpoint and when changing filesystem, for both partition creation and partition edit dialogs?
* [x] Set the directory / to allow filesystems ext4, xfs, and btrfs. Create a partition, set the mountpoint as /. and then select an unmountable partition type such as unformatted. No warning shown?
* [x] Set the directory / to allow filesystems ext4, set "efi" to allow fat16 and fat32, do NOT set onlyWhenMountpoint on the efi entry. Create a partition, set the mountpoint as /, and set the filesystem to ext4. Results in a warning about incompatible filesystems, and also prints to the log "WARNING: no filesystems are valid for path ' "/" ', check directoryFilesystemRestrictions for issues"?
* [x] Install a distro on an EFI machine with default filesystem restrictions, creating a single fat32 partition at the proper location for an EFI system partition, and a single ext4 partition at /. Installation succeeds?
* [x] On a BIOS system, set / to allow ext4 or btrfs, set /home to allow ext4 only, then create a partitioning layout with / set to ext4 in a primary partition, and /home set to minix in an extended partition. Warning shows up about the /home filesystem not being allowable when leaving manual partitioning?
* [x] On a BIOS system, set / to allow ext4 or btrfs, set /home to allow ext4 only, then create a partitioning layout with / set to btrfs in a primary partition, and /home set to ext4 in an extended partition. Works without errors?
* [x] On a BIOS system with two disks, set / to allow ext4 or btrfs, set /home to allow ext4 only. Dedicate the first disk to /home with ext4. On the second disk, make the first partition /boot on ext3, and make the second partition / on BTRFS inside an extended partition. Works without errors?
* [x] Code builds against both Qt5 and Qt6?

ArrayBolt3:
Ran into a bug and gdb is being *extremely* uncooperative, so it's taking me a while to figure out what's wrong. Edit: found it, I accidentally have "your setup is bad and no filesystems are allowed for this mountpoint" being interpreted as "all filesystems are allowed for this mountpoint".

ArrayBolt3:
Sigh, almost everything works, but something about my extended partition handling is bad. I don't have time to fix it today, so I'll have to sort it out probably tomorrow night.

ArrayBolt3:
Found the bug, I was using an `i` index variable where I meant to use a `j` index variable. Testing underway, should be done soon.

dalto8:
A sure sign you should be using meaningful names instead of `i` and `j` :)

ArrayBolt3:
They were counter variables in for loops :) but yes in almost any other situation I would have used descriptive names. Heck, maybe this is an argument for using descriptive names even in for loops.

ArrayBolt3:
Alright, I officially declare this ready for final review and merge. Thanks for your patience with me.

src/modules/partition/core/DirFSRestrictLayout.cpp
```diff
@@ -0,0 +1,234 @@
+/* === This file is part of Calamares - <https://calamares.io> ===
+ *
+ *   SPDX-FileCopyrightText: 2014-2017 Teo Mrnjavac <teo@kde.org>
+ *   SPDX-FileCopyrightText: 2017-2018 Adriaan de Groot <groot@kde.org>
+ *   SPDX-FileCopyrightText: 2018-2019 Collabora Ltd <arnaud.ferraris@collabora.com>
```
adriaandegroot, line 5: This is all your own work, so please put in a single line for yourself, and leave these out since they simply are not true.

src/modules/partition/core/DirFSRestrictLayout.h
```diff
@@ -0,0 +1,86 @@
+/* === This file is part of Calamares - <https://calamares.io> ===
+ *
+ *   SPDX-FileCopyrightText: 2018-2019 Collabora Ltd <arnaud.ferraris@collabora.com>
+ *   SPDX-FileCopyrightText: 2019 Adriaan de Groot <groot@kde.org>
```
adriaandegroot, line 4: Same for these attributions.

src/modules/partition/core/DirFSRestrictLayout.h
```diff
@@ -0,0 +1,86 @@
...
+#include <QVariantMap>
+
+class DirFSRestrictLayout
+{
+public:
+    struct DirFSRestrictEntry
+    {
+        QString dirPath;
+        QList< FileSystem::Type > dirAllowedFSTypes;
+        bool useOnlyWhenMountpoint;
```
adriaandegroot: Is uninitialized with the default constructor you've written. If you in-class initialize it, you can `= default` the default constructor.

adriaandegroot:
Looks good enough to me. I think there's a bunch of `const` we can add, and some strengthened types here and there and maybe some code to lose (eg. relying on aggregate initialization instead of writing out some constructors). But let's do those in follow-up MRs where someone (e.g. me) can Have An Opinion at the code.

src/modules/partition/core/DirFSRestrictLayout.cpp
ArrayBolt3: It's not strictly my work, I copied an existing file in Calamares and then modified it until it did what I wanted, so I think the existing attributions are accurate. I'm happy to add my own too though.

ArrayBolt3: Added.

src/modules/partition/core/DirFSRestrictLayout.h
ArrayBolt3: Added a line for myself to this file.

src/modules/partition/core/DirFSRestrictLayout.h
ArrayBolt3: Fixed using your suggestion, good catch and thanks for the tip about using `= default`!

PR #2401, opened 2024-11-22
[dracut] Add an options setting for additional Dracut parameters

tsimonq2:
Lubuntu is in the process of switching to Dracut.

Due to [Ubuntu bug 2084809](https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/2084809), we need to be able to override the options passed to Dracut.

This adds the `options` config entry, which is optional.

### DIFF
--- a/src/modules/dracut/dracut.conf
+++ b/src/modules/dracut/dracut.conf
@@ -8,3 +8,8 @@
 # set a custom name, including the path
 #
 initramfsName: /boot/initramfs-freebsd.img
+
+# Optional: define a list of strings to be passed as arguments to Dracut
+# By default, -f is always included
+options:
+  - "-f"
--- a/src/modules/dracut/dracut.conf
+++ b/src/modules/dracut/dracut.conf
@@ -11,5 +11,4 @@ initramfsName: /boot/initramfs-freebsd.img
 
 # Optional: define a list of strings to be passed as arguments to Dracut
 # By default, -f is always included
-options:
-  - "-f"
+options: [ "-f" ]
--- a/src/modules/dracut/main.py
+++ b/src/modules/dracut/main.py
@@ -35,15 +35,22 @@ def run_dracut():
 
     :return:
     """
+    # Fetch the job configuration
+    cli_options = ["-f"]
+    initramfs_name = libcalamares.job.configuration.get('initramfsName', None)
+    dracut_options = libcalamares.job.configuration.get('options', [])
+
+    # Parse the custom options if there are any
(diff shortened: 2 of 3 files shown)


src/modules/dracut/main.py
```diff
@@ -35,15 +35,22 @@ def run_dracut():
...
+    dracut_options = libcalamares.job.configuration.get('options', [])
+
+    # Parse the custom options if there are any
+    for option in dracut_options:
+        # Deduplication check
+        if option not in cli_options:
+            cli_options.append(option)
+
+    if initramfs_name:
+        cli_options.append(initramfs_name)
```
dalto8: If we give people the ability specify options, I don't think we should force `-f`.  We should make it the default, but allow them to remove it if they choose to do so.

This is just my thought on how it could look but feel free to write in your own style if you prefer something different.

```suggestion
    # Fetch the job configuration
    initramfs_name = libcalamares.job.configuration.get('initramfsName', None)
    dracut_options = libcalamares.job.configuration.get('options', ["-f"])
    
    if initramfs_name:
        dracut_options.append(initramfs_name)
```

src/modules/dracut/dracut.conf
```diff
@@ -8,3 +8,8 @@
 # set a custom name, including the path
 #
 initramfsName: /boot/initramfs-freebsd.img
+
+# Optional: define a list of strings to be passed as arguments to Dracut
+# By default, -f is always included
+options:
+  - "-f"
```
dalto8: Although it would be OK either way.  I think this would be more consistent with how we treat arrays of options in other conf files.

```suggestion
options: [ "-f" ]
```

src/modules/dracut/main.py
```diff
@@ -35,15 +35,22 @@ def run_dracut():
...
     try:
-        initramfs_name = libcalamares.job.configuration['initramfsName']
-        target_env_process_output(['dracut', '-f', initramfs_name])
-    except KeyError:
-        try:
-            target_env_process_output(['dracut', '-f'])
-        except subprocess.CalledProcessError as cpe:
-            libcalamares.utils.warning(f"Dracut failed with output: {cpe.output}")
-            return cpe.returncode
+        target_env_process_output(['dracut'] + cli_options)
```
dalto8:
```suggestion
        target_env_process_output(['dracut'] + dracut_options)
```

tsimonq2:
@dalto8 @adriaandegroot Does this look better?

dalto8 (approved):

PR #2402, opened 2024-11-23
Properly close LUKS partitions after installation is complete

tsimonq2:
The unmount module only unmounts the root partition, and does not close any encrypted partitions. This allows the user to still re-mount the drive without having to enter the encryption password again.

This pull request fixes it. There is some further cleanup to do in the `partition` module, to move the unmount and mount calls (encryption and not) to libcalamares. I'd like to see #2400 merged before that.

### DIFF
--- a/src/libcalamares/partition/Mount.cpp
+++ b/src/libcalamares/partition/Mount.cpp
@@ -84,6 +84,15 @@ unmount( const QString& path, const QStringList& options )
     return r.getExitCode();
 }
 
+int
+closeCryptsetup( const QString& luksMapperName )
+{
+    QStringList args = { "cryptsetup", "close", luksMapperName };
+    auto r = Calamares::System::runCommand( QStringList { "cryptsetup", "close", luksMapperName } , std::chrono::seconds( 10 ) );
+    sync();
+    return r.getExitCode();
+}
+
 struct TemporaryMount::Private
 {
     QString m_devicePath;
--- a/src/libcalamares/partition/Mount.h
+++ b/src/libcalamares/partition/Mount.h
@@ -51,6 +51,13 @@ DLLEXPORT int mount( const QString& devicePath,
  */
 DLLEXPORT int unmount( const QString& path, const QStringList& options = QStringList() );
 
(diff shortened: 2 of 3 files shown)


dalto8:
Can you elaborate how this change would be advantageous?

Some people like to modify the installation before rebooting and this would introduce some friction into that process.

I don't see an obvious security issue here unless I am missing it.  Luks volumes normally stay unlocked unless they are explicitly locked by the user so it seems somewhat odd to close it when the installer finishes.

Lastly, there is no user data in that volume at the time the installer is completing.

I am not strongly opposed to the change, I am just not seeing the benefits that would outweigh the downsides.

tsimonq2:
Thanks for your response.

Under normal circumstances where the installation is unencrypted, the root partition is unmounted anyway when using the unmount module. This does not add friction in my opinion, it simply ensures that, if an unmount is requested, it's done such that the disk can be completely detached.

If someone is working with an ISO that is older (maybe released e.g. 6 months ago) and there are both security vulnerabilities present and network access, it is theoretically possible to mount that volume again. The idea is, if the user decides to continue using the live instance after installing and are unaware that the LUKS partition is still decrypted but doesn't show up as mounted in the file manager, the average person is likely not going to notice. Lots of "ifs," but something bad is possible.

That being said, I think it's less for security and more of a "clean unmount."

Let me know if there is anything else I can do to help.

dalto8:
This is just my personal opinion but it seems like closing the luks volume at the end of install is counterintuitive to expected behavior.

kkofler:
I think it is expected. We unmount everything from the installed system when done, so I would also expect LVM and LUKS volumes to get closed.

adriaandegroot:
I get the feeling this is like a "after-the-end" action that might be taken (e.g. after the "finished" page, before the application really shuts down) if the user makes some suitable choice there. After all, both "shut it down for security" and "leave it open for post-install shenanigans" are plausible use-cases.

adriaandegroot:
Closing this after pulling it into the repo, in a separate branch. We can continue the topic at the other side of migration.

Issue #2409, opened 2024-12-10
Allow skipping over squashfses that don't exist


ArrayBolt3:
**Is your feature request related to a problem? Please describe.**
The current support for multiple squashfses would be very useful for situations in which you have a number of related distros with the same Calamares config, but that may have squashfses located in different places. For instance, Debian's live-build tool supports both initramfs-tools and dracut, which have different ideas about where to mount the ISO under `/run`. calamares-settinsg-debian is only able to specify one location for the squashfs, thus it has to default to one, and live-build has to manually change the unpackfs configuration in order to point it at the other if needed. This is hacky, and if live-build *does* have to edit the config, it will show up as modified in the output of `debsums` and `dpkg -V`.

**Describe the solution you'd like**
Allow one to specify that a squashfs or other unpackable filesystem is optional, meaning that if it doesn't exist, Calamares will skip over it without erroring out. This way both filesystems could be mentioned in the unpackfs configuration, and whichever one doesn't exist will be skipped over.

**Describe alternatives you've considered**
None, though I'm open to other alternatives that don't require modifying the Calamares configuration dynamically.

**Additional context**
Attempting to do this with the existing Calamares code results in an error when attempting to unpack a squashfs that doesn't exist. (Note that I'm showing a Lubuntu window here, but that's because it was the easiest way I could think of to get an up-to-date Calamares for testing with. This feature request would be useful to Kicksecure in particular, not Lubuntu, though I guess it could be handy for Lubuntu theoretically too.)

![Screenshot_20241209_202033](https://github.com/user-attachments/assets/71205578-f71c-46b1-a80b-22b680850839)

PR #2410, opened 2024-12-11
Allow for optional items in unpackfs (Fixes: #2409)

tsimonq2:
This simply allows for items to be skipped if optional is specified and the item is not present on the filesystem.

@ArrayBolt3 WDYT?

### DIFF
--- a/src/modules/unpackfs/main.py
+++ b/src/modules/unpackfs/main.py
@@ -48,7 +48,7 @@ class UnpackEntry:
     :param destination:
     """
     __slots__ = ('source', 'sourcefs', 'destination', 'copied', 'total', 'exclude', 'excludeFile',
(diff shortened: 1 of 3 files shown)


ArrayBolt3:
Looks like it will work based on a quick skim, I'll try to test it later today and report back the results.

ArrayBolt3:
Tested it in the following situations:

* Only squashfs files that exist are present in unpackfs.conf: Expected: Works, Actual: Works
* One squashfs file that doesn't exist is present in unpackfs,conf: Expected: Doesn't work, Actual: Doesn't work
* One squashfs file that doesn't exist is present in unpackfs.conf and marked optional: Expected: Works, Actual: Works

So it seems to be working :D

PR #2419, opened 2024-12-25
i18n: [calamares] Fix Ukrainian localization

SoulPrudya:


adriaandegroot:
We usually do -- or did -- the translations in Transifex, but would sort-of want to migrate off of that platform due to its increasing push of generative language models. So I'll merge this, then push the change "upstream" to Transifex, too.

PR #2422, opened 2025-01-05
Hybrid bootloader (BIOS+UEFI) installation support

ArrayBolt3:
It would be very handy for some use case scenarios if Calamares could be configured to install both BIOS and GRUB bootloaders on the same system. The primary use for this would be portable installations, where someone does a full install of a distro onto a USB flash drive that they then intend to run on multiple physical machines with varying types of firmware. Currently Calamares can only install the bootloader matching the firmware type of the system being booted though. This PR attempts to change that by adding an option `installHybridGRUB` to `bootloader.conf`. When enabled, Calamares will attempt to install both BIOS and UEFI GRUB bootloaders.

Note, I'm not sure if the approach I'm taking here is correct - it seems a bit overly hacky. This PR also isn't complete - it still needs to tackle a problem with partitioning (namely, the EFI system partition is only made on EFI systems, so on BIOS systems, one has to specify an ESP themselves in partition.conf... but if you do that, you end up with two ESPs on EFI systems, because Calamares will add an ESP regardless of whether one already exists in the partition layout or not).

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -658,7 +658,7 @@ def run_grub_install(fw_type, partitions, efi_directory, install_hybrid_grub):
                 return
             
         # boot_loader_install_path points to the physical disk to install GRUB
-        # to. It should start with "/dev/", and be at least as long as the 
+        # to. It should start with "/dev/", and be at least as long as the
         # string "/dev/sda".
         if not boot_loader_install_path.startswith("/dev/") or len(boot_loader_install_path) < 8:
             raise ValueError(f"boot_loader_install_path contains unexpected value '{boot_loader_install_path}'")
--- a/src/modules/partition/core/PartitionActions.cpp
+++ b/src/modules/partition/core/PartitionActions.cpp
@@ -89,6 +89,11 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
     Calamares::GlobalStorage* gs = Calamares::JobQueue::instance()->globalStorage();
 
     const bool isEfi = PartUtils::isEfiSystem();
+    
+    bool createHybridBootloaderLayout = false;
+    if ( gs->contains( "createHybridBootloaderLayout" ) ) {
+        createHybridBootloaderLayout = gs->value( "createHybridBootloaderLayout" ).toBool();
+    }
 
     // Partition sizes are expressed in MiB, should be multiples of
     // the logical sector size (usually 512B). EFI starts with 2MiB
@@ -105,7 +110,7 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
     PartitionTable::TableType partType = PartitionTable::nameToTableType( o.defaultPartitionTableType );
     if ( partType == PartitionTable::unknownTableType )
     {
-        partType = isEfi ? PartitionTable::gpt : PartitionTable::msdos;
+        partType = ( isEfi || createHybridBootloaderLayout ) ? PartitionTable::gpt : PartitionTable::msdos;
     }
     // last usable sector possibly allowing for secondary GPT using 66 sectors (256 entries)
     const qint64 lastUsableSector = dev->totalLogical() - ( partType == PartitionTable::gpt ? 67 : 1 );
@@ -118,7 +123,7 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
 
     core->createPartitionTable( dev, partType );
 
-    if ( isEfi )
+    if ( createHybridBootloaderLayout || isEfi )
     {
         qint64 uefisys_part_sizeB = PartUtils::efiFilesystemRecommendedSize();
         qint64 efiSectorCount = Calamares::bytesToSectors( uefisys_part_sizeB, dev->logicalSize() );
@@ -144,6 +149,25 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
         }
         core->createPartition( dev, efiPartition, KPM_PARTITION_FLAG_ESP );
         firstFreeSector = lastSector + 1;
+
+        if ( createHybridBootloaderLayout )
+        {
+            qint64 bios_part_sizeB = 8_MiB;
+            qint64 biosSectorCount = Calamares::bytesToSectors( bios_part_sizeB, dev->logicalSize() );
+            Q_ASSERT( biosSectorCount > 0 );
+
+            qint64 lastSector = firstFreeSector + biosSectorCount - 1;
+            Partition* biosPartition = KPMHelpers::createNewPartition( dev->partitionTable(),
+                                                                       *dev,
+                                                                       PartitionRole( PartitionRole::Primary ),
+                                                                       FileSystem::Unformatted,
+                                                                       QString(),
+                                                                       firstFreeSector,
+                                                                       lastSector,
+                                                                       KPM_PARTITION_FLAG( None ) );
+            core->createPartition( dev, biosPartition, KPM_PARTITION_FLAG( BiosGrub ) );
+            firstFreeSector = lastSector + 1;
+        }
     }
 
     const bool mayCreateSwap
--- a/src/modules/partition/gui/PartitionLabelsView.cpp
+++ b/src/modules/partition/gui/PartitionLabelsView.cpp
@@ -13,6 +13,7 @@
 #include "core/ColorUtils.h"
 #include "core/PartitionModel.h"
 #include "core/SizeUtils.h"
+#include "core/KPMHelpers.h"
 
 #include "utils/Gui.h"
 #include "utils/Logger.h"
@@ -20,6 +21,7 @@
 
 #include <kpmcore/core/device.h>
 #include <kpmcore/fs/filesystem.h>
+#include <kpmcore/core/partition.h>
 
 // Qt
 #include <QGuiApplication>
@@ -39,6 +41,12 @@ buildUnknownDisklabelTexts( Device* dev )
     QStringList texts = { QObject::tr( "Unpartitioned space or unknown partition table", "@info" ),
                           formatByteSize( dev->totalLogical() * dev->logicalSize() ) };
     return texts;
+}
+
+static uint
+getPartitionModelIndexFlags( const QModelIndex& index )
+{
+    return static_cast< Partition* >( index.data( PartitionModel::PartitionPtrRole ).value< void* >() )->property( "_calamares_flags" ).toUInt();
 }
 
 PartitionLabelsView::PartitionLabelsView( QWidget* parent )
@@ -189,6 +197,11 @@ PartitionLabelsView::buildTexts( const QModelIndex& index ) const
                       && index.data( PartitionModel::FileSystemTypeRole ).toInt() == FileSystem::Fat32 )
             {
                 firstLine = tr( "EFI system", "@label" );
+            }
+            else if ( index.data( PartitionModel::FileSystemTypeRole ).toInt() == FileSystem::Unformatted
+                      && getPartitionModelIndexFlags( index ) & KPM_PARTITION_FLAG( BiosGrub ) )
+            {
+                firstLine = tr("BIOS boot", "@label" );
             }
             else if ( index.data( PartitionModel::FileSystemTypeRole ).toInt() == FileSystem::LinuxSwap )
             {
--- a/src/modules/bootloader/bootloader.conf
+++ b/src/modules/bootloader/bootloader.conf
@@ -81,3 +81,6 @@ efiBootMgr: "efibootmgr"
 # to add another module to optionally install the fallback on those
 # boards that need it.
 installEFIFallback: true
+
+# Optionally install both BIOS and UEFI GRUB bootloaders.
+installHybridGRUB: false
--- a/src/modules/bootloader/bootloader.schema.yaml
+++ b/src/modules/bootloader/bootloader.schema.yaml
@@ -24,3 +24,4 @@ properties:
 
     efiBootloaderId:  { type: string }
     installEFIFallback: { type: boolean }
+    installHybridGRUB: { type: boolean }
--- a/src/modules/bootloader/main.py.orig
+++ b/src/modules/bootloader/main.py.orig
@@ -637,20 +637,16 @@ def run_grub_install(fw_type, partitions, efi_directory, install_hybrid_grub):
                                    "--bootloader-id=" + efi_bootloader_id,
                                    "--force"])
     else:
-        if efi_directory is not None and not install_hybrid_grub:
-            libcalamares.utils.warning(_("Cannot install BIOS bootloader on UEFI installation when install_hybrid_grub is 'False'!"))
-            return
+        if efi_directory is not None:
+            if not install_hybrid_grub:
+                return
 
         if libcalamares.globalstorage.value("bootLoader") is None:
             efi_install_path = libcalamares.globalstorage.value("efiSystemPartition")
             if efi_install_path is None or efi_install_path == "":
                 efi_install_path = "/boot/efi"
-            find_esp_disk_command = f"lsblk -o PKNAME \"$(df --output=source '{efi_install_path}' | tail -n1)\""
+            find_esp_disk_command = f"echo \"/dev/$(lsblk -o PKNAME $(df --output=source '{efi_install_path}' | tail -n1) | tail -n1)\""
             boot_loader_install_path = check_target_env_output(["sh", "-c", find_esp_disk_command]).strip()
-            if not "\n" in boot_loader_install_path:
-                libcalamares.utils.warning(_("Cannot find the drive containing the EFI system partition!"))
-                return
-            boot_loader_install_path = "/dev/" + boot_loader_install_path.split("\n")[1]
         else:
             boot_loader = libcalamares.globalstorage.value("bootLoader")
             boot_loader_install_path = boot_loader["installPath"]
@@ -658,8 +654,8 @@ def run_grub_install(fw_type, partitions, efi_directory, install_hybrid_grub):
                 return
             
         # boot_loader_install_path points to the physical disk to install GRUB
-        # to. It should start with "/dev/", and be at least as long as the 
-        # string "/dev/sda".
+        # to. It should start with `/dev/` be at least as long as the string
+        # `/dev/sda`.
         if not boot_loader_install_path.startswith("/dev/") or len(boot_loader_install_path) < 8:
             raise ValueError(f"boot_loader_install_path contains unexpected value '{boot_loader_install_path}'")
 
--- a/src/modules/partition/PartitionViewStep.cpp
+++ b/src/modules/partition/PartitionViewStep.cpp
@@ -918,6 +918,8 @@ PartitionViewStep::setConfigurationMap( const QVariantMap& configurationMap )
                       "will use gpt for efi or msdos for bios";
     }
     gs->insert( "defaultPartitionTableType", partitionTableName );
+    gs->insert( "createHybridBootloaderLayout",
+                Calamares::getBool( configurationMap, "createHybridBootloaderLayout", false ) );
 
     // Now that we have the config, we load the PartitionCoreModule in the background
     // because it could take a while. Then when it's done, we can set up the widgets
--- a/src/modules/partition/partition.conf
+++ b/src/modules/partition/partition.conf
@@ -178,6 +178,13 @@ initialSwapChoice: none
 #
 # defaultPartitionTableType: msdos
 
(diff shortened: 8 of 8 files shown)


src/modules/bootloader/main.py
```diff
@@ -637,25 +637,47 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+                return
+
+            # We're actually on an EFI system even though fw_type is "bios".
+            # We need to find the proper drive to install to based on which
+            # drive the ESP is located on.
+            installation_root_path = libcalamares.globalstorage.value("rootMountPoint")
+            install_efi_directory = installation_root_path + efi_directory
+            find_esp_disk_command = "echo \"/dev/$(lsblk -o PKNAME $(df --output=source" \
+                + install_efi_directory \
+                + " | tail -n1) | tail -n1)\""
```
dalto8: Would this work(and be simpler)?  `f"findmnt -no source {install_efi_directory}"`

Also, try to use f strings when it can improve readability.  When the module was written it wasn't supported but for new code it is available now.

That being said, I am not sure this should be really be happening here.  It seems like this should come from the partition module.

src/modules/bootloader/main.py
```diff
@@ -637,25 +637,47 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+            boot_loader_install_path = subprocess.run(["sh", "-c", find_esp_disk_command],
+                capture_output=True).stdout.decode().strip()
```
dalto8: Please don't use subprocess.run, instead use Calamares's python functions for running commands.

src/modules/bootloader/main.py
```diff
@@ -923,7 +956,11 @@ def run():
             return None
 
     try:
-        prepare_bootloader(fw_type)
+        if (boot_loader == "grub" or efi_boot_loader == "grub") and install_hybrid_grub:
+            prepare_bootloader("bios")
+            prepare_bootloader("efi")
+        else:
+            prepare_bootloader(fw_type)
```
dalto8: I think this would be cleaner if we implemented as part of the bootloader selection logic inside `prepare_bootloader()`

dalto8 (commented):
Specific comments below but, generally, it seems like this really needs changes and support in the partition module.  This will otherwise be too fragile I think.

We need to be ensuring that the partition setup is correct for this implementation method.  It would also solve the issue with the duplicate ESPs.

ArrayBolt3:
Agreed, the partition module will need some help. I'd really like if there's some way to make it so that this change only requires things to be tweaked in one spot (a bootloader setting) rather than needing both bootloader and partition config changes, but I'm not sure how practical that is yet.

Thanks for the feedback, I'll try some more things and see if I can improve on this. (This implementation actually does work at least in the most straightforward scenarios, if you combine it with some partition layout tweaking, but I'm not thrilled with it and would like to make it cleaner.)

src/modules/bootloader/main.py
ArrayBolt3: `findmnt` doesn't quite work because it gives me the partition device file, not the disk device file (which is what `grub-install` wants when doing a BIOS-style installation). It could probably replace the `df` stuff in the innermost subshell, but the `lsblk` and `echo` will still be needed if this kind of logic is used.

dalto8: Is that not in global storage already in `partitions`?

ArrayBolt3: It does not seem to be based on some debugging. On BIOS systems, it's in the bootLoader key, but on EFI systems it's not there at all. On the VM I just tested this on, there's some strange "empty" device (`{'claimed': False, 'device': '', 'features': {}, 'fs': 'unknown', 'fsName': 'unknown', 'mountPoint': '', 'partattrs': 0, 'partlabel': '', 'parttype': '', 'partuuid': '', 'uuid': ''}`), then there's four devices, one for each of /dev/vda1 through /dev/vda4. /dev/vda itself isn't in the list.

I don't want to just transform the partition name to a disk name since it's not guaranteed the two will have a clear correlation  - we see a minor version of this with /dev/sda and /dev/sda1 vs. /dev/mmcblk0 and /dev/mmcblk0p1.

src/modules/bootloader/main.py
ArrayBolt3: Moved.

ArrayBolt3:
It's definitely not perfect yet, but this is significantly better than the initial draft. Support for getting the partition layout right regardless of host firmware is now added to the partition module, and the actual bootloader installation code is a bit less messy. Currently the BIOS boot partition is simply labeled "New Partition", so that will need fixed, but this is getting close to ready.

src/modules/bootloader/main.py
ArrayBolt3: Got it, latest commits do it the correct way.

ArrayBolt3:
@dalto8 Would like another preliminary review when you have time. This is pretty close to ready, I have BIOS boot partition detection "working", though I'm not satisfied with the strategy I'm using at the moment (it's assuming any filesystemless partition 8 MiB or less in size is BIOS boot, which may not always be true, but checking for a bios-boot flag turns out to be hard).

ArrayBolt3:
Alright, I fixed the bios-grub partition detection problem! I think this is now ready for review. Probably will need a commit squash before merge though.

src/modules/partition/core/PartitionActions.cpp
```diff
@@ -105,7 +110,11 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
     PartitionTable::TableType partType = PartitionTable::nameToTableType( o.defaultPartitionTableType );
     if ( partType == PartitionTable::unknownTableType )
     {
-        partType = isEfi ? PartitionTable::gpt : PartitionTable::msdos;
+        partType = PartitionTable::msdos;
+        if ( isEfi || createHybridBootloaderLayout )
+        {
+            partType = PartitionTable::gpt;
+        }
```
dalto8:
```suggestion
        partType = ( isEfi || createHybridBootloaderLayout ) ? PartitionTable::gpt : PartitionTable::msdos;
```

ArrayBolt3:
Alright, commits now cleaned up a bit.

src/modules/bootloader/main.py
```diff
@@ -909,8 +914,19 @@ def run():
     """
 
     fw_type = libcalamares.globalstorage.value("firmwareType")
+    boot_loader = libcalamares.globalstorage.value("bootLoader")
+
+    try:
+        install_hybrid_grub = libcalamares.job.configuration["installHybridGRUB"]
+    except KeyError:
+        install_hybrid_grub = False
```
dalto8:
```suggestion
    install_hybrid_grub = libcalamares.job.configuration.get("installHybridGRUB", False)
```

src/modules/bootloader/main.py
```diff
@@ -909,8 +914,19 @@ def run():
...
+
+    try:
+        efi_boot_loader = libcalamares.job.configuration["efiBootLoader"]
+    except KeyError:
+        efi_boot_loader = ""
```
dalto8:
```suggestion
    efi_boot_loader = libcalamares.job.configuration.get("efiBootLoader", "")
```

src/modules/bootloader/main.py
```diff
@@ -923,7 +939,12 @@ def run():
             return None
 
     try:
-        prepare_bootloader(fw_type)
+        prepare_bootloader(fw_type, install_hybrid_grub)
+#        if efi_boot_loader == "grub" and install_hybrid_grub:
+#            prepare_bootloader("bios")
+#            prepare_bootloader("efi")
+#        else:
+#            prepare_bootloader(fw_type)
```
dalto8: If this isn't needed we should remove it.

src/modules/bootloader/main.py
```diff
@@ -893,8 +898,8 @@ def prepare_bootloader(fw_type):
         install_secureboot(efi_directory)
     elif efi_boot_loader == "refind" and fw_type == "efi":
         install_refind(efi_directory)
-    elif efi_boot_loader == "grub" or fw_type != "efi":
-        install_grub(efi_directory, fw_type)
+    elif efi_boot_loader == "grub" or fw_type != "efi" or install_hybrid_grub:
```
dalto8: Are we sure we want this check here?  Do we really want to follow the hybrid grub path on EFI system where something other than "grub" is specified?

src/modules/bootloader/main.py
```diff
@@ -637,28 +637,33 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
             check_target_env_call([libcalamares.job.configuration["grubInstall"],
                                    "--target=i386-pc",
                                    "--recheck",
                                    "--force",
-                                   boot_loader["installPath"]])
+                                   boot_loader_install_path])
 
 
-def install_grub(efi_directory, fw_type):
+def install_grub(efi_directory, fw_type, install_hybrid_grub):
```
dalto8, line 679: Don't forget to add the new param to the doc

src/modules/bootloader/main.py
```diff
@@ -671,7 +676,7 @@ def install_grub(efi_directory, fw_type):
         libcalamares.utils.warning(_("Failed to install grub, no partitions defined in global storage"))
         return
 
-    if fw_type == "efi":
+    if fw_type == "efi" or install_hybrid_grub:
         libcalamares.utils.debug("Bootloader: grub (efi)")
```
dalto8, line 697: Should we add the state of `install_hybrid_grub` to this debug message?

src/modules/bootloader/main.py
```diff
@@ -712,9 +717,9 @@ def install_grub(efi_directory, fw_type):
             efi_file_target = os.path.join(install_efi_boot_directory, efi_boot_file)
 
             shutil.copy2(efi_file_source, efi_file_target)
-    else:
+    if fw_type == "bios" or install_hybrid_grub:
         libcalamares.utils.debug("Bootloader: grub (bios)")
-        run_grub_install(fw_type, partitions, None)
+        run_grub_install("bios", partitions, efi_directory, install_hybrid_grub)
```
dalto8, line 741: We should probably have an error case here in the event we fall through all those conditions.

src/modules/bootloader/main.py
```diff
@@ -637,28 +637,33 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+        if efi_directory is not None:
+            if not install_hybrid_grub:
+                return
 
-        boot_loader = libcalamares.globalstorage.value("bootLoader")
-        if boot_loader["installPath"] is None:
-            return
+        if libcalamares.globalstorage.value("bootLoader") is None:
+            find_esp_disk_command = "echo \"/dev/$(lsblk -o PKNAME $(df --output=source /boot/efi | tail -n1) | tail -n1)\""
+            boot_loader_install_path = check_target_env_output(["sh", "-c", find_esp_disk_command]).strip()
```
dalto8: If bootloader is None and fw_type is efi and install_hybrid_grub is false shouldn't we be not installing a bootloader here?

dalto8 (changes requested):

src/modules/bootloader/main.py
ArrayBolt3: Hmm, that's a good question. Really, if `install_hybrid_grub` is `True`, setting a boot loader other than GRUB should be illegal. That would make this check obsolete, since the BIOS path will always install with GRUB anyway, and the EFI path will install with GRUB if it's required in order to use `install_hybrid_grub`.

src/modules/bootloader/main.py
ArrayBolt3: Yeah, that would be a good idea.

src/modules/bootloader/main.py
ArrayBolt3: `fw_type` is essentially a boolean right now - it's only ever `bios` or `efi`, to my awareness. However, I do see how it could end up blank if something goes wrong, and Calamares might in theory support more firmware types in the future. Maybe it would be a good idea to error out at the *top* if `fw_type` is anything other than `bios` or `efi`? That would be more efficient and still catch the kinds of problems that could surface here.

src/modules/bootloader/main.py
ArrayBolt3: Yes, but I think that's already the case, since the context is:

```python
if fw_type == "efi":
    ...do EFI things here...
else:
   ...
   libcalamares.globalstorage.value("bootLoader") is None:
        find_esp_disk_command = "long line of bash"
        boot_loader_install_path = check_target_env_output([...]).strip()
```

i.e., this code will only run if `fw_type` is NOT `efi`.

src/modules/bootloader/main.py
```diff
@@ -637,28 +637,33 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
-            return
+        if efi_directory is not None:
+            if not install_hybrid_grub:
+                return
 
-        boot_loader = libcalamares.globalstorage.value("bootLoader")
-        if boot_loader["installPath"] is None:
-            return
+        if libcalamares.globalstorage.value("bootLoader") is None:
+            find_esp_disk_command = "echo \"/dev/$(lsblk -o PKNAME $(df --output=source /boot/efi | tail -n1) | tail -n1)\""
```
dalto8: We shouldn't hardcode `/boot/efi`

The efi mountpoint is configurable in the partition module.

ArrayBolt3: gah, I knew that, but forgot it here. Will fix in the next push.

src/modules/bootloader/main.py
dalto8, line 649: We should probably check to ensure that `boot_loader_install_path` is something like we expect.

src/modules/bootloader/main.py
ArrayBolt3: Um, wait, the efi mountpoint *is* configurable, but I'm pretty sure that's stored in `libcalamares.globalstorage.value("bootLoader")["installPath"]` no? This is the codepath for if `libcalamares.globalstorage.value("bootLoader") is None`, in which case we don't have that info. Is there another convenient spot to grab it from, or should I be parsing the partition layout to find the right spot and falling back to `/boot/efi` if all else fails?

dalto8: I think it is also in global storage.  Something like `efiSystemPartition`

src/modules/bootloader/main.py
ArrayBolt3: Well this is fine, looks like `bootloader.schema.yaml` is being ignored entirely. Not sure why, but I can set any parameter to any value and Calamares doesn't choke on it when first loading.

ArrayBolt3: @dalto8 Are the schema.yaml files actually used to check config at Calamares startup? I just tried using git master, 3.3.14, 3.3.5 (I think), and 3.2.61 and *all* of them are unable to notice when I mangle a module's config file.

src/modules/partition/core/PartitionActions.cpp
ArrayBolt3: Applied.

src/modules/bootloader/main.py
ArrayBolt3: Applied.

src/modules/bootloader/main.py
ArrayBolt3: Applied.

src/modules/bootloader/main.py
ArrayBolt3: Removed.

src/modules/bootloader/main.py
ArrayBolt3: Gave up on using the schema, added a validation blurb near the start of `run()` and removed the unnecessary check.

src/modules/bootloader/main.py
ArrayBolt3: Added.

src/modules/bootloader/main.py
ArrayBolt3: Added.

src/modules/bootloader/main.py
ArrayBolt3: Added a check at the top as described.

src/modules/bootloader/main.py
ArrayBolt3: Was able to use `efiSystemPartition`. `/boot/efi` is still used as a final fallback, but only if all other options fail.

src/modules/bootloader/main.py
ArrayBolt3: Added a check to make sure the path begins with `/dev/` and is at least eight characters long (the length of `/dev/sda`, to my awareness there aren't any two-character disk names in Linux).

ArrayBolt3:
Did a smoke test (hybrid installation on BIOS hardware, try to boot it with both BIOS and UEFI hardware), the latest commits appear to work.

src/modules/bootloader/main.py
```diff
@@ -637,42 +637,61 @@ def run_grub_install(fw_type, partitions, efi_directory):
                                    "--bootloader-id=" + efi_bootloader_id,
                                    "--force"])
     else:
-        assert efi_directory is None
-        if libcalamares.globalstorage.value("bootLoader") is None:
-            return
+        if efi_directory is not None:
```
adriaandegroot: Perhaps combine these? `if efi-directory-set but not install-hybrid: print(error); return;`, I think it would be good to log some kind of error anyway.

src/modules/bootloader/main.py
```diff
@@ -637,42 +637,61 @@ def run_grub_install(fw_type, partitions, efi_directory):
...
+                return
 
-        boot_loader = libcalamares.globalstorage.value("bootLoader")
-        if boot_loader["installPath"] is None:
-            return
+        if libcalamares.globalstorage.value("bootLoader") is None:
+            efi_install_path = libcalamares.globalstorage.value("efiSystemPartition")
+            if efi_install_path is None or efi_install_path == "":
+                efi_install_path = "/boot/efi"
+            find_esp_disk_command = f"echo \"/dev/$(lsblk -o PKNAME $(df --output=source '{efi_install_path}' | tail -n1) | tail -n1)\""
```
adriaandegroot: I get the feeling we're doing things in shell that we don't have to. `lsblk $(df | tail -n1)` ok, you want the substitution there, and *maybe* the second `tail -n1` as well, but why do string-manipulation there as well, .. you could do this in one or more steps with target_env_output and then add the `/dev/` bit in Python.

src/modules/partition/core/PartitionActions.cpp
```diff
@@ -144,6 +149,26 @@ doAutopartition( PartitionCoreModule* core, Device* dev, Choices::AutoPartitionO
         }
         core->createPartition( dev, efiPartition, KPM_PARTITION_FLAG_ESP );
         firstFreeSector = lastSector + 1;
+
+        if ( createHybridBootloaderLayout )
+        {
+            // TODO: Avoid hardcoding 8 MiB?
+            qint64 bios_part_sizeB = 8388608;
```
adriaandegroot: 537:    qint64 uefisys_part_sizeB = 300_MiB;

src/modules/partition/gui/PartitionLabelsView.cpp
```diff
@@ -190,6 +192,11 @@ PartitionLabelsView::buildTexts( const QModelIndex& index ) const
             {
                 firstLine = tr( "EFI system", "@label" );
             }
+            else if ( index.data( PartitionModel::FileSystemTypeRole ).toInt() == FileSystem::Unformatted
+                      && static_cast< Partition* >( index.data( PartitionModel::PartitionPtrRole ).value< void* >() )->property( "_calamares_flags" ).toUInt() & KPM_PARTITION_FLAG( BiosGrub ) )
```
adriaandegroot: Can we give all that casting-and-wrangling a nicer name, make it a static function in this translation unit?

src/modules/bootloader/main.py
ArrayBolt3: Good idea. Implemented a warning message with `libcalamares.utils.warning`.

src/modules/bootloader/main.py
ArrayBolt3: Valid. You may be able to tell I'm a bit handier with bash than with Python by this :P

The second `tail -n1` can be removed without much trouble, and it removes a fork, which should help performance a tiny bit.

src/modules/partition/core/PartitionActions.cpp
ArrayBolt3: Oh nice, didn't know that we had `_MiB` literals in Cala. I see them now. Will use this.

src/modules/partition/gui/PartitionLabelsView.cpp
ArrayBolt3: Will do.

ArrayBolt3:
Gentle ping, anything left to do on this?

PR #2424, opened 2025-01-27
Save obfuscated luks passphrase in calamares globals variables

SixK:
Save obfuscated luks passphrase in calamares globals variables.
We use the same obfuscating function as for user password.
Passphrase can be used later in a dedicated calamares module to associates token device with passphrase to unlock a Luks partition at boot.


adriaandegroot:
Let's land this and then clean up the code after.

SixK:
Thank's for the merge and sorry if code is not clean.

PR #2428, opened 2025-02-13
Improve exceptions to ClearMountsJob

adriaandegroot:
Inspired by #2427 , which adds one-more-special-case, I decided to add a configuration option so that distro's can configure whatever tool or name is used when special filesystems / live-filesystems are in use.

@JustTNE @ArrayBolt3


ArrayBolt3:
This looks pretty straightforward to me, I do question if `doNotClose` is the best name for this option though since we're unmounting things, not closing files. Other than that, this compiles just fine for me on Kubuntu 24.04. I haven't tested it personally due to time limitations, but this looks good to me.

JustTNE:
LGTM

dalto8 (approved):

PR #2435, opened 2025-02-24
Handle Keyboard on Ubuntu 24.10+ when using Wayland and Gnome.

SixK:
With Ubuntu 24.10+ (Oracular) it seem's that way to handle Keyboard has changed when using Gnome and Wayland.
X11 commands don't work anymore and Keyboard management seem's to be per account.

Here is a way to handle keyboard using gsettings. All configurations are handle as before (extra-keyboard, shortcuts, etc..)

As Calamares need root privileges, we need to get default user DBUS_SESSION_BUS_ADRESSE to change keyboard for it.
This is done getting username associated to userid 1000.

By default Calamares will not use this code, but it can be activated changing this keyboard.conf value to True:
```
configure:
    gnome: false
```

### DIFF
--- a/src/modules/keyboard/keyboard.conf
+++ b/src/modules/keyboard/keyboard.conf
@@ -43,3 +43,5 @@ configure:
     # compositor KWin with command-line argument `--locale1`. That
     # argument makes this configuration option unnecessary.
     kwin: false
+    # Configure keyboard when using Wayland with Gnome on Ubuntu 24.10+
+    gnome: false
--- a/src/modules/keyboard/Config.cpp
+++ b/src/modules/keyboard/Config.cpp
@@ -375,6 +375,98 @@ applyKWin( const BasicLayoutInfo& settings, AdditionalLayoutInfo& extra )
     }
 }
 
+// Try to generate such string -> "[('xkb', 'uk+latin1'), ('xkb','en')]"
+QString tupleListToString(const QList<QPair<QString, QString>>& tupleList) {
+    QString result = "[";
+    for (int i = 0; i < tupleList.size(); ++i) {
+        const auto& tuple = tupleList[i];
+        result += QString("('%1', '%2')").arg(tuple.first).arg(tuple.second);
+        if (i < tupleList.size() - 1) {
+            result += ", ";
+        }
+    }
+    result += "]";
+    return result;
+}
+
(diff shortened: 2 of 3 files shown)


tsimonq2:
@SixK If you need help getting this into the Ubuntu archive for 24.10, let me know! I'd be happy to sponsor, just needs [some paperwork](https://documentation.ubuntu.com/sru/en/latest/). :)

src/modules/keyboard/keyboard.conf
```diff
@@ -43,3 +43,5 @@ configure:
     # compositor KWin with command-line argument `--locale1`. That
     # argument makes this configuration option unnecessary.
     kwin: false
+    # Configure keyboard when using Wayland with Gnome on Ubuntu 24.10+
```
dalto8, line 46: Is this really Ubuntu specific or does it apply to certain versions of Gnome/Wayland universally?

SixK: I guess pb will happen to any distribution using Gnome and Wayland as default. 
I had the problem with Ubuntu, I don't know if any other distribution use this.
I also found nobody reporting the problem here or in lubuntu calamares github project.

kkofler: As far as I know, `setxkbmap` is useless on all Wayland desktop environments. Unfortunately, what should be used instead depends on the desktop environment. I guess this code should work for any modern version of GNOME (Mutter), KWin has its own way for which there is already code present, other compositors will do their own thing for which code has yet to be written. There is unfortunately no standardized protocol like xkb under Wayland.

SixK:
> @SixK If you need help getting this into the Ubuntu archive for 24.10, let me know! I'd be happy to sponsor, just needs [some paperwork](https://documentation.ubuntu.com/sru/en/latest/). :)

@tsimonq2, Thank's for the proposition, what do you mean by paperwork ? 
Who would create it ? 
Is there a template to follow ?

kkofler:
The "paperwork" (bureaucracy) is if you want this to get into the official Ubuntu 24.10 updates repository at Ubuntu (Canonical), because 24.10 was already released, so there needs to be a good reason to get a package updated there.

The link that @tsimonq2 posted: https://documentation.ubuntu.com/sru/en/latest/ already explains what needs to be done for that, in particular: https://documentation.ubuntu.com/sru/en/latest/howto/

Issue #2416, opened 2024-12-23
Enhance GRUB UEFI fallback bootloader installation support with OS-specific code


ArrayBolt3:
**Is your feature request related to a problem? Please describe.**
Calamares currently has a setting `installEFIFallback` in `/etc/calamares/modules/bootloader.conf`. This feature is supposed to install a copy of the OS bootloader to the removable media path (e.g., `/boot/efi/EFI/BOOT`), allowing the installation to be booted even if the UEFI BIOS isn't able to find the OS-specific bootloader.

While dynamite in theory, the implementation is a bit too simple to work reliably on all distros. Right now all it does (simplifying away arch-specific considerations) is finds the `grubx64.efi` file from the normal bootloader location for the OS, and copies it to `bootx64.efi` in the removable media path. While I guess this could work on some distros, this doesn't work at all on Debian (and by extension Kicksecure) for at least two reasons:

* Debian's GRUB requires a stub `grub.cfg` file to be shipped alongside the GRUB EFI executable itself. This stub config tells GRUB how to find the real `grub.cfg`. Without this stub config file, the bootloader simply drops you to a GRUB shell.
* `grubx64.efi` can't be booted directly with Secure Boot in 99.9% of circumstances. You have to boot `shimx64.efi`. The shim then boots `grubx64.efi` which in turn boots the kernel and initramfs. If you copy `grubx64.efi` to the removable media path on a Secure Boot system, it is unusable because Secure Boot will reject it.

Under Debian, the proper way to install GRUB to the fallback bootloader location is:

```
sudo debconf-set-selections <<< "grub-efi-amd64 grub2/force_efi_extra_removable boolean true"
sudo dpkg-reconfigure --default-priority grub-efi-amd64
```

In essence, this sets a configuration variable in debconf that tells Debian "any time you're reinstalling GRUB, make sure to install it to both the normal *and* fallback bootloader locations." Then it tells Debian to reconfigure the grub-efi-amd64 package, thus resulting in GRUB being reinstalled to both the normal and fallback locations as requested.

Obviously, this won't work on anything other than Debian and *some* of its derivatives (notably, this can't be used on Ubuntu because Ubuntu's GRUB package doesn't have the `grub2/force_efi_extra_removable` debconf setting), but for Debian and suitably similar derivatives, it would be far preferable to the current way of installing the fallback bootloader.

**Describe the solution you'd like**
[...]

PR #2446, opened 2025-03-23
[bootloader] Enhance fallback bootloader installation on Debian

ArrayBolt3:
The fallback bootloader installation code previously only copied the GRUB bootloader's EFI executable to the removable media path. This doesn't work at all on Debian, the bootloader ends up being unusable at worst (on Secure Boot systems) and drops you to a GRUB shell at best. Add Debian-specific fallback bootloader installation code to fix this.

Implements https://github.com/calamares/calamares/issues/2416 for Debian only. The changes should allow customizations for other distros to be added with relatively little difficulty.

### DIFF
--- a/src/modules/bootloader/main.py
+++ b/src/modules/bootloader/main.py
@@ -656,7 +656,7 @@ def run_grub_install(fw_type, partitions, efi_directory, install_hybrid_grub):
             boot_loader_install_path = boot_loader["installPath"]
             if boot_loader_install_path is None:
                 return
-            
+
         # boot_loader_install_path points to the physical disk to install GRUB
         # to. It should start with "/dev/", and be at least as long as the
         # string "/dev/sda".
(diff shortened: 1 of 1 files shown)


dalto8:
I don't really like the idea of hardcoded OS strings that leads to distro specific implementations in the bootloader module.

ArrayBolt3:
@dalto8 Is there a way to introduce distro-specific implementations some other way that you'd prefer?

dalto8:
If a distro needs custom fallback code, I think it might be better to set efifallback to false and handle it in a distro-specific module.

ArrayBolt3:
I mean, maybe, but arguably if a feature is completely broken on a major mainstream distro, it seems like it would make sense to fix the feature.

This does raise the question, are there any distros the EFI fallback feature does work on (beyond providing a GRUB shell)? If it's broken everywhere (don't know if it is or isn't yet), then it probably *needs* to have distro-specific code, or it may need to be removed entirely in favor of distro-specific mechanisms for everyone.

dalto8:
> are there any distros the EFI fallback feature does work on

Yes.  Probably most/all distros that don't support secure boot. I suspect that this makes up a pretty large percentage of the distros using Calamares.

ArrayBolt3:
On Debian, the primary problem isn't Secure Boot, but is the requirement for there to be a stub grub.cfg file pointing to the real grub.cfg. Even without Secure Boot, all you get is a GRUB shell on Debian.

That being said, I just gave Artix a spin and it does seem to work just fine with this feature, so I guess this way of doing things is a weird Debian-ism or something.

adriaandegroot:
Closing this after pulling it into the repo, in a separate branch. We can continue the topic at the other side of migration.

PR #2447, opened 2025-04-03
[users] Add a choice to display autologin checkbox

Neixen911:
By default, the autologin checkbox is display to the user and he can choose if he wants to enable it or not. By adding a variable to display the checkbox or not, this checkbox can be not displaying to the user and avoid him to change it.

### DIFF
--- a/src/modules/users/Config.cpp
+++ b/src/modules/users/Config.cpp
@@ -1017,6 +1017,7 @@ Config::setConfigurationMap( const QVariantMap& configurationMap )
 
     // Renaming of Autologin -> AutoLogin in 4ffa79d4cf also affected
     // configuration keys, which was not intended. Accept both.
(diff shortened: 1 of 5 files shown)


dalto8:
If we want to make this optional, I think we should add a setting in the users.conf file for it.

Neixen911:
I added a setting in the users.conf file like you said to make it optional.

dalto8:
Thanks!

The default should be set to match the existing behavior and then distros can override in the config.  Since, this is currently enabled by default, the default should be `true`.

Also, don't forget to update `users.schema.yaml` with the new key.

Neixen911:
Thanks for your feedback !

I have set the value of displayAutoLogin setting to `true` by default now.

Plus, I updated the file `users.schema.yaml` with this setting too.

dalto8 (approved):

PR #2448, opened 2025-04-03
bootloader: Drop bootloader.orig file

ptr1337:
Has been somehow merged in #2445 

This should be not in the repo and therefore lets remove it.


dalto8:
Thanks!

